PR #10, opened 2023-10-23
feat: Rework consensus storage crate

slowli:
# What ❔

- Extracts storage API into trait(s) and make them async.
- Introduces buffered storage (i.e., one that schedules sequential block execution).

## Why ❔

Part of preparations for the integration with the server codebase.


node/actors/consensus/src/replica/state_machine.rs
```diff
@@ -118,6 +120,16 @@ impl StateMachine {
...
+        let store_result = scope::run_blocking!(ctx, |ctx, s| {
+            let backup_future = self.storage.put_replica_state(ctx, &backup);
+            s.spawn(backup_future).join(ctx).block()?;
+            Ok(())
+        });
+        match store_result {
+            Ok(()) => { /* Everything went fine */ }
+            Err(StorageError::Canceled(_)) => tracing::trace!("Storing replica state was canceled"),
+            Err(StorageError::Database(err)) => panic!("Failed storing replica state: {err}"),
+            // ^ We don't know how to recover from DB errors, so panicking is the only option so far.
```
pompon0, line 132: nit: the idiomatic thing would be to return an error here, not panic. It is not up to the code which detected an error to assess whether it is recoverable or not. And even if it is not recoverable now, it doesn't mean that it won't be recoverable in the future. In this particular case, if we abstract out the storage it might be possible that the error is transient for example.

node/libs/storage/src/lib.rs
```diff
@@ -35,46 +57,50 @@ pub struct Storage {
...
+            Ok(scope
+                .spawn_blocking(|| Self::new_blocking(genesis_block, path))
+                .join(ctx)
+                .await?)
+        })
+        .await
+    }
+
+    /// Blocking version of [`Self::new()`].
+    fn new_blocking(genesis_block: &validator::FinalBlock, path: &Path) -> anyhow::Result<Self> {
```
pompon0, line 80: why do we need the blocking version of new? Is it so that a single blocking thread is used for all initial db accesses?

brunoffranca (approved):
Looks good. I only had a couple of minor questions. Also, I would suggest organizing the storage crate a bit differently to make stuff easier to find:
- Moving all the traits into a traits.rs file.
- Moving RocksdbStorage and all it's methods/implementations into a rocksdb.rs file.

node/libs/storage/src/buffered.rs
```diff
@@ -0,0 +1,276 @@
...
+use std::{collections::BTreeMap, ops};
+
+/// [`BlockStore`] variation that upholds additional invariants as to how blocks are processed.
+///
+/// The invariants are as follows:
+///
+/// - Stored blocks always have contiguous numbers; there are no gaps.
+/// - Blocks can be scheduled to be added using [`Self::schedule_next_block()`] only. New blocks do not
+///   appear in the store otherwise.
+#[async_trait]
```
pompon0, line 24: Are you sure this is the right place for specifying this abstraction? AFAIU it is supposed to be the wrapper for postgres in zksync-era, and the consensus repo doesn't really have any direct use for it. How about moving ContiguousBlockStore and BufferedStorage to zksync-era? And perhaps we won't need to make it generic then?

pompon0 (changes requested):
added comments

node/libs/storage/src/buffered.rs
```diff
@@ -0,0 +1,276 @@
...
+                buffer.set_store_block(store_block_number);
+                buffer.next_block_for_store()
+            };
+            if let Some(block) = next_block_for_store {
+                self.inner.schedule_next_block(ctx, &block).await?;
+                let block_number = block.block.number;
+                tracing::trace!(%block_number, "Block scheduled in underlying storage");
+            }
+
+            #[cfg(test)]
```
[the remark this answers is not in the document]
slowli: `rustc` will complain about unused fields in `BufferedStorageEvent` otherwise. Do you think it would be better to make `BufferedStorageEvent` public?

node/libs/storage/src/buffered.rs
slowli: I thought about `BufferedStorage` as a helper that is generic enough that it can be placed close to functionality that provides (i.e., `BlockStore` / `WriteBlockStore` traits), as a helper. In general terms, it just expresses sequential block execution, which is not specific to `zksync-era` / Postgres-based DAL. Does this make sense, or not quite? Regarding generics, I wouldn't remove them in any case because they make `BufferedStorage` easier to test.

node/actors/executor/src/main.rs
```diff
@@ -95,10 +93,12 @@ async fn main() -> anyhow::Result<()> {
         configs.validator_key.clone(),
         validator_set.clone(),
         storage.clone(),
-    );
+    )
+    .await
+    .context("consensus")?;
     // FIXME(slowli): Run `sync_blocks` actor once it's fully functional
```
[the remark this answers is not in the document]
slowli: My fault; didn't fix this in time. Added this logic now.

brunoffranca (approved):

pompon0 (approved):

PR #11, opened 2023-10-23
Switch from bls12_381 to bn254

moshababo:
Switching from [bls12_381](https://hackmd.io/@benjaminion/bls12-381) to [bn254](https://hackmd.io/@jpw/bn254) for validator keys due to [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) delays, making the former incompatible with EVM.

This PR implements [BLS signature scheme](https://medium.com/cryptoadvance/bls-signatures-better-than-schnorr-5a7fe30ea716) for bn254.

### Notes:
* Code was duplicated from [crypto/src/bls12_381](https://github.com/matter-labs/era-consensus/tree/bn254/node/libs/crypto/src/bls12_381) to [crypto/src/bn254](https://github.com/matter-labs/era-consensus/tree/bn254/node/libs/crypto/src/bn254)
* [crypto/src/bls12_381](https://github.com/matter-labs/era-consensus/tree/bn254/node/libs/crypto/src/bls12_381) code was not removed, but remains unused, for future upgrade

### TODO:
* ~Replace the underlying [crates/bn254](https://crates.io/crates/bn254) with [crates/ark-bn254](https://crates.io/crates/ark-bn254)~
* ~Verify that signature aggregation can support different messages (the newly added `bn254::tests::aggregate_signature_distinct_messages` should pass)~
* ~Improve `bn254::tests`~
* ~Remove `bls12_381` from protobuf schemas~
* ~Run integration tests~

Closing BFT-325.


node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+mod tests;
+
+pub struct SecretKey(PrivateKey);
+
+impl SecretKey {
+    /// Generates a random secret key
+    pub fn random<R: Rng>(rng: &mut R) -> Self {
+        let private_key = PrivateKey::random(rng);
+        return Self(private_key);
+    }
```
pompon0, line 25: implement "Distribution<SecretKey> for Standard" instead.

node/libs/crypto/src/bn254/tests.rs
```diff
@@ -0,0 +1,91 @@
...
+    let sig = sk.sign(&msg);
+
+    sig.verify(&msg, &pk).unwrap()
+}
+
+#[test]
+fn signature_failure_smoke() {
+    let mut rng = StdRng::seed_from_u64(29483920);
+
+    let sk1 = SecretKey::decode(&rng.gen::<[u8; 32]>()).unwrap();
```
pompon0, line 21: use rng.gen() to generate the secret key directly

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AggregateSignature(Signature);
+
+impl AggregateSignature {
+    pub fn aggregate<'a>(sigs: impl IntoIterator<Item = &'a Signature>) -> Self {
+        let sigs: Vec<bn254::Signature> = sigs.into_iter().map(|s| s.0).collect();
+        let mut agg = sigs[0];
```
pompon0, line 144: this will panic if sigs is empty. Please fix.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        ByteFmt::encode(self).cmp(&ByteFmt::encode(other))
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AggregateSignature(Signature);
+
+impl AggregateSignature {
+    pub fn aggregate<'a>(sigs: impl IntoIterator<Item = &'a Signature>) -> Self {
+        let sigs: Vec<bn254::Signature> = sigs.into_iter().map(|s| s.0).collect();
```
pompon0, line 143: Instead of building a vector, you can take the first element using next() and then iterate over the remaining elements.

moshababo:
@pompon0 thanks for the review. Most of the code was temp and will be changed due to the replacement of the curve implementation, but I'll address the parts which will remain.

brunoffranca (commented):
Looking good so far. The math is all correct.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/tests.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/Cargo.toml
```diff
@@ -26,17 +26,22 @@ assert_matches = "1.5.0"
 async-trait = "0.1.71"
 bit-vec = "0.6"
 blst = "0.3.10"
+bn254 = "0.0.1"
```
brunoffranca, line 29:
```suggestion
```
We no longer need this dependency.

node/deny.toml
```diff
@@ -44,7 +44,7 @@ allow = [
 
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "deny"
+multiple-versions = "warn"
```
brunoffranca, line 47:
```suggestion
multiple-versions = "deny"
```
So, just add it to the skip list below.

node/libs/crypto/src/bn254/hash.rs
```diff
@@ -0,0 +1,26 @@
+//! Hash operations.
+
+use ark_bn254::{G1Affine, G1Projective};
+use ark_ec::AffineRepr as _;
+use sha2::Digest as _;
+
+/// H(m) as a point in G1.
```
brunoffranca, line 7:
```suggestion
/// Hashes an arbitrary message and maps it to an elliptic curve point in G1.
```

node/libs/crypto/src/bn254/hash.rs
```diff
@@ -0,0 +1,26 @@
...
+pub(crate) fn hash_to_g1(msg: &[u8]) -> G1Projective {
+    for i in 0..100 {
+        // Hash the message with the index as suffix.
+        let bytes: [u8; 32] = sha2::Sha256::new()
+            .chain_update(msg)
+            .chain_update((i as u32).to_be_bytes())
+            .finalize()
+            .into();
+
+        // Try to get a G1 point from the hash.
```
brunoffranca, line 17:
```suggestion
        // Try to get a G1 point from the hash. The probability that this works is around 1/8.
```

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(Clone)]
+pub struct PublicKey(pub G2);
+
+impl PartialEq for PublicKey {
+    fn eq(&self, other: &Self) -> bool {
+        ByteFmt::encode(self).eq(&ByteFmt::encode(other))
+    }
+}
+
+impl Eq for PublicKey {}
```
brunoffranca, line 70: This should not be necessary. I think you are able to just derive this trait, since the G2 type already implements it.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+impl PartialEq for Signature {
+    fn eq(&self, other: &Self) -> bool {
+        ByteFmt::encode(self).eq(&ByteFmt::encode(other))
+    }
+}
+
+impl Eq for Signature {}
```
brunoffranca, line 132: Same, you're probably able to derive this trait.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+impl Ord for Signature {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        ByteFmt::encode(self).cmp(&ByteFmt::encode(other))
+    }
+}
+/// Type safety wrapper around [Signature] indicating that it is an aggregated signature.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AggregateSignature(Signature);
```
brunoffranca, line 160:
```suggestion
pub struct AggregateSignature(pub G1);
```
It would be preferable to store the point directly.

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> SecretKey {
+        let rand = ark_bn254::Fr::new(rng.gen());
+        SecretKey(rand)
+    }
+}
+
+/// Generates a random PublicKey. This is meant for testing purposes.
+impl Distribution<PublicKey> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> PublicKey {
+        rng.gen::<SecretKey>().public()
```
brunoffranca, line 17: It's simpler to just generate a random G2 point.

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+/// Generates a random Signature. This is meant for testing purposes.
+impl Distribution<Signature> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Signature {
+        let key = rng.gen::<SecretKey>();
+        let msg = rng.gen::<[u8; 4]>();
+        key.sign(&msg)
```
brunoffranca, line 26: Same. It's simpler to generate a random G1 point.

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+/// Generates a random AggregateSignature. This is meant for testing purposes.
+impl Distribution<AggregateSignature> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> AggregateSignature {
+        let sig = rng.gen();
+        AggregateSignature(sig)
```
brunoffranca, line 34: Same here.

node/libs/crypto/src/lib.rs
```diff
@@ -1,8 +1,8 @@
 //! Collection of cryptographic primitives used in zksync-bft repository.
 
-pub mod bls12_381;
```
brunoffranca, line 3: Don't delete the module. Let's keep it here.

brunoffranca (approved):
Looks good to me. You just need to change the bls12_281 fields in the protobufs schema and it's good to go.

node/deny.toml
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/testonly.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/testonly.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/testonly.rs
moshababo: Fixed.

node/libs/crypto/src/lib.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -0,0 +1,34 @@
...
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> SecretKey {
+        let rand = ark_bn254::Fr::new(rng.gen());
+        SecretKey(rand)
+    }
+}
+
+/// Generates a random PublicKey. This is meant for testing purposes.
+impl Distribution<PublicKey> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> PublicKey {
+        rng.gen()
```
pompon0, line 18: those sample() methods are recursive and cause stack overflows

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.0.serialize_compressed(&mut buf).unwrap();
+        buf
+    }
+}
+
+impl std::hash::Hash for PublicKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
```
slowli, line 78: Nit: This trait can be derived.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+use anyhow::anyhow;
+use ark_bn254::{Bn254, Fr, G1Projective as G1, G2Projective as G2};
+use ark_ec::{
+    pairing::{Pairing as _, PairingOutput},
+    Group as _,
+};
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+pub use error::Error;
+use num_traits::Zero as _;
+use std::{collections::HashMap, fmt::Debug, hash::Hasher};
```
slowli, line 13: `fmt::Debug` import is unused (Clippy doesn't complain because it's used as a derive macro, but it's imported as a derive macro anyway).

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        let mut agg = G1::zero();
+        for sig in sigs {
+            agg += sig.0
+        }
+
+        AggregateSignature(agg)
+    }
+
+    #[inline(never)]
+    fn verify_raw(&self, msgs_and_pks: &[(&[u8], &PublicKey)]) -> Result<(), Error> {
```
slowli, line 155: Random question just to clarify: Do I understand correctly that there are proofs of knowledge are available for the aggregated `PublicKey`s to prevent rogue key attacks?

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    fn encode(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        self.0.serialize_compressed(&mut buf).unwrap();
+        buf
+    }
+}
+
+impl PartialOrd for AggregateSignature {
```
slowli, line 203: Off-topic(?): Is `PartialOrd` / `Ord` required only for `BTreeMap`s in `PrepareQC`? Could `HashMap` be used there? It's not quite obvious why signatures should be ordered semantically.

node/libs/crypto/src/bn254/hash.rs
```diff
@@ -0,0 +1,28 @@
+//! Hash operations.
+
+use ark_bn254::{G1Affine, G1Projective};
+use ark_ec::AffineRepr as _;
+use sha2::Digest as _;
+use tracing::instrument;
+
+/// Hashes an arbitrary message and maps it to an elliptic curve point in G1.
+#[instrument(level = "trace", skip_all)]
```
slowli, line 9: Bikeshedding: IMO, this function is too low-level to instrument.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+pub mod hash;
+mod testonly;
+
+/// Type safety wrapper around a scalar value.
+pub struct SecretKey(pub Fr);
+
+impl SecretKey {
+    /// Gets the corresponding [`PublicKey`] for this [`SecretKey`]
+    pub fn public(&self) -> PublicKey {
+        let p = G2::generator() * self.0;
```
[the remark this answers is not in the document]
moshababo: Implementation is [here](https://github.com/arkworks-rs/algebra/blob/57be20e56a142b059bca05653961f8a9ca4f54ae/ec/src/models/twisted_edwards/mod.rs#L62), and it isn't constant-time.

@pompon0 @brunoffranca any thoughts on this?

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+pub mod error;
+
+#[cfg(test)]
+mod tests;
+
+pub mod hash;
+mod testonly;
+
+/// Type safety wrapper around a scalar value.
+pub struct SecretKey(pub Fr);
```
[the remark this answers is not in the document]
moshababo: I thought I followed the previous bls12_381 practice, but I didn't. 
Agree, fixed.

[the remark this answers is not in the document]
moshababo: I thought I followed the previous bls12_381 practice, but it's not public there as well. Fixed.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Type safety wrapper around G2.
+#[derive(Clone, PartialEq, Eq)]
+pub struct PublicKey(pub G2);
+
+impl ByteFmt for PublicKey {
+    fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
+        G2::deserialize_compressed(bytes)
+            .map(Self)
+            .map_err(|e| anyhow!("failed to decode public key: {e:?}"))
```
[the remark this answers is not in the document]
moshababo: It requires `ark_serialize::SerializationError` to satisfy `anyhow::context::ext::StdError` trait bound, hence to use a type wrapper which implements `Error` + `Display`. 
I thought it's too cumbersome.
What do you think?

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: Fixed.

node/libs/crypto/src/bn254/mod.rs
moshababo: > Is `PartialOrd` / `Ord` required only for `BTreeMap`s in `PrepareQC`? 

Yes.

>  Could `HashMap` be used there? It's not quite obvious why signatures should be ordered semantically.

I didn't find use case for the current type of ordering. @pompon0 @brunoffranca any thoughts?

node/libs/crypto/src/bn254/hash.rs
moshababo: @pompon0 was this added for your debugging?

node/libs/crypto/src/bn254/mod.rs
moshababo: I understood from @brunoffranca that this attack is not applicable, but i'm not sure whether proofs of knowledge are used. 

@brunoffranca

node/libs/crypto/src/bn254/hash.rs
pompon0: yeah, you can remove it

node/libs/crypto/src/bn254/mod.rs
pompon0: Before a block is signed by anyone, the set of public keys of valid signers is already fixed.

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
pompon0: Yeah, for errors which are not anyhow::Context-compatible we can't do better than this AFAIK.

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
pompon0: seems like a good point, is the library that we are using here actually suitable for cryptographic applications?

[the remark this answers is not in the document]
pompon0: for example, blst has this protection explicitly stated in its specification (although under "SHOULD" quantifier): https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature#name-side-channel-attacks

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
slowli: Sorry, didn't check whether the error implements the [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) trait; I thought it would. `ark-serialize` uses [a shim](https://github.com/arkworks-rs/std) for the standard library for no-std use cases, which can be "rerouted" to `std` if the corresponding feature is enabled:

```toml
ark-serialize = { version = "0.4.2", features = ["std"] }
```

In this case, `SerializationError` will implement `Error`, and the problem will be gone. IMO, since we rely on `std` anyway, enabling this feature is harmless.

node/libs/crypto/src/bn254/mod.rs
pompon0: We need reserialization of signable messages (like PrepareQC) to be an identity, so that signatures are valid at all times. This is kind of constraining, but the alternative would be to pass around the serialized message at all times so that reserialization is never needed (which is in fact more bullet-proof, but a bit costly and makes message handling more complex).

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
brunoffranca: No, it's not constant time. There's no way we are going to find a Rust library for BN254 that is constant time.
This crate is appropriate for use in production. The arkworks libraries are used in several blockchain and non-blockchain projects. But being resistant to side-channel attacks is a whole different level.

node/libs/crypto/src/bn254/mod.rs
brunoffranca: Yes, there will be PoPs. We will have a validator registry on L1 and you'll be required to provide a PoP to have your key included.

node/libs/crypto/src/bn254/mod.rs
brunoffranca: +1

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+/// Type safety wrapper around a G1 value.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Signature(G1);
+
+impl Signature {
+    /// Verifies a signature against the provided public key
+    #[inline(never)]
```
pompon0, line 92: add a comment explaining why we have this nin-generic function and why inline never

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -0,0 +1,207 @@
+//! BLS signature scheme for the BN254 curve.
```
pompon0, line 1: add a disclaimer that it is not secure against side channel timing attacks

pompon0 (dismissed):
I don't know if you copied over any code to this pr from external repos, but if you did please familiarize yourself with https://www.notion.so/matterlabs/Dependency-Attribution-dcb684a5455f483ba370c637c443157f

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
moshababo: Fixed.

moshababo:
No code was copied.

node/libs/crypto/src/bn254/mod.rs
moshababo: Added.

node/libs/crypto/src/bn254/mod.rs
moshababo: Added.

pompon0 (approved):

brunoffranca (approved):

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
moshababo: zcash's lib implementation is similarly not constant-time: https://github.com/zcash-hackworks/bn/blob/ebf8b553f7e66f59e6221cf9b7a80bb31ff393a5/src/groups/mod.rs#L253

PR #12, opened 2023-10-25
feat: implemented BlockHeader (BFT-357)

pompon0:
Implemented https://www.notion.so/matterlabs/5d238e08fbed476b8550f697d9c39df6?v=51a4bffaba05499e8ca11bd625fff54a&p=16c93154366c4a3994457b44043dea71&pm=s with some differences. Summary of changes:
* Replaced Block with BlockHeader with just a hash of the payload
* Made consensus messages include the whole BlockHeader, rather than just hash (since it is small, once we moved the payload out)
* Introduced a strongly typed Payload
* Added protocol_version to the BlockHeader (and relevant checks to the consensus) - currently hardcoded to 0

Minor fixes:
* split the leader::Error and replica::Error into per-method errors (since the error types returned by each are disjoint)
* relaxed the "stringly" typed assertions from sync_block tests (if needed they should use strongly typed errors instead)
* added "size" to the protobuf encoding of BitVec, so that the encoded BitVec value is context free


node/libs/storage/src/types.rs
```diff
@@ -56,6 +53,20 @@ impl DatabaseKey {
     }
 }
 
+/// A payload of a proposed block which is not known to be finalized yet.
+/// Replicas have to persist such proposed payloads for liveness:
+/// consensus may finalize a block without knowing a payload in case of reproposals.
+/// TODO(gprusak): we do not store the BlockHeader, because it is always
+///   available in the LeaderPrepare message. However for better
+///   visibility we might want to store it anyway.
```
brunoffranca, line 61:
```suggestion
/// Currently we do not store the BlockHeader, because it is always
///   available in the LeaderPrepare message.
```
Let's avoid adding any new TODOS. If it's not important enough to discuss with the team and/or create a Linear ticket, then it's not important.

brunoffranca (changes requested):
Looks good, I only have a few minor nits.

node/actors/consensus/src/leader/state_machine.rs
```diff
@@ -59,18 +59,30 @@ impl StateMachine {
...
+        let label = match &input.msg {
+            validator::ConsensusMsg::ReplicaPrepare(_) => {
+                let res = match self.process_replica_prepare(ctx, consensus, input.cast().unwrap())
+                {
+                    Err(err) => {
+                        tracing::warn!("{err:#}");
+                        Err(())
+                    }
+                    Ok(()) => Ok(()),
+                };
```
slowli, line 71: Nit:  This can be expressed via `map_err()`.

node/actors/consensus/src/leader/tests.rs
```diff
@@ -25,17 +23,16 @@ async fn replica_commit() {
...
-        Err(Error::ReplicaCommitMissingProposal)
-    );
+    match consensus.leader.process_replica_commit(
+        ctx,
+        &consensus.inner,
+        test_replica_msg.cast().unwrap(),
+    ) {
+        Err(super::replica_commit::Error::UnexpectedProposal) => {}
+        res => panic!("unexpected result {res:?}"),
+    }
```
slowli, line 37: Nit: Could be expressed via `assert_matches` (the corresponding crate is already used in the workspace).

node/actors/consensus/src/replica/leader_prepare.rs
```diff
@@ -1,11 +1,73 @@
...
+    #[error(
+        "high QC of a future view (high QC view: {high_qc_view:?}, current view: {current_view:?}"
+    )]
+    HighQCOfFutureView {
+        high_qc_view: validator::ViewNumber,
+        current_view: validator::ViewNumber,
+    },
+    #[error("new block proposal when the previous proposal was not finalized")]
+    ProposalWhenPreviousNotFinalized,
+    #[error("block proposal with invalid parent hash (correct parent hash: {correct_parent_hash:#?}, received parent hash: {received_parent_hash:#?}, block: {header:?})")]
```
slowli, line 40: Nit: Like with other string literals, it's possible to split these using `\`-escaped newlines; it'll consume the whitespace at the start of the next line.

node/libs/roles/src/validator/messages/block.rs
```diff
@@ -44,55 +94,71 @@ impl TextFmt for BlockHash {
...
     }
 }
 
-/// A block.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Block {
+/// A block header.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BlockHeader {
+    ///
```
slowli, line 106: Is this comment supposed to be empty? 🙃

node/libs/schema/src/std_conv.rs
```diff
@@ -80,3 +80,24 @@ impl ProtoFmt for time::Duration {
...
+            anyhow::bail!("'vector' has less than 'size' bits");
+        }
+        this.truncate(size);
+        Ok(this)
+    }
+
+    fn build(&self) -> Self::Proto {
+        Self::Proto {
+            size: Some(self.len() as u64),
+            vector: Some(self.to_bytes()),
```
slowli, line 100: Bikeshedding: Maybe, use `bytes` instead of `vector`? IMO, "vector" looks quite ambiguous.

node/libs/roles/src/validator/messages/block.rs
```diff
@@ -4,6 +4,56 @@ use super::CommitQC;
 use crypto::{sha256, ByteFmt, Text, TextFmt};
 use std::fmt;
 
+/// Version of the consensus algorithm used by the leader in a given view.
+/// Currently it is stored in the BlockHeader (TODO: consider whether it should be present in every
+/// consensus message).
```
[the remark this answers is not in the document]
pompon0: The argument in favor is that if we have a new version which changes the semantics of the consensus message it may happen that the validator running old binary will be incorrectly interpreted by a validator running a new binary.

node/libs/roles/src/validator/messages/block.rs
```diff
@@ -4,6 +4,56 @@ use super::CommitQC;
...
+/// Currently it also determines the format and semantics of the block payload
+/// (TODO: consider whether the payload should be versioned separately).
```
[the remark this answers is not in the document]
pompon0: I had some afterthoughts after that discussion in the context of the above: if we keep the version (only) in the block, then some messages don't include the version at all. And if we (hypothetically) added version to both header and the messages, then forcing them to be the same wouldn't bring any benefit and might be constraining. Not a strong opinion, just writing down my thoughts.

node/libs/schema/src/std_conv.rs
pompon0: "bytes" is a reserved word in protobuf syntax

node/actors/sync_blocks/src/peers/tests.rs
```diff
@@ -825,49 +826,25 @@ async fn processing_invalid_sync_states() {
 
     let mut invalid_sync_state = test_validators.sync_state(1);
     invalid_sync_state.first_stored_block = test_validators.final_blocks[2].justification.clone();
-    let err = peer_states
-        .validate_sync_state(invalid_sync_state)
-        .unwrap_err();
-    let err = format!("{err:?}");
-    assert!(err.contains("first_stored_block"), "{err}");
+    assert!(peer_states.validate_sync_state(invalid_sync_state).is_err());
```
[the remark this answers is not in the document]
pompon0: Checking for substrings in tests is a bad practice, because it turns the tests into change detectors and is imprecise in the first place (especially in the case of dynamically constructed error messages, where there is no guarantee that the error message won't include the expected substring by accident). I somehow missed that during the review, sorry. Please use strongly typed errors if you want to distinguish them programmatically (it can be as simple as having an enum with each variant wrapping anyhow error).

node/libs/roles/src/validator/tests.rs
```diff
@@ -182,7 +181,7 @@ fn test_commit_qc() {
     .unwrap();
     let validator_set2 =
         ValidatorSet::new(vec![rng.gen(), rng.gen(), rng.gen(), rng.gen(), rng.gen()]).unwrap();
-    let validator_set3 = ValidatorSet::new(vec![sk1.public(), sk2.public(), sk3.public()]).unwrap();
+    //let validator_set3 = ValidatorSet::new(vec![sk1.public(), sk2.public(), sk3.public()]).unwrap();
```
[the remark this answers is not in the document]
pompon0: I forgot to uncomment it after updating BitVec encoding.

node/libs/roles/src/validator/tests.rs
```diff
@@ -200,7 +199,7 @@ fn test_commit_qc() {
 
     // Mismatching validator sets.
     assert!(qc.verify(&validator_set2, 3).is_err());
-    assert!(qc.verify(&validator_set3, 3).is_err());
+    //assert!(qc.verify(&validator_set3, 3).is_err());
```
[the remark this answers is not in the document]
pompon0: uncommented

node/actors/consensus/src/leader/state_machine.rs
pompon0: Done. I did it this way because some people don't like map() with side effects.

node/actors/consensus/src/replica/leader_prepare.rs
pompon0: I don't really care either way, but it's kind of a burden to do it manually. Is there an autoformat option that we could use for that?

node/libs/roles/src/validator/messages/block.rs
pompon0: its very descriptive, what do you not understand? XD

node/actors/sync_blocks/src/peers/tests.rs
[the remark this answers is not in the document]
slowli: Right, will change to strongly typed error then.

node/actors/consensus/src/replica/leader_prepare.rs
slowli: > but it's kind of a burden to do it manually

Not a burden for me, so I'll fix this separately.

node/actors/consensus/src/replica/state_machine.rs
```diff
@@ -96,38 +107,56 @@ impl StateMachine {
...
 
     /// Backups the replica state to disk.
     pub(crate) fn backup_state(&self, ctx: &ctx::Ctx) -> anyhow::Result<()> {
+        let mut proposals = vec![];
+        for (number, payloads) in &self.block_proposal_cache {
+            proposals.extend(payloads.values().map(|p| storage::Proposal {
+                number: *number,
+                payload: p.clone(),
+            }));
+        }
```
slowli, line 153: Nit: You could probably use `flat_map()` and `collect()` on an iterator instead of imperatively extending `proposals`; not that it matters much.

node/actors/consensus/src/leader/replica_prepare.rs
```diff
@@ -1,12 +1,44 @@
...
+#[derive(thiserror::Error, Debug)]
+#[allow(clippy::missing_docs_in_private_items)]
+pub(crate) enum Error {
+    #[error("past view/phase (current view: {current_view:?}, current phase: {current_phase:?})")]
+    Old {
+        current_view: validator::ViewNumber,
+        current_phase: validator::Phase,
+    },
+    #[error("we are not a leader for this message's view")]
+    WhenNotLeaderInView,
```
slowli, line 19: Bikeshedding: IMO, `When` prefix is redundant; `NotLeaderInView` would look better.

node/actors/consensus/src/replica/leader_commit.rs
```diff
@@ -1,9 +1,33 @@
...
+use anyhow::Context as _;
 use concurrency::ctx;
 use roles::validator;
 use tracing::instrument;
 
+#[derive(thiserror::Error, Debug)]
+#[allow(clippy::missing_docs_in_private_items)]
+pub(crate) enum Error {
+    #[error("bad protocol version")]
+    BadProtocolVersion,
```
slowli, line 12: It could make sense to include the expected and observed versions here for debug purposes.

node/libs/roles/src/validator/messages/block.rs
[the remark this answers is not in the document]
brunoffranca: I think we should do it then.

node/libs/roles/src/validator/messages/block.rs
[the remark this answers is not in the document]
brunoffranca: Well, if we include the version in every message, then we don't need to include it in the block header. The block header is never sent by itself over the wire, either it is sent in a leader prepare or in a final block (which would contain the version in the commit QC). It would even make some sense since the protocol version isn't technically a property of the block, but of the nodes.

node/actors/consensus/src/replica/state_machine.rs
pompon0: ack

node/actors/consensus/src/leader/replica_prepare.rs
pompon0: done

node/actors/consensus/src/replica/leader_commit.rs
pompon0: done

slowli (approved):

brunoffranca (approved):

PR #13, opened 2023-10-26
feat: Full node mode

slowli:
# What ❔

Allows to run a node without the consensus module. Refactors the `executor` library to be usable as a library and moves its binary to the `tools` crate.

## Why ❔

We need to be able to use `executor` as a library from the consensus code; including when running an external node.


node/actors/sync_blocks/src/lib.rs
```diff
@@ -27,6 +27,8 @@ mod tests;
 pub use crate::config::Config;
 use crate::peers::PeerStates;
 
+// FIXME(slowli): when run on validator node, the actor creates unnecessary `GetBlocks` requests
```
slowli, line 30: I'd prefer to create a separate PR to fix this.

node/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+        })
+    }
+
+    /// Sets validator-related data for the executor.
+    pub fn set_validator(
+        &mut self,
+        config: ConsensusConfig,
+        key: validator::SecretKey,
+        replica_state_store: Arc<dyn ReplicaStateStore>,
+        blocks_sender: channel::UnboundedSender<FinalBlock>,
```
slowli, line 86: These options could be made more fluent (e.g., if `S` implements `ReplicaStateStore`, it doesn't need to be provided again, and "standard" `blocks_sender`s could be considered (e.g., one that saves a block to the store, or one that does nothing). I've decided against this to make code simpler.

node/tools/src/config/config_paths.rs
```diff
@@ -17,6 +19,15 @@ impl ConfigPaths {
     /// it as an environment variable. If that also fails, we just use a default value.
     #[instrument(level = "trace")]
     pub(crate) fn resolve(args: &[String]) -> Self {
+        let validator_key = PathSpec {
+            name: "Validator key",
+            flag: "--validator-key",
+            env: "VALIDATOR_KEY",
+            default: "validator_key",
+        }
```
slowli, line 27: Is there a particular reason `clap` isn't used for arg parsing here?

pompon0: nobody bothered to refactor this.

node/actors/consensus/src/replica/state_machine.rs
```diff
@@ -27,47 +24,33 @@ pub(crate) struct StateMachine {
...
+    pub(crate) storage: FallbackReplicaStateStore,
 }
 
 impl StateMachine {
     /// Creates a new StateMachine struct. We try to recover a past state from the storage module,
     /// otherwise we initialize the state machine with whatever head block we have.
     pub(crate) async fn new(
         ctx: &ctx::Ctx,
-        storage: Arc<dyn ReplicaStateStore>,
+        storage: FallbackReplicaStateStore,
```
pompon0, line 35: why "fallback"? It is supposed to be the primary source of state, with empty state being the fallback, no?

pompon0: or you mean WithFallback? But then do we need that verbosity?

node/libs/schema/proto/executor/config.proto
```diff
@@ -66,24 +65,32 @@ message GossipConfig {
...
+  // Executor configuration.
+  optional ExecutorConfig executor = 1;
+
+  // IP:port to serve metrics data for scraping.
+  // Use `0.0.0.0:<port>` to listen on all network interfaces.
+  // If not set, metrics data won't be served.
+  optional string metrics_server_addr = 2; // IpAddr
+
+  // Consensus network config.
+  optional ConsensusConfig consensus = 3;
```
pompon0, line 95: while you are at it please annotate all fields in this file as either optional or required (by just suffixing the field definition with a comment: "// required" or "// optional".

node/Cargo.toml
```diff
@@ -26,7 +26,7 @@ assert_matches = "1.5.0"
 async-trait = "0.1.71"
 bit-vec = "0.6"
 blst = "0.3.10"
-clap = { version = "4.3.3", features = ["derive"] }
+clap = { version = "4.3.3", features = ["derive", "env"] }
```
pompon0, line 29: IMO we can drop support for specifying stuff via env vars. We don't use it and we will use eventually the zksync-era style configuration anyway.

node/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+            *public == key.public(),
+            "config.consensus.key = {public:?} doesn't match the secret key {key:?}"
+        );
+        anyhow::ensure!(
+            self.executor_config
+                .validators
+                .iter()
+                .any(|validator_key| validator_key == public),
+            "Key {public:?} is not a validator per validator set {:?}",
+            self.executor_config.validators
```
pompon0, line 99: Having a validator config, even though you are not a validator currently should be a valid setup. Once the validator set is dynamic, a node should be able to join and leave the consensus network. For now it would be enough to just ignore the ConsensusNetwork config if node is not a validator.

node/actors/consensus/src/replica/state_machine.rs
slowli: Yeah, I've meant "store with fallback" (fallback being based on a `BlockStore`). Would something like `FullReplicaStateStore` work better?

pompon0: I guess that would be better. But why not just ReplicaStateStore? The fallback is an implementation detail from the pov of this function. Perhaps we should make FallbackReplicaStateStore into a trait instead of ReplicaStateStore?

pompon0 (dismissed):

brunoffranca (approved):

node/libs/concurrency/src/metrics.rs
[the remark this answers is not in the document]
slowli: The extension traits defined in this module are used in multiple crates (`network` and `consensus`), so IMO it's still useful.

brunoffranca (approved):

pompon0 (approved):

PR #18, opened 2023-10-31
feat: In-memory storage implementation

slowli:
# What ❔

Adds in-memory block and replica state storage implementation. Uses this implementation for tests instead of RocksDB implementation. The RocksDB implementation remains in the `storage` crate, but it's gated behind an off-by-default feature.

## Why ❔

RocksDB involves more dependencies (incl. RocksDB compilation), which could be an overkill, especially for tests.


pompon0 (approved):
nice

PR #20, opened 2023-10-31
fix: Fix race condition in network tests

slowli:
# What ❔

Fixes a race condition in `getting_blocks_from_peers` test.

## Why ❔

Bugs are bad.

### DIFF
--- a/node/actors/network/src/gossip/tests.rs
+++ b/node/actors/network/src/gossip/tests.rs
@@ -505,7 +505,7 @@ async fn getting_blocks_from_peers(node_count: usize, gossip_peers: usize) {
             let (network_pipe, dispatcher_pipe) = pipe::new();
             let (node_stop_sender, node_stop_receiver) = oneshot::channel::<()>();
             s.spawn_bg(async {
-                scope::run!(ctx, |ctx, s| async {
+                let result = scope::run!(ctx, |ctx, s| async {
                     s.spawn_bg(run_network(ctx, node.state.clone(), network_pipe));
                     s.spawn_bg(run_get_block_dispatcher(
                         ctx,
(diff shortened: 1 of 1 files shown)


node/actors/network/src/gossip/tests.rs
```diff
@@ -515,7 +515,14 @@ async fn getting_blocks_from_peers(node_count: usize, gossip_peers: usize) {
                     node_stop_receiver.recv_or_disconnected(ctx).await?.ok();
                     Ok(())
                 })
-                .await
+                .await;
+
+                match result {
+                    Ok(()) => Ok(()),
+                    Err(err) if err.root_cause().is::<ctx::Canceled>() => Ok(()),
```
pompon0, line 522: how do you distinguish a success from a timeout here?

slowli: Good point; these cases are not distinguished rn. How would it best to implement such as check? Obviously, it would be possible to have an atomic / `signal::Once` set after the test logic is completed, but I recall that you didn't like this before. Or should test timeout be removed altogether?

pompon0: I don't know about you, but I usually use timeouts just to make sure that we don't have stupidly long tests (rather than using them for detecting regressions). Therefore I'd remove it and just rely on the nextest config, which already indicates long running tests (we can also tune it to turn the warnings into actual errors).

On the other hand the idea behind the Context is that the methods should be able to tell on their own when the Canceled error is expected and return an Ok when they encounter it. In particular, ideally there never should be a need to inspect the anyhow::Error programmatically (like checking the root_cause() here).

pompon0 (approved):

PR #21, opened 2023-10-31
feat: Minor extensions for server interoperability

slowli:
# What ❔

- Simplifies generating executor configurations.
- Adds byte conversions for `BlockHeaderHash`.
- Makes `SyncBlocks` actor terminate successfully on context cancellation.

## Why ❔

Part of preparations for the integration with the server codebase.

### DIFF
--- a/node/actors/sync_blocks/src/lib.rs
+++ b/node/actors/sync_blocks/src/lib.rs
@@ -13,7 +13,7 @@ use concurrency::{
 };
 use network::io::SyncState;
 use std::sync::Arc;
-use storage::WriteBlockStore;
+use storage::{StorageError, StorageResult, WriteBlockStore};
 use tracing::instrument;
 use utils::pipe::ActorPipe;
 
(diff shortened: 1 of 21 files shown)


node/libs/roles/src/validator/messages/block.rs
```diff
@@ -61,24 +62,38 @@ impl fmt::Display for BlockNumber {
...
+
+    /// Returns a reference to the bytes of this hash.
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        self.0.as_bytes()
+    }
+}
+
 impl TextFmt for BlockHeaderHash {
+    fn decode(text: Text) -> anyhow::Result<Self> {
+        text.strip("block_hash:sha256:")?.decode_hex().map(Self)
```
pompon0, line 84: pls change to "block_header_hash:"

node/actors/sync_blocks/src/lib.rs
```diff
@@ -48,6 +48,10 @@ impl SyncBlocks {
         storage: Arc<dyn WriteBlockStore>,
         config: Config,
     ) -> anyhow::Result<Self> {
+        if config.consensus_threshold == 0 {
+            tracing::warn!("QC validation is disabled; this is not secure");
+        }
```
pompon0, line 53: this introduces non-uniform interpretation of consensus_threshold, as for >0 all signatures are validated, even if there is more signatures than a threshold. IMO it would be better to introduce a separate parameter OR wrap consensus_thereshold into an Option<_>, so that the fact of ignoring the signatures is distinguished.

pompon0 (commented):
Are you sure we need this? Perhaps we can make it just easier to generate well-signed blocks?

slowli:
> Are you sure we need this? Perhaps we can make it just easier to generate well-signed blocks?

I thought that was the approach we agreed upon. For tests / initial PoC it would be easy to generate well-signed blocks, but for production, we'd need to introduce key management and reliably broadcast information about the validator set (I guess using L1 smart contracts). @brunoffranca WDYT?

brunoffranca:
Hey, if you guys think it's easier to introduce well-signed blocks now, then it's fine by me. It seemed simpler to me at the time to just turn off QC validation. But at the time I also thought we were not going to be running the consensus actor at all.

pompon0 (approved):
please remember to address "block_hash:" -> "block_header_hash:"

pompon0 (approved):

pompon0 (approved):

node/libs/roles/src/validator/messages/block.rs
brunoffranca: +1

brunoffranca (approved):

PR #23, opened 2023-11-03
Switch from sha256 to keccak256

moshababo:
Switching from [sha256](https://crates.io/crates/sha2) to [keccak256](https://crates.io/crates/sha3) as the general-purpose hash function, to enhance compatibility with EVM/zkEVM. 

Closing BFT-361.


node/libs/concurrency/src/ctx/rng.rs
```diff
@@ -43,7 +43,7 @@ impl Provider {
     pub(super) fn test() -> Provider {
         Self::Split(
             SplitProvider {
-                builder: sha2::Sha256::new().chain("Lzr81nDW8eSOMH".as_bytes()),
+                builder: sha3::Keccak256::new().chain("Lzr81nDW8eSOMH".as_bytes()),
```
pompon0, line 46: this has nothing to do with evm. Is keccak faster than sha?

pompon0 (approved):

node/libs/concurrency/src/ctx/rng.rs
brunoffranca: Depends. On hardware implementations Keccak/SHA3 is faster, but if there's no hardware support then it's slower than SHA2.
That being said, I just think it's easier for now to just use one hash algorithm for everything. If/when we do code profiling and realize that hashing is a bottleneck, we can change to a faster algorithm like BLAKE3 wherever we can.

node/libs/crypto/src/keccak256/test.rs
```diff
@@ -0,0 +1,33 @@
+#[test]
+fn test_keccak256() -> Result<(), Box<dyn std::error::Error>> {
+    use crate::keccak256::Keccak256;
+
+    // Test vectors obtained from a trusted source
```
brunoffranca, line 5: Maybe specify the source.

brunoffranca (dismissed):

node/libs/crypto/src/keccak256/mod.rs
```diff
@@ -1,17 +1,18 @@
-//! Wrappers for the SHA256 cryptographic hash algorithm.
+//! Wrappers for the Keccak256 cryptographic hash algorithm.
 use crate::ByteFmt;
-use sha2::{digest::Update as _, Digest as _};
+use sha3::{digest::Update as _, Digest as _};
 
+mod test;
 pub mod testonly;
```
slowli, line 6: Off-topic (?): Why does this module need to be public? AFAICT, it doesn't contain public items.

node/deny.toml
```diff
@@ -59,7 +59,7 @@ skip = [
     { name = "regex-syntax", version = "=0.6.29" },
 
     # Old versions required by hyper.
-    { name = "socket2", version = "=0.4.9" },
+    { name = "socket2", version = "=0.4.10" },
```
slowli, line 62: Nit: You can use a caret requirement here (`version = "^0.4"`) to make it more future-proof.

slowli (approved):

node/deny.toml
pompon0: I don't think we want it to be future proof in that sense.

node/libs/crypto/src/keccak256/test.rs
moshababo: Added.

node/libs/crypto/src/keccak256/mod.rs
moshababo: Not sure if you refer to line 5 (added) or 6 (already exists). 
* `testonly`: it needs to be public, although it works without the pub modifier, because visibility is bound to the trait implementation target type. 
* `test`: I forgot to add the `#[cfg(test)]` attribute. Added now.

node/deny.toml
moshababo: Fixed.

brunoffranca (approved):

PR #24, opened 2023-11-03
Upgraded protobuf to support cross crate imports.

pompon0:
Upgraded protobuf to support cross crate imports.
We need it to allow defining protos in zksync-era as well.
Proto packages share a single common namespace, so I prefixed our packages with zksync.<crate name>.
I couldn't name the crate "protobuf", because cargo deny has mistaken it for the public crate with the same name. I've named it "zksync_protobuf", to be consistent with zksync-era repo.

CI protobuf compatibility check is now more complex, because we need to collect the descriptors from the cargo build artifacts. Currently it is implemented as a bash command, but can be upgraded to a proper github action in the future.


node/libs/protobuf_build/src/syntax.rs
```diff
@@ -0,0 +1,244 @@
...
+        elem.0.len() >= self.0.len() && elem.0[0..self.0.len()] == self.0
+    }
+
+    /// Strips a given prefix from the name.
+    pub fn relative_to(&self, prefix: &Self) -> anyhow::Result<Self> {
+        if !prefix.contains(self) {
+            anyhow::bail!(
+                "{} does not contain {}",
+                self.to_string(),
+                prefix.to_string()
```
slowli, line 188: Nit: `to_string()` calls are redundant for types implementing `Display`.

node/libs/protobuf_build/src/syntax.rs
```diff
@@ -0,0 +1,244 @@
...
+/// recommend the following naming convention: proto package "a.b.c" should be defined either:
+/// a) in a single file "a/b/c.proto", or
+/// b) in a collection of files under "a/b/c/" directory
+/// Option b) is useful for defining large packages, because there is no equivalent of "pub use" in proto syntax.
+#[derive(Clone, PartialEq, Eq)]
+pub struct ProtoName(Vec<String>);
+
+impl ProtoName {
+    /// Checks if package path starts with the given prefix.
+    pub fn contains(&self, elem: &Self) -> bool {
```
slowli, line 178: Bikeshedding: IMO, having a method `starts_with` that can be called as `elem.starts_with(_)` would be more expected for such functionality.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+mod ident;
+mod syntax;
+
+/// Traversed all the files in a directory recursively.
+fn traverse_files(
+    path: &Path,
+    f: &mut impl FnMut(&Path) -> anyhow::Result<()>,
+) -> anyhow::Result<()> {
+    if !path.is_dir() {
+        f(path).with_context(|| path.to_str().unwrap().to_string())?;
```
slowli, line 49: Nit: `Path` has a [`display()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) method to conveniently and infallibly convert it to a displayable string

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+        pool.clone()
+    }
+}
+
+/// Code generation config. Use it in build scripts.
+pub struct Config {
+    /// Input directory relative to $CARGO_MANIFEST_DIR with the proto files to be compiled.
+    pub input_root: InputPath,
+    /// Implicit prefix that should be prepended to proto paths of the proto files in the input directory.
+    pub proto_root: ProtoPath,
```
slowli, line 116: Bikeshedding: IMO, here we should adhere to the "explicit is better than implicit" principle and require the Protobuf files always have the full directory structure. (E.g., `schema` files would be put in `proto/zksync/schema/` directory instead of `proto/`.)

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+                    |err| anyhow::anyhow!("{:?}", err),
+                )?);
+            proto_paths.push(path);
+            Ok(())
+        })?;
+
+        // Compile the proto files
+        let mut compiler = protox::Compiler::with_file_resolver(
+            protox::file::DescriptorSetFileResolver::new(pool_raw),
+        );
```
slowli, line 195: Stupid question: is this separate compiler (as opposed to `prost_build`) used just to create a file descriptor set? Does `prost_build` lack the necessary functionality to do so? It allows [to output FDS at the specified path](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.file_descriptor_set_path).

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            .iter()
+            .map(|d| format!("&{}::DESCRIPTOR", d.0))
+            .collect::<Vec<_>>()
+            .join(",");
+        let this = self.this_crate().to_string();
+        output.append(&format!("\
+            pub static DESCRIPTOR : {this}::Lazy<{this}::Descriptor> = {this}::Lazy::new(|| {{\
+                {this}::Descriptor::new({:?}.into(), vec![{rust_deps}], &include_bytes!({descriptor_path:?}))\
+            }});\
+        ",package_root.to_string()));
```
slowli, line 272: This looks fragile and error-prone. Ideally, we'd want to move as much code as possible to the crate itself, probably by using macro (since we need to use `include_bytes` macro), so that it's use in the generated code is as simple as possible, something like

```rust
{this}::declare_descriptor!({descriptor_path:?}, {package_root:?}, {rust_deps});
```

(here, `rust_deps` is a comma-separated list of `path`s [in terms of macros](https://doc.rust-lang.org/beta/reference/macros-by-example.html)).

Even more ideally, we'd want users to declare the descriptor themselves (cf. [`tonic::include_file_descriptor_set!`](https://docs.rs/tonic/latest/tonic/macro.include_file_descriptor_set.html)), but since we'd need to specify `rust_deps` in 2 places, this seems infeasible.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+        /// Global descriptor pool.
+        static POOL: Lazy<Mutex<prost_reflect::DescriptorPool>> = Lazy::new(Mutex::default);
+        let pool = &mut POOL.lock().unwrap();
+        self.load(pool).unwrap();
+        pool.clone()
+    }
+}
+
+/// Code generation config. Use it in build scripts.
+pub struct Config {
```
slowli, line 112: Bikeshedding: IMO, it could make sense to use a builder pattern here (cf. `prost_build::Config`), since manyfields have reasonable defaults. E.g., `input_root` may default to `proto`, `dependencies` to an empty `Vec`, `protobuf_crate` to `::zksync_protobuf`.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            .to_rust_type()
+            .to_string();
+        let proto_name = proto_name.to_string();
+        let this = self.this_crate().to_string();
+        format!("impl {this}::prost_reflect::ReflectMessage for {rust_name} {{\
+            fn descriptor(&self) -> {this}::prost_reflect::MessageDescriptor {{\
+                static INIT : {this}::Lazy<{this}::prost_reflect::MessageDescriptor> = {this}::Lazy::new(|| {{\
+                    DESCRIPTOR.load_global().get_message_by_name({proto_name:?}).unwrap()\
+                }});\
+                INIT.clone()\
```
slowli, line 145: - As with the descriptor definition, ideally we'd want to move as much code from here to the crate itself as possible. Again, this'll probably require using a macro like `derive_reflect_message!($message:ty, $descriptor:ident)`.
- Stupid question: since `DESCRIPTOR` is a static already, does it make sense to cache separate `MessageDescriptor`s here instead of just returning a fresh clone each time? Is it significantly more efficient?

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+/// Code generation config. Use it in build scripts.
+pub struct Config {
+    /// Input directory relative to $CARGO_MANIFEST_DIR with the proto files to be compiled.
+    pub input_root: InputPath,
+    /// Implicit prefix that should be prepended to proto paths of the proto files in the input directory.
+    pub proto_root: ProtoPath,
+    /// Descriptors of the dependencies and the rust absolute paths under which they will be available from the generated code.
+    pub dependencies: Vec<(RustName, &'static Descriptor)>,
```
slowli, line 118: Maybe, a stupid ask: Theoretically, we can add crate name (via `env!("CARGO_CRATE_NAME")`) and module path (via [`module_path!`](https://doc.rust-lang.org/std/macro.module_path.html)) information to `Descriptor` instances, so that each descriptor knows where it's located. In this case, `RustName` could be avoided. (There are caveats, such as renaming the package dependency [using `package = ...`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml).)

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+pub use syntax::*;
+
+mod canonical;
+mod ident;
+mod syntax;
+
+/// Traversed all the files in a directory recursively.
+fn traverse_files(
+    path: &Path,
+    f: &mut impl FnMut(&Path) -> anyhow::Result<()>,
```
slowli, line 46: Nit: Usually, `mut f: impl FnMut(..)` is used in such cases; having to provide a reference to a closure looks weird.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+use prost::Message as _;
+pub use prost_reflect;
+use std::{fs, path::Path, sync::Mutex};
+pub use syntax::*;
+
+mod canonical;
+mod ident;
+mod syntax;
+
+/// Traversed all the files in a directory recursively.
```
slowli, line 43: Typo: Traverse**d** -> Traverse**s**.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+//! buf breaking /tmp/sum.binpb --against /tmp/sum.binpb
+#![allow(clippy::print_stdout)]
+use anyhow::Context as _;
+pub use once_cell::sync::Lazy;
+// Imports accessed from the generated code.
+pub use prost;
+use prost::Message as _;
+pub use prost_reflect;
+use std::{fs, path::Path, sync::Mutex};
+pub use syntax::*;
```
slowli, line 37: Nit: It's beneficial to prefix crate-local imports or re-exports with `crate::` / `self::` to cleanly distinguish them from external imports.

node/actors/network/src/frame.rs
```diff
@@ -1,13 +1,13 @@
-//! Simple frame encoding format (length ++ value) for protobuf messages,
-//! since protobuf messages do not have delimiters.
+//! Simple frame encoding format (length ++ value) for zksync_protobuf messages,
+//! since zksync_protobuf messages do not have delimiters.
```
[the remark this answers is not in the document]
brunoffranca: +1

brunoffranca (changes requested):

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+//! and after.binpb are the concatenated descriptors from those 2 revisions.
+//!
+//! The proto files are not expected to be self-contained - to import proto files from
+//! different crates you need to specify them as dependencies in the Config.dependencies.
+//! It is not possible to depend on a different proto bundle within the same crate (because
+//! these are being built simultaneously from the same build script).
+//!
+//! cargo build --all-targets
+//! perl -ne 'print "$1\n" if /PROTOBUF_DESCRIPTOR="(.*)"/' `find target/debug/build/*/output -type f` | xargs cat > /tmp/sum.binpb
+//! buf breaking /tmp/sum.binpb --against /tmp/sum.binpb
```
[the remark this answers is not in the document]
brunoffranca: +1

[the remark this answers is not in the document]
pompon0: I forgot to remove that

node/libs/protobuf_build/src/lib.rs
pompon0: prost_build doesn't include the compiler. It proxies the compilation to a protoc compiler which has to be included in the system or be imported as a binary. This compiler is more lightweight and more straightforward to instrument.

node/libs/protobuf_build/src/lib.rs
pompon0: I want the user to set all the fields and understand what they do.

node/libs/protobuf_build/src/lib.rs
pompon0: load_global() is inefficient because it checks whether all the files of the descriptor have been already loaded. Plus, cloning prost_reflect::DescriptorPool in load_global() is inefficient because it is poorly written.

node/libs/protobuf_build/src/lib.rs
pompon0: I tried that and it works, but making the path to the proto location public as-is looks very inconvenient to me - more likely the generated proto will be exported via "pub use" statement somewhere. Also, as you say, we need a rename functionality anyway, so making things explicit looks like a better option to me.

node/actors/network/src/frame.rs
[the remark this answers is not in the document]
pompon0: done

node/libs/protobuf_build/src/lib.rs
pompon0: aaand that won't work because I pass the closure recursively to multiple recursive calls.

node/libs/protobuf_build/src/lib.rs
pompon0: done

node/libs/protobuf_build/src/lib.rs
pompon0: done

node/libs/protobuf_build/src/syntax.rs
pompon0: done

node/libs/protobuf_build/src/syntax.rs
pompon0: done

node/libs/protobuf/src/bin/conformance_test/proto/conformance.proto
```diff
@@ -33,10 +33,7 @@
 
 syntax = "proto3";
 
-package conformance;
-
-option java_package = "com.google.protobuf.conformance";
-option objc_class_prefix = "Conformance";
+package zksync.protobuf.conformance_test;
```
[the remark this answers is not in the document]
pompon0: it is the crate name, which I use to separate namespaces.

node/libs/protobuf_build/src/canonical.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Checks whether messages in the given file descriptor set support canonical encoding.
+/// pool should contain all transitive dependencies of files in descriptor.
+pub(crate) fn check(
+    descriptor: &prost_types::FileDescriptorSet,
+    pool: &prost_reflect::DescriptorPool,
+) -> anyhow::Result<()> {
+    for f in &descriptor.file {
+        if f.syntax() != "proto3" {
+            anyhow::bail!("only proto3 syntax is supported");
```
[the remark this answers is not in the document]
pompon0: done

node/libs/protobuf_build/src/canonical.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::syntax::extract_message_names;
+use anyhow::Context as _;
+use std::collections::HashSet;
+
+#[derive(Default)]
+struct Check(HashSet<String>);
+
+impl Check {
+    /// Checks if messages of type `m` support canonical encoding.
+    fn message(&mut self, m: &prost_reflect::MessageDescriptor) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
pompon0: done

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Direct dependencies of this descriptor.
+    dependencies: Vec<&'static Descriptor>,
+}
+
+impl Descriptor {
+    /// Constructs a Descriptor.
+    pub fn new(
+        proto_root: ProtoName,
+        dependencies: Vec<&'static Descriptor>,
+        descriptor_bytes: &impl AsRef<[u8]>,
```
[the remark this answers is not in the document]
pompon0: true

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+                return Ok(());
+            };
+
+            let file_raw = fs::read_to_string(path).context("fs::read()")?;
+            let path = ProtoPath::from_input_path(path, &self.input_root, &self.proto_root);
+            pool_raw
+                .file
+                .push(protox_parse::parse(path.to_str(), &file_raw).map_err(
+                    // rewrapping the error, so that source location is included in the error message.
+                    |err| anyhow::anyhow!("{:?}", err),
```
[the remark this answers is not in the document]
pompon0: done

node/libs/protobuf_build/src/syntax.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub(super) fn to_str(&self) -> &str {
+        self.0.to_str().unwrap()
+    }
+
+    /// Converts the relative input path to an absolute path in the local file system
+    /// (under $CARGO_MANIFEST_DIR).
+    pub(super) fn abs(&self) -> PathBuf {
+        PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap())
+            .canonicalize()
+            .unwrap()
```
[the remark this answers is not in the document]
pompon0: no reason, fixed

brunoffranca (dismissed):

node/libs/protobuf_build/src/lib.rs
pompon0: discussed offline, we will stay with implicit paths for now. If it turns out to be confusing to the users we can expand the paths later.

brunoffranca (approved):

node/libs/protobuf_build/src/lib.rs
pompon0: made it a macro

brunoffranca (dismissed):

node/libs/schema/build.rs
```diff
@@ -1,159 +1,14 @@
...
-    fs::write(proto_output.join("mod.rs"), prettyplease::unparse(&file))?;
-    Ok(())
+//! Generates rust code from protobufs.
+fn main() {
+    zksync_protobuf::build::Config {
+        input_root: "proto".into(),
+        proto_root: "zksync/schema".into(),
+        dependencies: vec![(
+            "::zksync_protobuf::proto".into(),
+            &zksync_protobuf::proto::DESCRIPTOR,
```
slowli, line 8: Having Protobuf dependencies as *build* dependencies is suboptimal since it requires the dependency to be fully built before building this crate even begins. I don't have immediate suggestions how to improve this though.

slowli (commented):
Couple of tricks after I've refreshed my memory on Protobuf handling in Exonum (it was used there as the main serialization format as well):

- It's possible to use the [`quote`](https://docs.rs/quote/latest/quote/) crate to generate Rust code more eloquently. See an example [here](https://github.com/exonum/exonum/blob/master/components/build/src/lib.rs).
- It's possible to pass information from deps to the consumers using [fake linking](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key). E.g., [here](https://github.com/exonum/exonum/blob/2d2fa22e5f5bc451d08c155c2398956f11dce06e/components/merkledb/build.rs#L27) the Merkle tree crate publishes a path to its Protobuf files, which will then be available [as the `DEP_EXONUM_PROTOBUF_MERKLEDB_PROTOS` env variable](https://github.com/exonum/exonum/blob/2d2fa22e5f5bc451d08c155c2398956f11dce06e/components/build/src/lib.rs#L446). The var name is `DEP_$link_$var`, where `link` is a fake crate that the exporting library [links to](https://github.com/exonum/exonum/blob/2d2fa22e5f5bc451d08c155c2398956f11dce06e/components/merkledb/Cargo.toml#L11) (it doesn't need to exist), and `$var` is the name in `println!("cargo:$var={}", ...);`. Multiple vars can be passed using this way; e.g., we could use this to pass paths to Protobuf descriptors to solve the "dependent needs to wait for all direct dependencies to fully compile" issue. (With this approach, it would be improved to "dependent needs to wait for the dependencies build scripts to execute".)

Not sure whether the second trick would be helpful, but using `quote` is IMO certainly preferable to generating Rust code as strings.

node/libs/schema/build.rs
pompon0: I have some ideas, but they will be even more complex than what is here now. I was planning to first check the compilation time impact on zksync-era to see if it is worth the extra effort.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+        for d in &self.dependencies {
+            d.load(pool)?;
+        }
+        pool.add_file_descriptor_set(self.descriptor_proto.clone())?;
+        Ok(())
+    }
+
+    /// Loads the descriptor to the global pool and returns a copy of the global pool.
+    pub fn load_global(&self) -> prost_reflect::DescriptorPool {
```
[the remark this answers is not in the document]
pompon0: done

node/libs/schema/build.rs
slowli: Using exporting variables from the build scripts via fake linking could help; see [my comment above](https://github.com/matter-labs/era-consensus/pull/24#pullrequestreview-1717624969). OTOH, it could be useless, and it creates some complexity on its own (e.g., having to declare a fake `links` field in the crate manifest).

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+    proto_root: ProtoName,
+    /// Raw descriptor proto.
+    descriptor_proto: prost_types::FileDescriptorSet,
+    /// Direct dependencies of this descriptor.
+    dependencies: Vec<&'static Descriptor>,
+}
+
+impl Descriptor {
+    /// Constructs a Descriptor.
+    pub fn new(
```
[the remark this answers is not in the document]
pompon0: ack

node/libs/protobuf_build/src/canonical.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::syntax::extract_message_names;
+use anyhow::Context as _;
+use std::collections::HashSet;
+
+#[derive(Default)]
+struct Check(HashSet<String>);
+
+impl Check {
+    /// Checks if messages of type `m` support canonical encoding.
+    fn check_message(&mut self, m: &prost_reflect::MessageDescriptor) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
pompon0: sure

pompon0:
I've replaced string generation with quote.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// from a message of the given `proto_name`.
+    fn reflect_impl(&self, proto_name: &ProtoName) -> anyhow::Result<String> {
+        let rust_name = proto_name
+            .relative_to(&self.proto_root.to_name().context("invalid proto_root")?)
+            .unwrap()
+            .to_rust_type()
+            .to_string();
+        let rust_name: syn::Path = syn::parse_str(&rust_name).context("rust_name")?;
+        let proto_name = proto_name.to_string();
+        let this: syn::Path = syn::parse_str(&self.this_crate().to_string()).context("this")?;
```
slowli, line 149: Nit: A more conventional way to do this is to use [`parse_quote!`](https://docs.rs/syn/latest/syn/macro.parse_quote.html), or make `RustName` convertible to `syn::Path` (IIUC, `RustName` *is* semantically equivalent to it). With the current `RustName` internal structure, this is hardly possible.

node/libs/protobuf_build/src/syntax.rs
```diff
@@ -0,0 +1,244 @@
...
+                .map(|(name, m)| format!("pub mod {name} {{ {} }}\n", m.collect())),
+        );
+        entries.join("")
+    }
+
+    /// Collects the code of the module and formats it.
+    pub(crate) fn format(&self) -> anyhow::Result<String> {
+        let s = self.collect();
+        Ok(prettyplease::unparse(
+            &syn::parse_str(&s).with_context(|| format!("syn::parse_str({s:?})"))?,
```
slowli, line 164: You can have `RustModule` wrap a `syn::File` internally (in general, dealing with `syn` types rather than strings) in order to get rid of AST -> string -> AST conversions. I'm happy to try doing this in a follow-up PR, so that it the basic functionality is delivered faster.

node/libs/schema/build.rs
```diff
@@ -1,159 +1,14 @@
...
-        m.insert(&name, entry.path());
-    }
-    let file = m.generate();
-    let file = syn::parse_str(&file).unwrap();
-    fs::write(proto_output.join("mod.rs"), prettyplease::unparse(&file))?;
-    Ok(())
+//! Generates rust code from protobufs.
+fn main() {
+    zksync_protobuf::build::Config {
+        input_root: "proto".into(),
```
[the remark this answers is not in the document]
slowli: Not sure I agree with this convention ("hiding" Protobuf files in the Rust source dir look suboptimal), but this can be elaborated later.

slowli (approved):
Looks good enough™ for the first iteration. I'm eager to follow up with minor changes as discussed previously / in line comments (renaming vars + relying on `syn` types instead of on strings where it's appropriate).

PR #25, opened 2023-11-06
Prefixed crate names with zksync_consensus_

pompon0:
Prefixed crate names with zksync_consensus_ to make the names more scope specific.
We cannot afford keeping the generic names we used so far, as this repo will be used in zksync-era now.
The exception is zksync_concurrency (not zksync_consensus_concurrency) which is intended to be extracted from this repo eventually.


node/libs/crypto/src/bn254/tests.rs
```diff
@@ -1,8 +1,6 @@
-use std::iter::repeat_with;
-
-use rand::{rngs::StdRng, Rng, SeedableRng};
-
 use crate::bn254::{AggregateSignature, PublicKey, SecretKey, Signature};
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use std::iter::repeat_with;
```
slowli, line 3: Off-topic: Weird that this incorrect formatting isn't caught in CI.

node/actors/consensus/src/replica/tests.rs
```diff
@@ -1,8 +1,9 @@
 use crate::testonly;
-use concurrency::{ctx, scope, time};
-use network::io::{ConsensusInputMessage, Target};
 use rand::Rng;
-use roles::validator::{self, ViewNumber};
+use zksync_concurrency as concurrency;
```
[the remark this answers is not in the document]
pompon0: done

node/libs/crypto/src/bn254/tests.rs
pompon0: you mean multiple import blocks?

.github/workflows/protobuf_conformance.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           path: "protobuf"
       - uses: mozilla-actions/sccache-action@v0.0.3
       - name: build test
-        run: cargo build -p schema --bin conformance_test
+        run: cargo build -p zksync_consensus_schema --bin conformance_test
```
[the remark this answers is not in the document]
pompon0: I would prefer to minimize the usage of the feature flags if possible. I don't know much about the crate publishing standards though.

brunoffranca (dismissed):

node/actors/consensus/Cargo.toml
```diff
@@ -1,27 +1,26 @@
 [package]
-name = "consensus"
+name = "zksync_consensus_consensus"
```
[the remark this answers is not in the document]
brunoffranca: Agree that the naming looks weird, but I think having inconsistent naming is worse.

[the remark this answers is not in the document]
slowli: Why would it be inconsistent, provided that the crate directory is renamed as well? Not insisting on renaming, just curious. One could argue that `bft` is a more specific suffix than `consensus` because it specifies the kind of consensus in terms of security assumptions (i.e., validators can behave arbitrarily).

node/actors/consensus/src/replica/state_machine.rs
```diff
@@ -1,10 +1,11 @@
 use crate::{metrics, ConsensusInner};
 use anyhow::Context as _;
-use concurrency::{ctx, metrics::LatencyHistogramExt as _, scope, time};
-use roles::validator;
 use std::collections::{BTreeMap, HashMap};
-use storage::{FallbackReplicaStateStore, StorageError};
 use tracing::instrument;
+use zksync_concurrency::{ctx, metrics::LatencyHistogramExt as _, scope, time};
+use zksync_consensus_roles::validator;
+use zksync_consensus_storage as storage;
```
slowli, line 7: Nit: This is used for `Proposal` and `ReplicaState`, which could arguably be imported directly.

node/actors/consensus/src/testonly/run.rs
```diff
@@ -154,7 +155,7 @@ async fn run_nodes(
                 for (_, recv) in network_recv {
                     s.spawn(async {
                         let mut recv = recv;
-                        use network::io;
+                        use zksync_consensus_network::io;
```
slowli, line 158: Off-topic (?): `use` after an executed statement looks weird; it would look more natural to place it before `let`. Also, since `network` is imported into the scope, this change is not strictly necessary 🙃 (IMO, it makes sense still.)

slowli (approved):

node/actors/consensus/src/testonly/run.rs
pompon0: done

node/actors/consensus/src/replica/state_machine.rs
pompon0: I want to keep them prefixed so that it is obvious that these are not from the current crate (and I'm not really fond of renaming the direct imports.

node/actors/consensus/Cargo.toml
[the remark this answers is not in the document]
brunoffranca: If the crate directory is also renamed then I'm in favor. I just didn't want the crate and the directory to have different names.

brunoffranca (approved):

slowli (approved):

PR #27, opened 2023-11-07
fix: Fix extra `GetBlock` requests for validators

slowli:
# What ❔

If the `SyncBlocks` actor runs on a validator node, the node previously produced unnecessary `GetBlock` requests to peers. This is because `SyncBlocks` logic assumed that new blocks are only added within the actor. This PR lifts this assumption; it filters `GetBlock` requests and cancels pending requests reacting to new blocks being put in the storage.

## Why ❔

Extra `GetBlock` requests waste traffic and processing time of peers.


node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -139,6 +158,31 @@ impl PeerStates {
         }
     }
 
+    /// Cancels pending block retrieval for blocks that appear in the storage using other means
+    /// (e.g., thanks to the consensus algorithm). This works at best-effort basis; it's not guaranteed
+    /// that this method will timely cancel all block retrievals.
```
pompon0, line 163: why do we need to compromise here?

brunoffranca (approved):
Looks good. Just to be sure, this only cancels ongoing requests for the next block? It doesn't prevent validators from requesting non-existent blocks to begin with, correct?

slowli:
@brunoffranca WDYM by non-existent blocks? `SyncBlocks` would only request blocks for which there's a proof of existence (either of the block itself or a following block).

node/actors/sync_blocks/src/peers/mod.rs
slowli: That's the most general approach I've come up with that doesn't make many assumptions as to what could update blocks externally and how these updates would look (e.g., whether it's consensus or some other actor). I guess with the knowledge of consensus we could do marginally better by cancelling not only the latest block produced by consensus, but all the preceding blocks as well; is this what you had in mind? If you have any suggestions, I'm happy to listen.

pompon0: I mean you are using watch to signal the received block from external source, which naturally will lead to NOT cancelling some blocks if multiple blocks are received from an external source at the same time. Also AFAICT there is a race condition between spawning a fetching routine for a block and receiving this block from an external source.

brunoffranca:
I misremembered how sync_blocks works. I thought it would always try to fetch block N+1 if we had block N. But you just ask peers for their head number. My question doesn't apply then.

pompon0 (approved):

PR #28, opened 2023-11-08
Switch from ark-bn254 to matterlabs/pairing (BFT-325)

moshababo:
Switching [bn254](https://github.com/matter-labs/era-consensus/pull/11) implementation from [crates/ark-bn254](https://crates.io/crates/ark-bn254) to [matter-labs/pairing](https://github.com/matter-labs/pairing) due to compatibility issues of [zksync-era](https://github.com/matter-labs/zksync-era) with the former. 

### Notes:
* `bn254::tests::byte_fmt_correctness` tests were added, previously not covered

### TODO:
* ~`bn254::tests::aggregate_signature_distinct_messages` is failing, need to figure out why~
* ~`pairing` dependency was redirected to a [fork](https://github.com/moshababo/pairing), due to missing functionality. If no other solution is found, it should be merged to upstream and dependency to be redirected back~
* ~`bn254::testonly`: need to find a solution for `rand::Rng (rand = "0.8.0")` trait object incompatibility with `rand04::Rng (rand = "0.4.6")` trait bound~


node/Cargo.toml
```diff
@@ -57,6 +55,7 @@ prettyplease = "0.2.6"
 pretty_assertions = "1.4.0"
 quick-protobuf = "0.8.1"
 rand = "0.8.0"
+rand04 = "0.1.1"
```
slowli, line 58: Why would you need to rely on a separate crate? You can rename `rand` v0.4 [using something like](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml)

```toml
rand04 = { package = "rand", version = "0.4" }
```

node/deny.toml
```diff
@@ -59,11 +59,14 @@ skip = [
     { name = "regex-syntax", version = "=0.6.29" },
 
     # Old versions required by hyper.
-    { name = "socket2", version = "=0.4.9" },
-    { name = "hashbrown", version = "=0.12.3" }, # (hyper -> h2 -> indexmap -> hashbrown)
+    { name = "socket2", version = "=0.4.10" },
+
+    # Old versions required by pairing_ce & ff_ec.
```
slowli, line 64: Typo: `ff_ec` - > `ff_ce`.

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -1,33 +1,50 @@
...
+        let mut rng = rng08_to_rng04(rng);
+        let p = G1::rand(&mut rng);
+        AggregateSignature(p)
     }
 }
+
+fn rng08_to_rng04<R: Rng + ?Sized>(rng: &mut R) -> rand04StdRng {
+    let val = rng.gen::<usize>();
+    let seed: &[usize] = std::slice::from_ref(&val);
+    rand04::StdRng::from_seed(seed)
```
slowli, line 49: I'd expect to see a wrapper for `rand08::RngCore` that implements `rand04::Rng`; this should be trivial, since the latter requires [a single method](https://docs.rs/rand/0.4.6/rand/trait.Rng.html#tymethod.next_u32), which is [readily available](https://docs.rs/rand/latest/rand/trait.RngCore.html#tymethod.next_u32) in `rand08::RngCore`.

So something like

```rust
struct RngWrapper<R>(R);

impl<R: RngCore> rand04::Rng for RngWrapper<R> {
    // ...
}

// Then, you can use it like
let p = G1::rand(&mut RngWrapper(rng));
```

moshababo: Good idea, thanks!

node/Cargo.toml
moshababo: I went for it because I thought it offers some compatibility, but forgot to revert. 

Fixed.

moshababo:
Pending https://github.com/matter-labs/pairing/pull/11 to be merged.

node/Cargo.toml
```diff
@@ -63,6 +61,7 @@ pretty_assertions = "1.4.0"
 quick-protobuf = "0.8.1"
 quote = "1.0.33"
 rand = "0.8.0"
+rand04 = { package = "rand", version = "0.4" }
```
pompon0, line 64: since you contribute to the fork, can you upgrade rand as well?

node/libs/crypto/src/bn254/testonly.rs
```diff
@@ -1,33 +1,46 @@
...
 use super::{AggregateSignature, PublicKey, SecretKey, Signature};
-use rand::{distributions::Standard, prelude::Distribution, Rng};
+use pairing::bn256::{Fr, G1, G2};
+use rand::{distributions::Standard, prelude::Distribution, Rng, RngCore};
+use rand04::Rand;
+
+struct RngWrapper<R>(R);
+
+impl<R: RngCore> rand04::Rng for RngWrapper<R> {
+    fn next_u32(&mut self) -> u32 {
```
pompon0, line 11: for efficiency implement next_u64 as well

node/Cargo.toml
```diff
@@ -40,10 +40,8 @@ assert_matches = "1.5.0"
 async-trait = "0.1.71"
 bit-vec = "0.6"
 blst = "0.3.10"
-ark-bn254 = "0.4.0"
-ark-ec = "0.4.2"
-ark-serialize = { version = "0.4.2", features = ["std"] }
-num-traits = "0.2.17"
+pairing = { package = "pairing_ce", git = "https://github.com/moshababo/pairing.git" }
```
pompon0, line 43: wouldn't it be easier to push a temporary branch to matterlabs pairing repo? Also please pin to a commit

moshababo: It was a temp solution, already removed. 

Commit is pinned in `Cargo.lock`, is it enough?

node/Cargo.toml
moshababo: If you mean to upgrade `pairing_ce`, I didn't attempt to do so because the underlying `ff_ce` is also using "0.4".

node/libs/crypto/src/bn254/testonly.rs
moshababo: Fixed.

brunoffranca (dismissed):

node/Cargo.toml
brunoffranca: No, cargo.lock will get updated whenever you compile the crate. If you want to pin to a particular commit, you need to do it on cargo.toml.

moshababo: I added the git rev.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -25,52 +27,66 @@ pub mod hash;
...
     }
 }
 
 impl ByteFmt for SecretKey {
     fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
-        Fr::deserialize_compressed(bytes)
-            .map(Self)
-            .context("failed to decode secret key")
+        let mut fr_repr = FrRepr::default();
+        fr_repr.read_be(Cursor::new(bytes))?;
```
slowli, line 51: - Nit: `Cursor` is not strictly necessary here; `std::io::Read` is implemented for `&mut &[u8]` as well (i.e., a shared reference to bytes, where the reference itself can change). Thus, you can declare `mut bytes: &[u8]` and call `fr_repr.read_be(&mut bytes)` and maybe check that all bytes were read.
- Is big-endian encoding canonical for BN254 (i.e., used in other implementations)? Or is there no convention? I'm not entirely sure, but the `ark` implementation looks little-endian, although [the Go one by Consensys](https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bn254) is big-endian.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -25,52 +27,66 @@ pub mod hash;
 mod testonly;
 
 /// Type safety wrapper around a scalar value.
+#[derive(Debug, PartialEq)]
```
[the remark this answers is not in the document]
brunoffranca: +1

node/Cargo.toml
pompon0: ack

node/libs/crypto/src/bn254/mod.rs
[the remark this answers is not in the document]
pompon0: use other private key types in crypto crate as example

pompon0, line 30: Don't implement PartialEq for private key. Always compare corresponding public keys instead.

moshababo: What about this: https://github.com/matter-labs/era-consensus/pull/28/commits/e3074cbee597d0c722dc005d0f4370d22003a14e?

brunoffranca (approved):

node/libs/crypto/src/bn254/mod.rs
moshababo: > - Nit: `Cursor` is not strictly necessary here; `std::io::Read` is implemented for `&mut &[u8]` as well (i.e., a shared reference to bytes, where the reference itself can change). Thus, you can declare `mut bytes: &[u8]` and call `fr_repr.read_be(&mut bytes)` and maybe check that all bytes were read.

Right it's not necessary. Simply removing it seems to be enough. 

> - Is big-endian encoding canonical for BN254 (i.e., used in other implementations)? Or is there no convention? I'm not entirely sure, but the `ark` implementation looks little-endian, although [the Go one by Consensys](https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bn254) is big-endian.

I'm not sure, but I don't think that there is a convention regarding the endianness.

PR #30, opened 2023-11-08
Decomposed schema crate.

pompon0:
Now that cross-crate proto imports are supported, we can move the proto files to crates which are actually using them.


brunoffranca (approved):
At last!

.github/workflows/protobuf.yaml
```diff
@@ -23,12 +23,12 @@ env:
 
 jobs:
   compatibility:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-22.04-github-hosted-16core]
```
slowli, line 26: Just to clarify: Is this change somehow related to decomposing the `schema` crate, or is it a general change that would be performed anyway?

slowli (approved):

PR #32, opened 2023-11-09
feat: Use `package.links` trick for Protobuf compilation

slowli:
# What ❔

Propagates Protobuf information via env variables by relying on [package.links](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key).

## Why ❔

This simplifies Protobuf configuration and eliminate misconfiguration risks.


node/libs/protobuf/src/build.rs
```diff
@@ -0,0 +1,85 @@
...
+/// Global descriptor pool.
+static POOL: Lazy<RwLock<prost_reflect::DescriptorPool>> = Lazy::new(RwLock::default);
+
+/// Protobuf descriptor + info about the mapping to rust code.
+#[derive(Debug)]
+pub struct Descriptor(());
+
+impl Descriptor {
+    /// Constructs a descriptor and adds it to the global pool.
+    pub fn new(_dependencies: &[&'static Self], descriptor_bytes: &[u8]) -> Self {
```
pompon0, line 19: if unused, please remove

node/libs/protobuf_build/src/canonical.rs
```diff
@@ -1,6 +1,7 @@
 //! Checks whether messages in the given file descriptor set support canonical encoding.
 use crate::syntax::extract_message_names;
 use anyhow::Context as _;
+use protox::prost_reflect::{self, prost_types};
```
pompon0, line 4: If we depend on reexports, let's do this consistently. protobuf won't compile if protox::prost_reflect is different than prost_reflect (right?), but if we don't want to import everywhere protox (only in the build scripts) then lets import prost_reflect directly.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -52,70 +56,68 @@ fn traverse_files(
...
-    proto_root: ProtoName,
-    /// Raw descriptor proto.
-    descriptor_proto: prost_types::FileDescriptorSet,
-    /// Direct dependencies of this descriptor.
-    dependencies: Vec<&'static Descriptor>,
+    proto_root: ProtoPath,
+    /// Absolute path to the descriptor.
+    descriptor_path: String,
+    /// Tuples of `proto_root` and absolute paths to the corresponding descriptor for all dependencies
+    /// including transitive ones.
```
pompon0, line 67: including transitive deps is a tradeoff, it should be possible to avoid that.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -52,70 +56,68 @@ fn traverse_files(
...
+    dependencies: Vec<(ProtoPath, String)>,
```
pompon0, line 68: std::path::Path?

node/libs/protobuf_build/src/lib.rs
```diff
@@ -52,70 +56,68 @@ fn traverse_files(
...
-            proto_root,
-            dependencies,
-            descriptor_proto: prost_types::FileDescriptorSet::decode(descriptor_bytes).unwrap(),
-        }
-    }
+impl Manifest {
+    /// Loads manifest from the environment variable.
+    fn from_env(rust_root: &RustName) -> anyhow::Result<Self> {
+        let crate_name = rust_root.crate_name().context("empty `rust_root`")?;
+        let env_name = format!("DEP_{}_PROTO_MANIFEST", crate_name.to_uppercase());
```
pompon0, line 75: wouldn't serde serialization work better than custom serialization here?

node/libs/protobuf_build/src/lib.rs
```diff
@@ -125,11 +127,14 @@ pub struct Config {
     pub input_root: InputPath,
     /// Implicit prefix that should be prepended to proto paths of the proto files in the input directory.
     pub proto_root: ProtoPath,
-    /// Descriptors of the dependencies and the rust absolute paths under which they will be available from the generated code.
-    pub dependencies: Vec<(RustName, &'static Descriptor)>,
-    /// Rust absolute path under which the protobuf crate will be available from the generated
-    /// code.
+    /// Descriptors of the direct dependencies and the rust absolute paths under which they will be available from the generated code.
+    /// Each dependency must be a direct dependency of the built crate. OTOH, it doesn't need to be a build dependency.
+    pub dependencies: Vec<RustName>,
```
pompon0, line 132: this means that from now on the proto module embedding the generated code will have to be publicly available (it cannot be private and reexported). Is that right? Can we do sth about that?

node/libs/protobuf_build/src/lib.rs
pompon0: ok, given that these are just paths to the descriptors and the descriptors themselves are not copied this should be fine

node/libs/protobuf_build/src/lib.rs
```diff
@@ -140,39 +145,93 @@ impl Config {
...
+            #this::impl_reflect_message!(#rust_name, &DESCRIPTOR, #proto_name);
         })
     }
 
-    /// Generates rust code from the proto files according to the config.
-    pub fn generate(&self) -> anyhow::Result<()> {
+    /// Validates this configuration.
+    fn validate(&self) -> anyhow::Result<()> {
+        /// Flag set to `true` if a public compilation target was encountered in a build script.
+        static HAS_PUBLIC_TARGET: AtomicBool = AtomicBool::new(false);
```
pompon0, line 163: Would it be hard to lift this constraint?

node/libs/protobuf_build/src/lib.rs
slowli: I've thought about it as well and considered against it 🙃 It would be possible by writing a manifest into a file somewhere in `$OUT_DIR` and passing around paths to these files. IMO, the downsides of this approach (an additional level of indirection) outweigh the benefits.

node/libs/protobuf_build/src/lib.rs
slowli: I've used the `serde` serialization initially, but decided against it since the stuff we're (de)serializing is essentially a sequence of `(ProtoPath, Path)` tuples (i.e., has a very simple structure).

node/libs/protobuf_build/src/lib.rs
slowli: I'm not sure I understand the question. The location of the original proto module *in the crate* doesn't matter for this purpose; what matters is that descriptors are publicly available (same as previously), and that they are available from the same crate they are generated for. The latter can be restricting; e.g. if `crate_b` re-exports `crate_a` in its entirety and we want to specify the dependency on it as `::crate_b::crate_a::proto`, it won't work. `crate_a` needs to be a direct dependency for `DEP_` env variables to be passed properly. Personally, this doesn't look like a deal breaker (`crate_a` is in the dependency graph of the target crate anyway), but it *is* a limitation.

pompon0: yes, that's the limitation I'm referring to. Also if you reexport crate_a::submodule::proto as crate_a::proto it also won't work.

pompon0: I'm not saying it's a deal breaker, I'm saying that it is a regression.

node/libs/protobuf_build/src/lib.rs
slowli: Theoretically, it can be lifted by changing the exported variable name (`manifest` rn) depending on the compilation target; but I didn't come up with a good way to ID compilation targets and passing these IDs to dependent crates. IMO, restricting a crate to at most one public target makes sense – it makes sure that all public generated code is in a single place. Do you have use cases for multiple public targets in mind?

node/libs/protobuf_build/src/lib.rs
slowli: `PathBuf`, but yeah, this works better 👍 (This is an artifact from `serde` serialization of `Manifest`, btw.)

node/libs/protobuf/src/build.rs
slowli: Having dependency refs [allows ensuring](https://github.com/matter-labs/era-consensus/blob/cd08b9cb49f11dad67c5f73e02c2083fe8a921ae/node/libs/protobuf/src/build.rs#L23) via type system that these dependencies are added in the global pool. Would it be better to force dependencies to get initialized via [`Lazy::force()`](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html#method.force), or in some other way?

slowli:
I should've said from the start: I'm not entirely sure that this approach is better than the one implemented currently.

- (+) It removes requirement to have Protobuf dependencies as build-time deps; this should improve compilation time.
- (+) It also simplifies global descriptor pool management and marginally simplifies Protobuf compilation config.
- (+)The error messages if `package.links` is forgotten or set to a wrong value should be helpful.
- (-) OTOH, it requires that Protobuf dependencies are direct dependencies, and that they are not renamed. The latter requirement could be lifted with some effort; the former one is fundamental because of how `package.links` works.

Global descriptor pool management could theoretically be simplified even further by making `Descriptor`s private. It would work by listing all deps (including transitive ones) when a descriptor is created; this is essentially the paths part to `Manifest`, which we build anyway (and which is topologically sorted by design). A couple of things that bother me about this approach:

- Will the compiler / linker properly deduplicate `include_bytes!` of the same file?
- Ideally, we'd want to not parse the same descriptor bytes into a descriptor multiple times, which would require keeping track of the loaded descriptors.

The current approach sidesteps both of these issues, and ensures topological sorting of descriptors via Rust type system, which IMO is quite cool.

node/libs/protobuf/src/build.rs
pompon0: Type system doesn't ensure any relation between the dependencies and the newly constructed descriptor, so this is as good as Lazy::force (or just dereferencing the lazy descriptors).

node/libs/protobuf_build/src/lib.rs
pompon0: ack

node/libs/protobuf_build/src/lib.rs
pompon0: For example roles crate could have a separate protobuf targets for validator and node roles. I don't see any large benefits though, so it can wait.

pompon0 (dismissed):

brunoffranca (dismissed):

node/libs/protobuf_build/src/lib.rs
slowli: > Also if you reexport crate_a::submodule::proto as crate_a::proto it also won't work.

It should work. I've tested locally that `zksync_protobuf::proto` works if re-exported as `zksync_protobuf::proto3`.

pompon0 (approved):

brunoffranca (approved):

PR #33, opened 2023-11-14
Made generated protobuf structs implement serde

pompon0:
Made generated protobuf structs implement serde.
This is a protobuf json-compatible serde encoding.
This change will allow embedding ProtoFmt structs in zksync-era structs which derive serde traits.


brunoffranca (dismissed):

node/libs/protobuf/src/proto_fmt.rs
```diff
@@ -282,14 +288,22 @@ pub fn encode_json<T: ProtoFmt>(x: &T) -> String {
     encode_json_proto(&x.build())
 }
 
+/// Implementation of serde::Deserialize for arbitrary ReflectMessage.
+pub fn serde_deserialize<'de, T: ReflectMessage + Default, D: serde::Deserializer<'de>>(
+    d: D,
+) -> Result<T, D::Error> {
+    let mut p = T::default();
+    let msg = prost_reflect::DynamicMessage::deserialize(p.descriptor(), d)?;
+    p.merge(msg.encode_to_vec().as_slice()).unwrap();
```
slowli, line 297: Nit: Is there a particular reason not to use [`DynamicMessage::transcode_to()`](https://docs.rs/prost-reflect/latest/prost_reflect/struct.DynamicMessage.html#method.transcode_to) here? It [seems to implement](https://docs.rs/prost-reflect/latest/src/prost_reflect/dynamic/mod.rs.html#576-582) the same logic under the hood.

node/libs/protobuf/src/build.rs
```diff
@@ -81,5 +82,36 @@ macro_rules! impl_reflect_message {
     };
 }
 
+/// Implements `serde::Serialize` compatible with protobuf json encoding.
+#[macro_export]
+macro_rules! impl_serde_serialize {
```
slowli, line 87: Just to understand better: Do we *want* for all generated messages to implement `Serialize` / `Deserialize`, or is it just the necessity because they are embedded into server types? If it's the latter, we could try [something like](https://serde.rs/field-attrs.html#with) `#[serde(with = "zksync_protobuf::serde")]` (i.e., provide a module that implements serialization / deserialization logic; essentially, neatly packaged `serde_serialize` / `serde_deserialize`). In this case, generated messages won't implement `Serialize` / `Deserialize`, but it would be possible to embed them into server types specifying this attribute.

slowli (commented):
What's with the commit history, btw? Not like it matters much since we squash it anyway, but looks like there are unrelated commits.

node/libs/protobuf/src/proto_fmt.rs
pompon0: It's a microoptimization, in a sense that I'm reusing the already existing instance p, which I had to create to call descriptor(). Actually I was thinking about getting rid of ReflectMessage trait altogether in favor of defining our own, so that descriptor can be a class method (not taking &self as an argument).

node/libs/protobuf/src/build.rs
pompon0: I don't really care, but I think it would be more straightforward to force users to use the well defined serialization rather than giving them opportunity to consider having a custom one. In fact the user is expected to define their own struct which maps to the generated struct (by implementing ProtoFmt) and embed THAT struct, which will need to have custom serde::Serialize/Deserialize implementation anyway and in this case there is no actual need for the generated struct to implement serde traits directly at all.

pompon0:
@slowli I've started this pr on top of your pr before it was merged. And with git it is easier to merge than rebase, so I just merged main afterwards.

node/libs/protobuf/src/build.rs
slowli: Sorry, I've realized after making the comment that you're defining `serde` impls for the generated types, but we'd want to embed not them, but rather their wrappers; that's certainly a source of confusion. But still, 

> which will need to have custom serde::Serialize/Deserialize implementation anyway

Not necessarily; it can be embedded into other types with `#[serde(with)]` using the approach I've mentioned. It'd look something like:

```rust
// mod `zksync_protobuf::serde`

use crate::ProtoFmt;
use prost_reflect::{prost::Message, ReflectMessage};
use serde::de::Error as _;

/// Implementation of serde::Serialize for arbitrary ReflectMessage.
pub fn serialize<T: ProtoFmt, S: serde::Serializer>(
    x: &T,
    s: S,
) -> Result<S::Ok, S::Error> {
    let opts = prost_reflect::SerializeOptions::new();
    x.build().transcode_to_dynamic().serialize_with_options(s, &opts)
}


/// Implementation of serde::Deserialize for arbitrary ReflectMessage.
pub fn deserialize<'de, T: ProtoFmt, D: serde::Deserializer<'de>>(
    d: D,
) -> Result<T, D::Error> {
    let mut p = <T::Proto>::default();
    let msg = prost_reflect::DynamicMessage::deserialize(p.descriptor(), d)?;
    p.merge(msg.encode_to_vec().as_slice()).unwrap();
    let p = T::read(&p).map_err(<D::Error>::custom)?;
    Ok(p)
}
```

Example of usage (on the example of `roles` types):

```rust
#[derive(Debug, Deserialize, Serialize)]
struct Test {
    #[serde(with = "zksync_protobuf::serde")]
    key: PublicKey,
    #[serde(with = "zksync_protobuf::serde")]
    signature: AggregateSignature,
}
```

Similarly, you can reinterpret `impl_serde_*` macros, so that they derive the corresponding traits for the type implementing `ProtoFmt` directly, something like

```rust
struct MyType { /* ... */ }

impl ProtoFmt for MyType { /* ... */ }

zksync_protobuf::serde::derive_based_on_proto_fmt!(MyType);
// ^ Derives `Serialize` and `Deserialize` using `serialize` and `deserialize` functions
```

pompon0: done, I've created a serde module with serialize and deserialize methods.

node/Cargo.lock
```diff
@@ -1956,9 +1956,9 @@ dependencies = [
 
 [[package]]
 name = "smallvec"
-version = "1.11.2"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4dccd0940a2dcdf68d092b8cbab7dc0ad8fa938bf95787e1b916b0e3d0e8e970"
+checksum = "942b4a808e05215192e39f4ab80813e599068285906cc91aa64f923db842bd5a"
```
slowli, line 1961: Nit: Some weird crate downgrades in the lockfile; don't seem to be related to the PR. Are they intentional?

slowli (dismissed):

brunoffranca (approved):

node/libs/protobuf/src/serde.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Encodes a generated proto message to json for arbitrary ReflectMessage.
+pub fn encode_json_proto<T: ReflectMessage>(x: &T) -> String {
+    let mut s = serde_json::Serializer::pretty(vec![]);
+    serialize_proto(x, &mut s).unwrap();
+    String::from_utf8(s.into_inner()).unwrap()
+}
+
+/// Encodes a proto message to json for arbitrary ProtoFmt.
+pub fn encode_json<T: ProtoFmt>(x: &T) -> String {
```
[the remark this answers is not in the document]
pompon0: these are just helper wrappers at this point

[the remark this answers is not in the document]
pompon0: removed

node/Cargo.lock
pompon0: they are not intentional, I don't know why my resolved makes this downgrade. I've bumped the tokio version requirement to prevent the downgrade.

slowli (approved):

brunoffranca (approved):

PR #34, opened 2023-11-14
feat: Enable snapshot syncing

slowli:
# What ❔

Changes `SyncBlocks` actor logic so it properly supports syncing from a snapshot (i.e., from block number greater than 0).

## Why ❔

Necessary to support snapshot syncing for full nodes. Also necessary if we decide not to retroactively sign past miniblocks.


node/actors/executor/src/tests.rs
```diff
@@ -24,6 +27,27 @@ async fn store_final_blocks(
 }
 
 impl FullValidatorConfig {
+    fn gen_blocks(&self, count: usize) -> Vec<FinalBlock> {
+        let genesis_block = self.node_config.genesis_block.clone();
+        let validators = &self.node_config.validators;
+        let blocks = iter::successors(Some(genesis_block), |parent| {
+            let payload = Payload(vec![]);
```
pompon0, line 34: how about randomizing payload?

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -223,6 +233,16 @@ impl PeerStates {
                  ({last_contiguous_stored_block}) is lesser than the old value ({prev_contiguous_stored_block})"
             );
         }
+        let prev_first_stored_block = peer_state.first_stored_block;
+        if prev_first_stored_block < first_stored_block {
+            tracing::warn!(
+                %prev_first_stored_block,
+                %first_stored_block,
+                "Bogus state update from peer: first stored block increased; this shouldn't happen"
```
pompon0, line 241: I thought that we would expect peers to GC old blocks (i.e. drop the prefix of the block) rather than backfill them. As I see it, we would expect peers to have the first block specified (number + hash) in the config, so it is not like it would dynamically move backwards then. What is your plan?

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -237,7 +257,7 @@ impl PeerStates {
         Ok(last_contiguous_stored_block)
     }
 
-    fn validate_sync_state(&self, state: SyncState) -> anyhow::Result<BlockNumber> {
+    fn validate_sync_state(&self, state: SyncState) -> anyhow::Result<(BlockNumber, BlockNumber)> {
```
pompon0, line 260: can we have some "block range" type here instead of tuple?

node/actors/executor/src/tests.rs
brunoffranca: +1

node/actors/sync_blocks/src/peers/mod.rs
brunoffranca: +1

node/actors/sync_blocks/src/peers/mod.rs
slowli: Sorry, I don't expect this warning to be long-term, it just corresponds to the fact that *now* there's no way for a node to prune history. So, if a node reports that its history is pruned, it's a reason to suspect a bug. In the longer term, we should at least support nodes pruning their history, and maybe also backfilling history (not quite sure on this; it *could* be the case that snapshot syncing could work that way).

slowli: There was indeed a logic bug here in that `PeerState.first_stored_block` was updated *conditionally*. I've fixed it and added tests for a peer that prunes / backfills history.

node/actors/executor/src/tests.rs
brunoffranca, line 1: I would prefer that we use the term full node rather than external node. The term external node should ideally be deprecated after our full node is finished.

brunoffranca (approved):
Looks good!

pompon0 (approved):

PR #35, opened 2023-11-15
test: Move Protobuf conformance test

slowli:
# What ❔

Moves the Protobuf conformance test binary to an example in the `zksync_protobuf` crate.

## Why ❔

`zksync_protobuf` relies on some "heavy" dependencies (e.g., Tokio, `tracing-subscriber`) just to implement the conformance test.


slowli:
There are a couple of other options which would have the same effect:

- Making `conformance_test` an example in the `zksync_protobuf` crate instead of a binary target. This would allow to have additional deps it brings as dev dependencies (i.e., they would not be included in the dependency graph for crates depending on `zksync_protobuf`).
- Gating `conformance_test` binary behind an opt-in feature. This looks kind of awkward since IMO `conformance_test` is not an integral part of `zksync_protobuf` (it's, well, a test).

pompon0:
I'd prefer to put it to examples then (assuming this doesn't have any nasty side-effects) as this is a test of this specific crate. IMO moving it to a crate with random stuff detaches it too much from the code under test.

brunoffranca (approved):

pompon0 (approved):

PR #37, opened 2023-11-16
Removed Cranky and started using Cargo's lints table

brunoffranca:
# What ❔

Removes the Cranky tool and moves all existing lints to the workspace's Cargo.toml.

## Why ❔

Lint configuration through Cargo was just [stabilized](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo) in Rust, so we no longer need a third-party tool for it.


node/Cargo.toml
```diff
@@ -21,26 +21,75 @@ authors = ["The Matter Labs Team <hello@matterlabs.dev>"]
...
+
+[workspace.lints.clippy]
+# restriction group
+create_dir = "warn"
+empty_structs_with_brackets = "warn"
+float_arithmetic = "warn"
+missing_docs_in_private_items = "warn"
+non_ascii_literal = "warn"
+partial_pub_fields = "warn"
+print_stdout = "warn"
```
pompon0, line 40: nit: can you move the lints stuff to the bottom of the Cargo.toml? It is not that important compared to dependencies specs.

brunoffranca: Sure.

node/Cargo.toml
```diff
@@ -104,3 +109,44 @@ opt-level = 3
...
+manual_string_new = "warn"
+match_bool = "warn"
+wildcard_imports = "warn"
+
+# cargo group
+wildcard_dependencies = "warn"
+
+# Produces too many false positives.
+redundant_locals = "allow"
+needless_pass_by_ref_mut = "allow"
```
slowli, line 152: Nit: missing newline at the end of the file here and in some other files.

node/libs/concurrency/Cargo.toml
```diff
@@ -18,3 +18,43 @@ tokio.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 vise.workspace = true
+
+[lints.rust]
```
slowli, line 22: Do I understand correctly that the only reason not to use `lints.workspace = true` here is to override `unsafe_code` setting? I'd suggest to demote this lint to the `deny` level; then, it could be overridden on the module level.

brunoffranca: You are indeed right. I'm not sure your suggestion works, let me try it.

slowli: I've tried it locally before suggesting 🙃

pompon0 (approved):

pompon0:
The lint that is failing CI seems to be false positive. Please disable the corresponding linting rule.

slowli:
@pompon0 The relevant rule (`redundant_locals`) is disabled; I think the problem is that the default Rust version included into the GitHub runner ATM [is 1.73](https://github.com/actions/runner-images/blob/ubuntu22/20231115.7/images/ubuntu/Ubuntu2204-Readme.md#rust-tools), which doesn't understand the `lint` section in workspace / crate manifests. We'd probably need to have a separate step installing Rust.

brunoffranca:
The Github runner is still running Rust 1.73. We just need to wait until they bump the new version. No need to change anything else.

PR #39, opened 2023-11-20
Add unit tests for `zksync_consensus_bft` (BFT-360)

moshababo:
Adding new unit tests coverage for `zksync_consensus_bft` crate.

### Notes
* Tests are located  at the `leader`/`replica` mods level.
* `testonly::ut_harness::UTHarness` struct was added to encapsulate all testing operations, and moreover to minimize redundancy of test code.

## Coverage
- replica prepare
  - [x] replica_prepare_sanity
  - [x] replica_prepare_sanity_yield_leader_prepare
  - [x] replica_prepare_sanity_yield_leader_prepare_reproposal
  - [x] replica_prepare_old_view
  - [x] replica_prepare_during_commit
  - [x] replica_prepare_not_leader_in_view
  - [x] replica_prepare_already_exists
  - [x] replica_prepare_num_received_below_threshold
  - [x] replica_prepare_invalid_sig
  - [x] replica_prepare_invalid_commit_qc
  - [x] replica_prepare_high_qc_of_current_view
  - [x] replica_prepare_high_qc_of_future_view
  - [x] replica_prepare_non_validator_signer // FAILS

- leader prepare
  - [x] leader_prepare_sanity
  - [x] leader_prepare_reproposal_sanity
  - [x] leader_prepare_sanity_yield_replica_commit
  - [x] leader_prepare_invalid_leader
  - [x] leader_prepare_old_view
  - [x] leader_prepare_invalid_sig
  - [x] leader_prepare_invalid_prepare_qc
  - [x] leader_prepare_invalid_high_qc
  - [x] leader_prepare_proposal_oversized_payload
  - [x] leader_prepare_proposal_mismatched_payload
  - [x] leader_prepare_proposal_when_previous_not_finalized
  - [x] leader_prepare_proposal_invalid_parent_hash
  - [x] leader_prepare_proposal_non_sequential_number
  - [x] leader_prepare_reproposal_without_quorum
  - [x] leader_prepare_reproposal_when_finalized
  - [x] leader_prepare_reproposal_invalid_block
- replica commit
  - [x] replica_commit_sanity
  - [x] replica_commit_sanity_yield_leader_commit
  - [x] replica_commit_old
  - [x] replica_commit_not_leader_in_view
  - [x] replica_commit_already_exists
  - [x] replica_commit_num_received_below_threshold
  - [x] replica_commit_invalid_sig
  - [x] replica_commit_unexpected_proposal
  - [x] replica_commit_protocol_version_mismatch // FAILS
- leader commit
  - [x] leader_commit_sanity
  - [x] leader_commit_sanity_yield_replica_prepare
  - [x] leader_commit_invalid_leader
  - [x] leader_commit_old
  - [x] leader_commit_invalid_sig
  - [x] leader_commit_invalid_commit_qc


brunoffranca (commented):
The direction looks good. I would just ask you to eventually move the utils to the testonly folder.

node/libs/crypto/src/bn254/error.rs
```diff
@@ -1,5 +1,5 @@
 /// Error type for generating and interacting with bn254.
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, PartialEq, thiserror::Error)]
```
slowli, line 2: Since derived traits is a part of the public interface of types, deriving `PartialEq` for errors is usually a bad idea; it could be difficult to maintain when new error variants are added. To match errors in tests, you could use `assert!(matches!(_))`, or the [`assert_matches`](https://docs.rs/assert_matches/latest/assert_matches/) crate (the latter produces more informative panic messages).

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+    pub(crate) struct Util {
+        ctx: Ctx,
+        consensus: Consensus,
+        rng: StdRng,
+        pipe: DispatcherPipe<InputMessage, OutputMessage>,
+        keys: Vec<SecretKey>,
+    }
+
+    impl Util {
+        pub async fn make() -> Util {
```
slowli, line 331: Bikeshedding: I know this convention isn't followed in some existing test methods, but "default" constructors are usually named `new()` in Rust.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+            Util {
+                ctx,
+                consensus,
+                rng,
+                pipe,
+                keys,
+            }
+        }
+
+        pub fn own_key(&mut self) -> &SecretKey {
```
slowli, line 356: You don't need `&mut self` here and in `key_at`; `&self` would suffice.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+            self.consensus.replica.view = view
+        }
+
+        pub fn set_replica_phase(&mut self, phase: Phase) {
+            self.consensus.replica.phase = phase
+        }
+
+        pub fn make_replica_prepare(
+            &mut self,
+            mutate_callback: Option<impl FnOnce(&mut ReplicaPrepare)>,
```
slowli, line 386: This looks awkward; you could require a mandatory callback (`mutate_callback: impl FnOnce(_)`) and provide a no-op (`make_replica_prepare(|_| {})`) when it's not needed.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    impl Util {
+        pub async fn make() -> Util {
+            Util::make_with(1).await
+        }
+
+        pub async fn make_with(num_validators: i32) -> Util {
+            let ctx = ctx::test_root(&ctx::RealClock);
+            let mut rng = StdRng::seed_from_u64(6516565651);
```
slowli, line 337: Why not take the RNG from `ctx`?

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+//         .await
+//         .unwrap();
+//
+//     assert_eq!(
+//         res,
+//         Err(LeaderPrepareError::InvalidPrepareQC(anyhow!("PrepareQC contains messages for different views!")))
+//     );
+// }
+
+mod util {
```
slowli, line 303: Inline modules are not very readable; in [our style guide](https://github.com/matter-labs/era-consensus/blob/main/docs/style.md#codebase-layout), we advise always placing modules in separate files.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+    let leader_prepare = util
+        .own_key()
+        .sign_msg(ConsensusMsg::LeaderPrepare(leader_prepare));
+    let res = util.dispatch_leader_prepare(leader_prepare).await;
+
+    assert_eq!(
+        res,
+        Err(LeaderPrepareError::InvalidPrepareQC(anyhow!(
+            "PrepareQC contains messages for different views!"
+        )))
```
slowli, line 269: Makes little sense to compare errors exactly. I *think* in this case matching an error to `InvalidPrepareQC` would be enough. If it's not enough, the error itself should be refactored, so that this variant wraps a more strongly typed error instead of `anyhow::Error`.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+/// - [ ] leader_commit_invalid_commit_qc_insufficient_signers
+/// - [ ] leader_commit_invalid_commit_qc_invalid_aggregate_sig
+///
+#[tokio::test]
+async fn replica_prepare_sanity() {
+    let mut util = Util::make().await;
+
+    let replica_prepare = util.make_replica_prepare(None::<fn(&mut ReplicaPrepare)>);
+    let res = util.dispatch_replica_prepare(replica_prepare);
+    assert_eq!(res, Ok(()));
```
slowli, line 77: In this case, `unwrap()`ing the result would be enough to check it.

node/actors/bft/src/tests/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod integration_tests;
+mod unit_tests;
```
slowli, line 2: Bikeshedding: I'd shorten module names to `integration` (or `network`) and `unit` (or something like `dispatch`) since they are already have the enclosing `tests` module.

node/actors/bft/src/replica/leader_prepare.rs
```diff
@@ -105,6 +105,12 @@ pub(crate) enum Error {
     Internal(#[from] anyhow::Error),
 }
 
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        self.to_string() == other.to_string()
+    }
+}
```
slowli, line 112: That's a bad idea for the same reasons as for `bn254::Error`. This error type isn't public; OTOH, in this case, unlike for `bn254::Error`, the implementation itself is rather dubious and brittle w.r.t. to refactoring.

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+    use zksync_consensus_roles::{
+        validator,
+        validator::{
+            BlockHeader, ConsensusMsg, LeaderPrepare, Payload, Phase, PrepareQC, ReplicaCommit,
+            ReplicaPrepare, SecretKey, Signed, ViewNumber,
+        },
+    };
+    use zksync_consensus_utils::pipe::DispatcherPipe;
+
+    pub(crate) struct Util {
```
slowli, line 322: IMO, using the helper struct for tests is the right approach, and its interface makes sense 👍 (other than some minor details).

node/actors/bft/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,599 @@
...
+        io::{InputMessage, OutputMessage},
+        leader::ReplicaPrepareError,
+        replica::LeaderPrepareError,
+        Consensus,
+    };
+    use rand::{rngs::StdRng, Rng, SeedableRng};
+    use zksync_concurrency::{ctx, ctx::Ctx, scope};
+    use zksync_consensus_network::io::ConsensusInputMessage;
+    use zksync_consensus_roles::{
+        validator,
```
slowli, line 314: Nit: You can use `vaidator::{self, ..}` below.

node/actors/bft/src/tests/unit_tests.rs
brunoffranca: +1

node/actors/bft/src/tests/unit_tests.rs
brunoffranca: +1

node/actors/bft/src/tests/unit_tests.rs
brunoffranca: or `assert!(res.is_ok())`.

node/actors/bft/src/tests/unit_tests.rs
brunoffranca: But if you are going to move some tests to the `roles` crate anyway, the best is to move from `anyhow` to `thiserror` like you had to do in the `consensus` crate.

node/libs/crypto/src/bn254/error.rs
moshababo: Agree. I refrained from the former due to lack of descriptive panics, but the latter does provides that. 
It's not identical though—it expects a pattern and not an expression for the 2nd leg, but if combined with `assert_eq!` for inner comparisons (assuming that you don't want to just save the expression result in a variable), it continues to provide descriptive panics. 

Changes here: https://github.com/matter-labs/era-consensus/pull/39/commits/ef8cc5f352a9e2243f186cf4d4a69167963d899d

node/actors/bft/src/replica/leader_prepare.rs
moshababo: https://github.com/matter-labs/era-consensus/pull/39#discussion_r1401072512

node/actors/bft/src/tests/unit_tests.rs
moshababo: Fixed.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Fixed.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Yes it turned out to be clumsy especially because type inference for `None` doesn't work. 
Removed.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Fixed.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Following https://github.com/matter-labs/era-consensus/pull/39#pullrequestreview-1739967766 I ended up moving it to `testonly`, although it's not being used outside of the module. 
Another option would be to simply remove the inline module, making the struct part of the enclosing unit tests module.

Change here: https://github.com/matter-labs/era-consensus/pull/39/commits/b489f0f90a45719bb0179e09842fe60c8e602112

node/actors/bft/src/tests/unit_tests.rs
moshababo: Yes, it was just a temp solution. 
Both solutions will be applied.

moshababo: Yes, it was just a temp solution.
Both suggestions will be applied.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Fixed.

node/actors/bft/src/tests/unit_tests.rs
moshababo: Fixed.

node/actors/bft/src/tests/mod.rs
moshababo: Fixed.

moshababo:
The `replica_prepare` and `replica_commit` message processing triggers a panic if the creation of the respective prepare/commit quorum certificate fails. This occurs after completing all checks and updating the local state. While it shouldn't be reproducible, I added two tests that demonstrate it:
* `replica_commit_protocol_version_mismatch`
* `replica_prepare_non_validator_signer`

The first is protected against during top-level processing, but I couldn't find safeguards for the latter.

brunoffranca:
Regarding the first test, we indeed do check the protocol version earlier. But its placement is not ideal, it's separate from all the other checks. And it only produces a tracing warning, it never produces an error that we can actually test. So I would move that check later, so that we do all the checks for the message in the same place.
The second test shows an actual bug, we never check that the replica that is sending us a message belongs to the current validator set, we need to check it. Add that extra check for both the prepare and commit phases and we're good.

brunoffranca:
Another point is if failing to create a QC should panic. Technically we could have that method just return an error, but I'm not sure how we could recover from it. By that time, the only thing we could do is discard all the replica prepare messages we got and let our turn timeout. Which I guess is better than panicking but it would still let an attacker stop the chain anyway.

moshababo:
> Another point is if failing to create a QC should panic. Technically we could have that method just return an error, but I'm not sure how we could recover from it. By that time, the only thing we could do is discard all the replica prepare messages we got and let our turn timeout. Which I guess is better than panicking but it would still let an attacker stop the chain anyway.

Does the current implementation support liveness in the event of a leader's timeout?

moshababo:
Since all shared util code is now located in `testonly` mod, the tests can be moved to `replica`/`leader` mods, closer to the tested code. Any opinions?

pompon0:
> Does the current implementation support liveness in the event of a leader's timeout?

Yes, and we should write tests for that.

> Since all shared util code is now located in testonly mod, the tests can be moved to replica/leader mods, closer to the tested code. Any opinions?

+1, tests should be as close to the code under test as possible.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,300 @@
...
+use zksync_concurrency::{ctx, ctx::Ctx, scope};
+use zksync_consensus_network::io::ConsensusInputMessage;
+use zksync_consensus_roles::validator::{
+    self, BlockHeader, CommitQC, ConsensusMsg, LeaderCommit, LeaderPrepare, Payload, Phase,
+    PrepareQC, ReplicaCommit, ReplicaPrepare, SecretKey, Signed, ViewNumber,
+};
+use zksync_consensus_utils::pipe::DispatcherPipe;
+
+#[cfg(test)]
+pub(crate) struct UTHarness {
```
pompon0, line 17: What UT stands for? Add a comment.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,300 @@
...
+    ctx: Ctx,
+    rng: StdRng,
```
pompon0, line 19: Don't store ctx in struct. Pass it around as an argument.
Don't store rng but rather call ctx.rng() on demand (if it is too expensive to create rng on demand for a given use case, pass it around as an argument).

moshababo: > Don't store ctx in struct. Pass it around as an argument.

I tried to minimize test code flexibility/responsibility when working with this struct, which suppose to be instantiated once per test. Is there a good reason to expose `ctx` in this case? 

> Don't store rng but rather call ctx.rng() on demand (if it is too expensive to create rng on demand for a given use case, pass it around as an argument).

I noticed the docs on using it on-demand, but I thought it’s not relevant for tests, which can also benefit determinism.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,394 @@
...
+use crate::testonly::ut_harness::UTHarness;
+use assert_matches::assert_matches;
+use rand::Rng;
+use zksync_consensus_roles::validator::{
+    self, CommitQC, ConsensusMsg, LeaderCommit, LeaderPrepare, Phase, ProtocolVersion,
+    ReplicaCommit, ViewNumber,
+};
+
+#[tokio::test]
+async fn replica_prepare_sanity() {
```
slowli, line 39: Nit: Cannot this be removed in favor of the following test? Ditto for other "sanity" tests.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,394 @@
...
+
+    util.set_view(view);
+    let qc = util.new_commit_qc(|msg| {
+        msg.view = qc_view;
+    });
+    let replica_prepare = util.new_current_replica_prepare(|msg| msg.high_qc = qc);
+    let res = util.dispatch_replica_prepare(replica_prepare);
+    assert_matches!(
+        res,
+        Err(ReplicaPrepareError::HighQCOfFutureView{ high_qc_view, current_view }) => {
```
slowli, line 182: Nit: Missing space after `HighQCOfFutureView` (should be `HighQCOfFutureView { high_qc_view`).

node/actors/bft/src/leader/mod.rs
```diff
@@ -8,4 +8,8 @@ mod state_machine;
 #[cfg(test)]
 mod tests;
 
+#[allow(unused_imports)]
```
slowli, line 11: Bikeshedding: IMO, using `#[cfg(test)]` on test-specific re-exports would make the intention clearer.

node/actors/bft/src/leader/mod.rs
```diff
@@ -8,4 +8,8 @@ mod state_machine;
...
+pub(crate) use replica_commit::Error as ReplicaCommitError;
```
slowli, line 12: Off-topic (?): IMO, it would make sense to prefix such re-exports with `self::` to clearly distinguish them from imports / re-exports from other crates.

node/actors/bft/src/replica/tests.rs
```diff
@@ -1,54 +1,477 @@
...
+        .msg;
+
+    let high_qc = util.rng().gen();
+    leader_prepare.justification = util.new_prepare_qc(|msg| msg.high_qc = high_qc);
+    let leader_prepare = util
+        .own_key()
+        .sign_msg(ConsensusMsg::LeaderPrepare(leader_prepare));
+
+    let res = util.dispatch_leader_prepare(leader_prepare).await;
+    assert_matches!(res, Err(LeaderPrepareError::InvalidHighQC(_)))
```
slowli, line 191: Nit: Missing `;` at the end of the statement (ditto in some other tests).

node/actors/bft/src/replica/tests.rs
```diff
@@ -1,54 +1,477 @@
...
+    let junk: BlockHeaderHash = util.rng().gen();
+    leader_prepare.proposal.parent = junk;
+    let leader_prepare_signed = util
+        .own_key()
+        .sign_msg(ConsensusMsg::LeaderPrepare(leader_prepare.clone()));
+
+    let res = util.dispatch_leader_prepare(leader_prepare_signed).await;
+    assert_matches!(
+        res,
+        Err(LeaderPrepareError::ProposalInvalidParentHash { correct_parent_hash, received_parent_hash, header }) => {
```
slowli, line 299: Nit: Rust formatter is quite bad at formatting code inside macros. This overly long line could be formatted as

```rust
assert_matches!(
    res,
    Err(LeaderPrepareError::ProposalInvalidParentHash {
        correct_parent_hash,
        received_parent_hash,
        header,
    }) => {
        assert_eq!(correct_parent_hash, replica_prepare.high_vote.proposal.hash());
        assert_eq!(received_parent_hash, junk);
        assert_eq!(header, leader_prepare.proposal);
    }
);
```

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,353 @@
...
+    ) -> Result<(), ReplicaPrepareError> {
+        messages
+            .iter()
+            .enumerate()
+            .map(|(i, msg)| {
+                let signed = keys[i].sign_msg(ConsensusMsg::ReplicaPrepare(msg.clone()));
+                self.dispatch_replica_prepare(signed)
+            })
+            .last()
+            .unwrap()
```
slowli, line 212: Are we indeed only interested in the last `Result` (as opposed to the cumulative one)?

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    #[allow(clippy::result_large_err)]
+    pub(crate) fn dispatch_replica_prepare_many(
+        &mut self,
+        messages: Vec<ReplicaPrepare>,
+        keys: Vec<SecretKey>,
+    ) -> Result<(), ReplicaPrepareError> {
+        messages
+            .iter()
+            .enumerate()
```
slowli, line 206: - `messages` can be consumed via `into_iter()` to avoid cloning later.
- Enumeration can be avoided by `zip`ping `messages` and `keys`; this is considered more idiomatic.
- AFIACT, `keys` are not consumed by the method, so you can pass `&[SecretKey]` as an arg.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,353 @@
...
+                    .await;
+                Ok(res)
+            })
+            .join(ctx)
+        })
+        .await
+        .unwrap()
+    }
+
+    pub(crate) async fn recv_signed(&mut self) -> Option<Signed<ConsensusMsg>> {
```
slowli, line 276: This method seems to always return `Some`; could it be simplified to return `Signed<_>` instead of an `Option`?

node/actors/bft/src/testonly/ut_harness.rs
brunoffranca: `ctx` already has a deterministic test rng. And you are already using the test ctx, so why not use the rng provided?

node/actors/bft/src/leader/tests.rs
moshababo: The following includes it, but I still separated for getting a better regression signals. I guess it can be removed.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,394 @@
...
+    util.set_replica_view(view);
+    util.set_leader_view(view);
+    assert_eq!(util.view_leader(view), util.own_key().public());
+
+    let replica_prepare = util.new_current_replica_prepare(|_| {});
+    let _ = util.dispatch_replica_prepare(replica_prepare.clone());
+    let res = util.dispatch_replica_prepare(replica_prepare.clone());
+    assert_matches!(
+    res,
+    Err(ReplicaPrepareError::Exists { existing_message }) => {
```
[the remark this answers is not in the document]
moshababo: Shouldn't these get fixed by `cargo fmt`? Im using:
```
cargo fmt -- --config imports_granularity=Crate --config group_imports=One
```

Is there another solution besides manual?

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,394 @@
...
+async fn replica_prepare_already_exists() {
+    let mut util = UTHarness::new_with(2).await;
+
+    let view = ViewNumber(2);
+    util.set_replica_view(view);
+    util.set_leader_view(view);
+    assert_eq!(util.view_leader(view), util.own_key().public());
+
+    let replica_prepare = util.new_current_replica_prepare(|_| {});
+    let _ = util.dispatch_replica_prepare(replica_prepare.clone());
```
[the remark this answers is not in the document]
moshababo: This was applied to expected `Err(NumReceivedBelowThreshold)` errors, which occur whenever `num_validators` > 1. 
Initially it was asserted, but I thought it's too verbose, and unrelated to the actual test scenario. 

I will add them back as part of `util.dispatch_replica_prepare_many`.

node/actors/bft/src/leader/mod.rs
moshababo: Yes, I didn't understand why it doesn't detect the import 🤦

node/actors/bft/src/leader/tests.rs
moshababo: Fixed.

Q: https://github.com/matter-labs/era-consensus/pull/39#discussion_r1408074659

node/actors/bft/src/leader/mod.rs
moshababo: Fixed: https://github.com/matter-labs/era-consensus/pull/39/commits/1aee2fecbac8a38f925d49077551a360fd27fa20

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,394 @@
...
+        .own_key()
+        .sign_msg(ConsensusMsg::ReplicaCommit(replica_commit));
+
+    let res = util.dispatch_replica_commit(replica_commit);
+    assert_matches!(
+    res,
+    Err(ReplicaCommitError::Old { current_view, current_phase }) => {
+        assert_eq!(current_view, util.current_replica_view());
+        assert_eq!(current_phase, util.current_replica_phase());
+    });
```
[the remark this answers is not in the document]
moshababo: Fixed.

Q: https://github.com/matter-labs/era-consensus/pull/39#discussion_r1408074659

node/actors/bft/src/replica/tests.rs
moshababo: Fixed.

node/actors/bft/src/replica/tests.rs
moshababo: Fixed.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,398 @@
...
+//! - [x] replica_commit_sanity
+//! - [x] replica_commit_sanity_yield_leader_commit
+//! - [x] replica_commit_old
+//! - [x] replica_commit_not_leader_in_view
+//! - [x] replica_commit_already_exists
+//! - [x] replica_commit_num_received_below_threshold
+//! - [x] replica_commit_invalid_sig
+//! - [x] replica_commit_unexpected_proposal
+//! - [x] replica_commit_protocol_version_mismatch // FAILS
+//!
```
brunoffranca, line 25: You can delete these comments.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,398 @@
...
+async fn replica_prepare_sanity_yield_leader_prepare() {
+    let mut util = UTHarness::new().await;
+
+    let replica_prepare = util.new_current_replica_prepare(|_| {});
+    util.dispatch_replica_prepare(replica_prepare).unwrap();
+    util.recv_signed()
+        .await
+        .unwrap()
+        .cast::<LeaderPrepare>()
+        .unwrap();
```
brunoffranca, line 59: We could check some fields on the `LeaderPrepare` message.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,398 @@
...
+}
+
+#[tokio::test]
+async fn replica_prepare_old_view() {
+    let mut util = UTHarness::new().await;
+
+    util.set_replica_view(ViewNumber(1));
+    util.set_leader_view(ViewNumber(2));
+    util.set_leader_phase(Phase::Prepare);
```
brunoffranca, line 68:
```suggestion
    util.set_leader_view(ViewNumber(2));
```
Unnecessary calls, I think.

node/actors/bft/src/leader/tests.rs
```diff
@@ -1,39 +1,398 @@
...
+    let mut util = UTHarness::new().await;
+
+    let replica_commit = util.new_procedural_replica_commit().await;
+    util.dispatch_replica_commit(replica_commit).unwrap();
+    util.recv_signed()
+        .await
+        .unwrap()
+        .cast::<LeaderCommit>()
+        .unwrap();
+}
```
brunoffranca, line 253: Sanity tests need to use more replicas. Using just one replica doesn't test the threshold well. Imagine the leader only creates a response when all the replicas answer, or when just one replica answer. We need to have 6 replicas at least, see that the leader creates a response with 5 replica messages.

node/actors/bft/src/replica/tests.rs
```diff
@@ -1,54 +1,477 @@
...
+//! - [x] leader_prepare_reproposal_when_finalized
+//! - [x] leader_prepare_reproposal_invalid_block
+//!
+//! - [x] leader_commit_sanity
+//! - [x] leader_commit_sanity_yield_replica_prepare
+//! - [x] leader_commit_invalid_leader
+//! - [x] leader_commit_old
+//! - [x] leader_commit_invalid_sig
+//! - [x] leader_commit_invalid_commit_qc
+//!
```
brunoffranca, line 25:
```suggestion
```

node/actors/bft/src/replica/tests.rs
```diff
@@ -1,54 +1,477 @@
...
+    let mut util = UTHarness::new().await;
+
+    let leader_prepare = util.new_procedural_leader_prepare().await;
+    util.dispatch_leader_prepare(leader_prepare).await.unwrap();
+    util.recv_signed()
+        .await
+        .unwrap()
+        .cast::<ReplicaCommit>()
+        .unwrap();
+}
```
brunoffranca, line 62: You want to add sanity tests for reproposals too. Now we are just testing new proposals.

node/actors/bft/src/leader/tests.rs
brunoffranca: Seems redundant, yes. No need to separate them.

node/actors/bft/src/leader/tests.rs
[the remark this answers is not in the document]
brunoffranca: It should work. If cargo fmt doesn't fix it automatically then don't bother with it (unless you really want to).

node/actors/bft/src/testonly/ut_harness.rs
moshababo: > - `messages` can be consumed via `into_iter()` to avoid cloning later.

👍
Fixed.
> - Enumeration can be avoided by `zip`ping `messages` and `keys`; this is considered more idiomatic.

👍
Fixed.

> - AFIACT, `keys` are not consumed by the method, so you can pass `&[SecretKey]` as an arg.

I refrained from using a reference because it overlaps with mutable borrowing in `util.dispatch_replica_prepare_many`. I didn't find a reasonable way to overcome this:

`ut_harness.rs`
```rust
pub(crate) fn keys(&self) -> Vec<&SecretKey> {
    self.keys.iter().collect()
}
``` 
`tests.rs`
```rust
let keys = util.keys();
let res = util.dispatch_replica_prepare_many(
    vec![replica_prepare; consensus_threshold(num_validators)],
    keys,
);
```
```rust
error[E0502]: cannot borrow `util` as mutable because it is also borrowed as immutable
   --> actors/bft/src/replica/tests.rs:364:15
    |
363 |       let keys = util.keys();
    |                  ---- immutable borrow occurs here
364 |       let res = util.dispatch_replica_prepare_many(
    |                 ^    ----------------------------- immutable borrow later used by call
    |  _______________|
    | |
365 | |         vec![replica_prepare; consensus_threshold(num_validators)],
366 | |         keys,
367 | |     );
    | |_____^ mutable borrow occurs here
```

node/actors/bft/src/testonly/ut_harness.rs
moshababo: Yes, originally it had `None` which turned out to be unnecessary, but the method signature remained. 

I changed it now to return `Result` which will propagate the error type from `self.pipe.recv` instead of just unwrapping it: https://github.com/matter-labs/era-consensus/pull/39/commits/aa85e36a11d004eae9ceea493a43bd1861574ae8

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,353 @@
...
+    pipe: DispatcherPipe<InputMessage, OutputMessage>,
+    keys: Vec<SecretKey>,
+}
+
+impl UTHarness {
+    pub(crate) async fn new() -> UTHarness {
+        UTHarness::new_with(1).await
+    }
+
+    pub(crate) async fn new_with(num_validators: i32) -> UTHarness {
```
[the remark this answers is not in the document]
moshababo: I changed it to `usize`, following:
```rust
fn faulty_replicas(num_validators: usize) -> usize
```

node/actors/bft/src/leader/tests.rs
moshababo: Removed.

node/actors/bft/src/replica/tests.rs
moshababo: Removed.

node/actors/bft/src/leader/tests.rs
moshababo: These are indeed the default value being set, but it might change (at least, the view number), and might not be self evident to everyone.

node/actors/bft/src/leader/tests.rs
moshababo: Added fields assersions for all 4 types: https://github.com/matter-labs/era-consensus/pull/39/commits/1c06852f9468e587327c2db43555ef1004e45649

node/actors/bft/src/leader/tests.rs
moshababo: I refactored sanity tests of the 4 types to use a larger validator set. Also, added assertions for every distinct msg processing from the set. 

Changes here: https://github.com/matter-labs/era-consensus/pull/39/commits/9cd6e59ef7fb080634d1f86d2d22f65a2c2c8095

node/actors/bft/src/leader/tests.rs
moshababo: I ended up using them differently: the first is for testing with a significant validator set size, the latter is for asserting the yielded message fields.

node/actors/bft/src/testonly/ut_harness.rs
moshababo: Only the last one is expected to succeed. In the new version I added assertions for the intermediate ones expected error.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -0,0 +1,353 @@
...
+            .secret_key
+            .sign_msg(ConsensusMsg::LeaderCommit(msg))
+    }
+
+    pub(crate) async fn new_procedural_leader_prepare(&mut self) -> Signed<ConsensusMsg> {
+        let replica_prepare = self.new_current_replica_prepare(|_| {});
+        self.dispatch_replica_prepare(replica_prepare.clone())
+            .unwrap();
+        self.recv_signed().await.unwrap()
+    }
```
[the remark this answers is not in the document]
moshababo: I added `_one` & `_many` suffix variants to the relevant methods, to avoid this confusion.

node/actors/bft/src/leader/tests.rs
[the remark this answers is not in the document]
slowli: `cargo fmt` is traditionally bad with code inside macros.

node/actors/bft/src/replica/tests.rs
moshababo: Added here: https://github.com/matter-labs/era-consensus/pull/39/commits/d02ee90607d30b45cec52d11d5a007be2e23d7e0

* `replica_prepare_sanity_yield_leader_prepare_reproposal` is focused on asserting the produced `leader_prepare` msg fields.
* `leader_prepare_reproposal_sanity` only checks that `replica_prepare` processing produced a reproposal, and then verifies a successful `leader_prepare` processing of it by the replica. 

I did not add a test for a reproposal with insufficiant quorum because it is already covered by `leader_prepare_reproposal_without_quorum`.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -89,12 +119,23 @@ impl UTHarness {
         self.consensus.replica.view = view
     }
 
+    pub(crate) fn new_unfinalized_replica_prepare(&self) -> Signed<ConsensusMsg> {
+        self.new_current_replica_prepare(|msg| {
+            msg.high_qc = self.new_commit_qc(|high_qc| {
+                // Set view to the previous view since it needs to belong to an earlier view.
+                high_qc.view = msg.view.prev();
+                // Increment high_qc proposal number to create a discrepancy between the high_qc and the high_vote.
+                high_qc.proposal.number = high_qc.proposal.number.next();
```
brunoffranca, line 128: This is a very unnatural situation for a replica to be in. It must have voted for some block N on view M while not being aware of other blocks N and N+1 having been finalized on previous views. And it's impossible for a quorum of replicas to be in this situation.
It's preferable if the high vote here is for a view and block number greater than the high QC. Like this:

```suggestion
            let mut high_vote = ReplicaCommit {
                protocol_version: validator::ProtocolVersion::EARLIEST,
                view: self.consensus.replica.view.next(),
                proposal: self.consensus.replica.high_qc.message.proposal,
            };

            high_vote.proposal.parent = high_vote.proposal.hash();
            high_vote.proposal.number = high_vote.proposal.number.next();

            msg.high_vote = high_vote;
```

brunoffranca (approved):
Looks good. There's only one change that needs to be committed. Approved!

PR #40, opened 2023-11-23
feat: Validate genesis block in executor config

slowli:
## What ❔

Validate genesis block self-consistency and correspondence to storage when an `Executor` is created.

## Why ❔

Allows not to accidentally break a node due to misconfiguration.

### DIFF
--- a/node/actors/executor/src/lib.rs
+++ b/node/actors/executor/src/lib.rs
@@ -61,6 +61,7 @@ impl<S: WriteBlockStore + 'static> Executor<S> {
         node_key: node::SecretKey,
         storage: Arc<S>,
     ) -> anyhow::Result<Self> {
+        node_config.validate()?;
         anyhow::ensure!(
             node_config.gossip.key == node_key.public(),
             "config.gossip.key = {:?} doesn't match the secret key {:?}",
--- a/node/actors/executor/src/lib.rs
+++ b/node/actors/executor/src/lib.rs
@@ -2,12 +2,12 @@
 
 use crate::io::Dispatcher;
 use anyhow::Context as _;
-use std::sync::Arc;
+use std::{any, sync::Arc};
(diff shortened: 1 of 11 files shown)


node/actors/executor/src/lib.rs
```diff
@@ -56,18 +56,30 @@ pub struct Executor<S> {
...
         anyhow::ensure!(
             node_config.gossip.key == node_key.public(),
             "config.gossip.key = {:?} doesn't match the secret key {:?}",
             node_config.gossip.key,
             node_key
         );
 
+        // While justifications may differ among nodes for an arbitrary block, we assume that
+        // the genesis block has a hardcoded justification.
+        let first_block = storage.first_block(ctx).await.context("first_block")?;
```
slowli, line 75: I'm not entirely sure whether it's valid for a storage to contain blocks before the genesis block. IMO, it makes reasoning about the storage more difficult, so the current check disallows it.

node/actors/executor/src/tests.rs
```diff
@@ -31,15 +31,68 @@ impl FullValidatorConfig {
...
-    fn into_executor(self, storage: Arc<InMemoryStorage>) -> Executor<InMemoryStorage> {
-        let mut executor = Executor::new(self.node_config, self.node_key, storage.clone()).unwrap();
+    async fn into_executor(
+        self,
+        ctx: &ctx::Ctx,
+        storage: Arc<InMemoryStorage>,
+    ) -> Executor<InMemoryStorage> {
+        let mut executor = Executor::new(ctx, self.node_config, self.node_key, storage.clone())
+            .await
+            .unwrap();
```
pompon0, line 41: is this unwrap safe?

pompon0 (approved):

brunoffranca (approved):

PR #41, opened 2023-11-28
feat: Allow configuring protocol version

slowli:
## What ❔

Allows configuring protocol version for consensus messages instead of hard-coding it.

## Why ❔

We intend to use the same protocol version as for the server, which makes hard-coding it not feasible for consensus–server integration.

### DIFF
--- a/node/actors/executor/src/config/mod.rs
+++ b/node/actors/executor/src/config/mod.rs
@@ -6,7 +6,7 @@ use std::{
 };
 use zksync_consensus_bft::misc::consensus_threshold;
 use zksync_consensus_crypto::{read_required_text, Text, TextFmt};
-use zksync_consensus_network::{consensus, gossip};
+use zksync_consensus_network::gossip;
 use zksync_consensus_roles::{node, validator};
 use zksync_protobuf::{read_required, required, ProtoFmt};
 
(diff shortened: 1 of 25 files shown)


node/actors/executor/src/config/proto/mod.proto
```diff
@@ -48,6 +48,8 @@ message NodeAddr {
 message ConsensusConfig {
   optional string key = 1; // [required] ValidatorPublicKey
   optional string public_addr = 2; // [required] IpAddr
+  // Currently used protocol version for consensus messages.
+  optional uint32 protocol_version = 3;
```
pompon0, line 52: [required]?

node/actors/executor/src/config/mod.rs
```diff
@@ -23,31 +23,28 @@ pub struct ConsensusConfig {
...
-    fn read(r: &Self::Proto) -> anyhow::Result<Self> {
+    fn read(proto: &Self::Proto) -> anyhow::Result<Self> {
+        let protocol_version = proto.protocol_version.context("protocol_version")?;
         Ok(Self {
-            key: read_required_text(&r.key).context("key")?,
-            public_addr: read_required_text(&r.public_addr).context("public_addr")?,
+            key: read_required_text(&proto.key).context("key")?,
+            public_addr: read_required_text(&proto.public_addr).context("public_addr")?,
+            protocol_version: validator::ProtocolVersion::try_from(protocol_version)
+                .context("protocol_version")?,
```
pompon0, line 39: `required(&proto.protocol_version).then(ProtocolVersion::try_from).context("protocol_version")?`

brunoffranca (approved):

PR #42, opened 2023-11-28
traits for proposing/verifying payload

pompon0:
We need a way to delegate proposing/verifying the block's payload to the user.
See the related pr: https://github.com/matter-labs/zksync-era/pull/554

Additionally introduced stronger typing for signed messages in bft tests.

Fixes BFT-388.


node/actors/executor/src/lib.rs
```diff
@@ -20,14 +19,15 @@ mod tests;
...
-#[derive(Debug)]
 struct ValidatorExecutor {
     /// Consensus network configuration.
     config: ConsensusConfig,
     /// Validator key.
     key: validator::SecretKey,
     /// Store for replica state.
     replica_state_store: Arc<dyn ReplicaStateStore>,
+    /// Payload proposer for new blocks.
+    payload_source: Arc<dyn PayloadSource>,
```
slowli, line 30: Is there a particular reason you didn't include `fmt::Debug` as a supertrait for `PayloadSource`? It would allow to retain `Debug` impls for `ValidatorExecutor` / `Executor` and not change `unwrap_err()` to `err().unwrap()` in tests. I'm not sure to which extent we'll copy our code style for published crates from the standard library, but I'd strongly advocate for [debuggability](https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug) (all public types implementing `Debug`) in any case.

node/actors/bft/src/leader/state_machine.rs
```diff
@@ -1,17 +1,19 @@
...
+use zksync_concurrency::{ctx, error::Wrap as _, metrics::LatencyHistogramExt as _, time};
 use zksync_consensus_roles::validator;
 
 /// The StateMachine struct contains the state of the leader. This is a simple state machine. We just store
 /// replica messages and produce leader messages (including proposing blocks) when we reach the threshold for
 /// those messages. When participating in consensus we are not the leader most of the time.
-#[derive(Debug)]
 pub(crate) struct StateMachine {
+    /// Payload provider for the new blocks.
+    pub(crate) payload_source: Arc<dyn PayloadSource>,
```
slowli, line 16: Dumb question: Is using `Arc` instead of `Box` and `&self` receiver on `propose` instead of `&mut self` motivated by some external reasoning? Or is it just chosen defensively? Within the repo, both can be changed without a problem. IMO, unlike storage, there's no inherent expectation that a proposer is shared.

node/libs/storage/src/traits.rs
```diff
@@ -45,53 +45,31 @@ pub trait BlockStore: fmt::Debug + Send + Sync {
     fn subscribe_to_block_writes(&self) -> watch::Receiver<BlockNumber>;
 }
 
-#[async_trait]
```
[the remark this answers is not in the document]
pompon0: I don't see why we should treat Arc specially here (i.e. why not Box, Rc, &, &mut, etc.). Also I lost a significant amount of time debugging why my implementation of verify_payload on X is overridden by the default implementation on Arc<X> (kind of my fault, but I'm not taking the full blame here ;P). As you say, it can always be wrapped in a newtype anyway.

node/actors/bft/src/lib.rs
```diff
@@ -64,13 +75,13 @@ impl Consensus {
...
             replica: replica::StateMachine::new(ctx, storage).await?,
-            leader: leader::StateMachine::new(ctx),
+            leader: leader::StateMachine::new(ctx, payload_source),
         })
     }
 
     /// Starts the Consensus actor. It will start running, processing incoming messages and
     /// sending output messages. This is a blocking method.
-    #[instrument(level = "trace", ret)]
+    #[instrument(level = "trace", skip(self) ret)]
```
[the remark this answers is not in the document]
pompon0: I don't know about ret vs err, I just added skip(self).

node/libs/storage/src/traits.rs
```diff
@@ -45,53 +45,31 @@ pub trait BlockStore: fmt::Debug + Send + Sync {
...
-}
-
 /// Mutable storage of L2 blocks.
 ///
-/// Implementations **must** propagate context cancellation using [`StorageError::Canceled`].
+/// Implementations **must** propagate context cancellation using [`ctx::Error::Canceled`].
 #[async_trait]
 pub trait WriteBlockStore: BlockStore {
+    /// Verify that `payload` is a correct proposal for the block `block_number`.
+    async fn verify_payload(
```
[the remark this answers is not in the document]
pompon0: I need it for tests. I can move it to InMemoryStorage and RocksDB storage directly if you prefer.

node/actors/bft/src/leader/state_machine.rs
pompon0: Well, ptal https://github.com/matter-labs/zksync-era/pull/554 . Here storage is payload source as well. I guess I can make it `Box<Arc<Storage>>` there, if you want.

node/libs/storage/src/traits.rs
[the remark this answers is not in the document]
pompon0: done

node/actors/bft/src/replica/leader_prepare.rs
```diff
@@ -266,6 +269,18 @@ impl StateMachine {
                         header: message.proposal,
                     });
                 }
+
+                // Payload should be valid.
+                if let Err(err) = self
+                    .storage
+                    .verify_payload(ctx, message.proposal.number, payload)
+                    .await
```
brunoffranca, line 277: Why delegate the verification to the storage? I think it would make more sense to have this method as part of the PayloadSource trait.

pompon0: PayloadSource is consumed by leader, verify_payload is consumed by replica. verify_payload() should be thought of a dryrun of put_block(), hence the placement. We expect the blockchains to be stateful, so verify_payload() will need to have access to the state (storage) anyway.

pompon0: I'm not saying that this API perfect though. It's just the best version I've come up with for now.

brunoffranca (approved):

node/actors/bft/src/leader/state_machine.rs
slowli: Sorry, I've looked at this PR first. No need to box an `Arc`, although you could theoretically split off `PayloadSource` implementation to a different struct than the one implementing `BlockStore` (it'd essentially wrap a `ConnectionPool`).

node/actors/executor/src/lib.rs
slowli: > I don't see how unwrap_err() is better than err().unwrap()

It allows to inspect the `Ok` variant contents, which improves debuggability. Likewise, having a fuller view into `ValidatorExecutor` improves debuggability as well (which is not public, but is present as a field in a public `Executor` type). IMO, having an `fmt::Debug` constraint is not a burden; and it's already a supertrait for adjacent storage traits. OTOH, that's a minor thing, and I'm not in position to insist 🤷

node/actors/bft/src/testonly/ut_harness.rs
slowli, line 1: Just to understand better: The changes in tests are not directly related to the new traits, it just happened that they are in the same PR, right?

slowli (approved):

node/actors/executor/src/lib.rs
pompon0: ack

node/actors/bft/src/leader/state_machine.rs
pompon0: ack

PR #46, opened 2023-12-01
Add liveness tests (BFT-387)

moshababo:
Adding new protocol liveness tests for when the network becomes idle in various stages:

- [x] `timeout_leader_no_prepares` - when leader have no cached prepare messages for the current view
- [x] `timeout_leader_some_prepares` - when leader have some cached prepare messages for the current view
- [x] `timeout_leader_in_commit` - when leader is in commit phase
- [x] `timeout_replica_in_commit` - when replica is in commit phase
- [x] `timeout_leader_some_commits` - when leader have some cached commit messages for the current view
- [x] `timeout_leader_in_consecutive_prepare` - when leader is in a consecutive prepare phase


node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -382,6 +405,22 @@ impl UTHarness {
         .unwrap()
     }
 
+    pub(crate) async fn sim_timeout(&mut self) {
+        scope::run!(&self.ctx, |ctx, s| {
```
pompon0, line 409: why do you wrap it in scope::run?

node/actors/bft/src/tests.rs
```diff
@@ -62,3 +65,163 @@ async fn byzantine_mock_network() {
...
+    let base_leader_view = util.leader_view();
+
+    util.sim_timeout().await;
+
+    assert_eq!(util.replica_view(), base_replica_view.next());
+    assert_eq!(util.replica_phase(), Phase::Prepare);
+    assert_eq!(util.leader_view(), base_leader_view);
+    assert_eq!(util.leader_phase(), Phase::Prepare);
+
+    util.check_recovery_after_timeout();
```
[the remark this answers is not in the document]
moshababo: I made some changes in tests and in `UTHarness::check_recovery_after_timeout` to support simulating a new block. 

The actual block is simulated by a full iteration, implemented in `UTHarness::iterate_next`. LMK if you recommend a different way for this step. 

Changes here: https://github.com/matter-labs/era-consensus/pull/46/commits/8bdfdba925550be4515d2552ab8f44da762b0d4b

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -406,19 +426,11 @@ impl UTHarness {
     }
 
     pub(crate) async fn sim_timeout(&mut self) {
-        scope::run!(&self.ctx, |ctx, s| {
```
pompon0: ditto for other calls you have wrapped in scope::run

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -71,26 +71,46 @@ impl UTHarness {
...
+            .unwrap();
+    }
 
-        assert_eq!(self.replica_phase(), Phase::Prepare);
+    pub(crate) async fn check_recovery_after_timeout(
+        &mut self,
+        view: ViewNumber,
+        high_vote_view: ViewNumber,
+        high_qc_view: ViewNumber,
+    ) {
```
pompon0, line 89: Personally I don't understand this function's signature. What are the arguments for? Can you make it simpler? And/or please document when exactly should it be called.

pompon0:
fyi, all these `cast()` calls are not really necessary. I'll prepare a pr to make those test utilities strongly typed (it will also make the tests easier to read).

node/actors/bft/src/testonly/ut_harness.rs
moshababo: As discussed it will be cleaned in a follow-up PR.

node/actors/bft/src/testonly/ut_harness.rs
moshababo: Docs added.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -67,6 +68,57 @@ impl UTHarness {
...
+                view: _view,
+                high_vote,
+                high_qc,
+                ..
+            } => {
+                assert_eq!(_view, view);
+                assert_eq!(high_vote.view, high_vote_view);
+                assert_eq!(high_qc.message.view, high_qc_view);
+            }
+        );
```
pompon0, line 116: nit: this assert_matches is unnecesary: you can just write the assert_eq! statements directly.

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -83,8 +135,8 @@ impl UTHarness {
         &self.consensus.inner.secret_key
     }
 
-    pub(crate) fn owner_as_view_leader(&self) -> ViewNumber {
-        let mut view = self.current_replica_view();
+    pub(crate) fn owner_as_view_leader_next(&self) -> ViewNumber {
```
pompon0, line 138: nit: IMO the "next" suffix seems to suggest at the very least that the returned view will be AFTER the current view.

pompon0 (approved):

node/actors/bft/src/testonly/ut_harness.rs
moshababo: I wanted to imply that it's not _beyond_ the next one. 
But perhaps there will be a use case where the current is to be skipped, so I made it explicit with a longer name: `owner_as_view_leader_current_or_next`.

PR #51, opened 2023-12-13
Construct quorum certificate incrementally

moshababo:
Constructing quorum certificates from a batch of signed messages creates redundant error checks which are hard to reason about. 

Since this operation's failure is unacceptable and can only occur as a result of a bug, it is preferred to construct incrementally upon receiving each message.

However, this refactoring ended up with no error checks during the incremental construction, delegating responsibility to the message processing.

Consequently, this change can be viewed merely as code cleanup which removes redundant error checks, and the incremental steps as a safer way to do that.

### Notes
* `PrepareQCBuilder` &`CommitQCBuilder` structs were added on top of the underlying structs
* Batch-mode functions throughout the stack are no longer used in non-test code, and were moved to `testonly`. They can be removed altogether if tests will construct incrementally as well.
    * `PrepareQC::from`
    * `CommitQC::from` (also seemed to had a duplicated error check which was removed)
    * `validator::AggregateSignature::aggregate`
    * `bn254::AggregateSignature::aggregate`

### TODOs (pending concept ACK)
* Refactor tests to construct incrementally (?)
* `PrepareQCBuilder::add`'s `BitVec` handling can be simplified
* `CommitQCBuilder::take` does not consume the builder in the same way as `PrepareQCBuilder::take` because `CommitQC` does not have a clear default value. This can be improved

### DIFF
--- a/node/actors/bft/src/leader/state_machine.rs
+++ b/node/actors/bft/src/leader/state_machine.rs
@@ -6,7 +6,10 @@ use std::{
 };
 use tracing::instrument;
 use zksync_concurrency::{ctx, error::Wrap as _, metrics::LatencyHistogramExt as _, time};
-use zksync_consensus_roles::validator;
+use zksync_consensus_roles::{
+    validator,
+    validator::{CommitQCBuilder, PrepareQCBuilder},
+};
 
 /// The StateMachine struct contains the state of the leader. This is a simple state machine. We just store
 /// replica messages and produce leader messages (including proposing blocks) when we reach the threshold for
@@ -29,11 +32,15 @@ pub(crate) struct StateMachine {
         validator::ViewNumber,
         HashMap<validator::PublicKey, validator::Signed<validator::ReplicaPrepare>>,
     >,
+    /// Prepare QC builders indexed by view number.
+    pub(crate) prepare_qc: BTreeMap<validator::ViewNumber, PrepareQCBuilder>,
     /// A cache of replica commit messages indexed by view number and validator.
     pub(crate) commit_message_cache: BTreeMap<
         validator::ViewNumber,
         HashMap<validator::PublicKey, validator::Signed<validator::ReplicaCommit>>,
     >,
+    /// Commit QC builders indexed by view number.
+    pub(crate) commit_qc: BTreeMap<validator::ViewNumber, CommitQCBuilder>,
 }
 
 impl StateMachine {
@@ -47,7 +54,9 @@ impl StateMachine {
             phase_start: ctx.now(),
             block_proposal_cache: None,
             prepare_message_cache: BTreeMap::new(),
+            prepare_qc: BTreeMap::new(),
             commit_message_cache: BTreeMap::new(),
+            commit_qc: BTreeMap::new(),
         }
     }
 
--- a/node/libs/roles/src/validator/messages/consensus.rs
+++ b/node/libs/roles/src/validator/messages/consensus.rs
@@ -1,10 +1,13 @@
 //! Messages related to the consensus protocol.
 
 use super::{BlockHeader, Msg, Payload, Signed};
-use crate::validator;
-use anyhow::{bail, Context};
+use crate::{validator, validator::Signature};
+use anyhow::bail;
 use bit_vec::BitVec;
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    mem,
+};
 use zksync_consensus_utils::enum_util::{BadVariantError, Variant};
 
 /// Version of the consensus algorithm that the validator is using.
@@ -169,10 +172,53 @@ pub struct LeaderCommit {
     pub justification: CommitQC,
 }
 
+/// Utility for incrementally constructing a `PrepareQC` instance .
+pub struct PrepareQCBuilder {
+    instance: PrepareQC,
+    val_set_size: usize,
+}
+
+impl PrepareQCBuilder {
+    /// Create a new builder instance for a given validator set size.
+    pub fn new(val_set_size: usize) -> Self {
+        Self {
+            instance: Default::default(),
+            val_set_size,
+        }
+    }
+
+    /// Add a validator's signed message to the `PrepareQC`.
+    pub fn add(&mut self, signed_message: &Signed<ReplicaPrepare>, validator_index: usize) {
+        // TODO: refactor to cleaner code
+        if self.instance.map.contains_key(&signed_message.msg) {
+            self.instance
+                .map
+                .get_mut(&signed_message.msg)
+                .unwrap()
+                .0
+                .set(validator_index, true);
+        } else {
+            let mut bit_vec = BitVec::from_elem(self.val_set_size, false);
+            bit_vec.set(validator_index, true);
+
+            self.instance
+                .map
+                .insert(signed_message.msg.clone(), Signers(bit_vec));
+        }
+
+        self.instance.signature.add(&signed_message.sig);
+    }
+
+    /// Take the constructed `PrepareQC`, consuming the builder in the process.
+    pub fn take(&mut self) -> PrepareQC {
+        mem::take(&mut self.instance)
+    }
+}
+
 /// A quorum certificate of replica Prepare messages. Since not all Prepare messages are
 /// identical (they have different high blocks and high QCs), we need to keep the high blocks
 /// and high QCs in a map. We can still aggregate the signatures though.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
 pub struct PrepareQC {
     /// Map from replica Prepare messages to the validators that signed them.
     pub map: BTreeMap<ReplicaPrepare, Signers>,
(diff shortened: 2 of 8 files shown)


node/actors/bft/src/leader/state_machine.rs
```diff
@@ -29,11 +32,15 @@ pub(crate) struct StateMachine {
...
+    /// Prepare QC builders indexed by view number.
+    pub(crate) prepare_qc: BTreeMap<validator::ViewNumber, PrepareQCBuilder>,
     /// A cache of replica commit messages indexed by view number and validator.
     pub(crate) commit_message_cache: BTreeMap<
         validator::ViewNumber,
         HashMap<validator::PublicKey, validator::Signed<validator::ReplicaCommit>>,
     >,
+    /// Commit QC builders indexed by view number.
+    pub(crate) commit_qc: BTreeMap<validator::ViewNumber, CommitQCBuilder>,
 }
```
brunoffranca, line 44: Note that you no longer need prepare or commit message caches.

brunoffranca (commented):
Overall, looks good. A small nit, I don't see the need to have a separate Builder type. A QC is still a QC, even if it doesn't have enough signers.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -169,10 +172,53 @@ pub struct LeaderCommit {
...
+            instance: Default::default(),
+            val_set_size,
+        }
+    }
+
+    /// Add a validator's signed message to the `PrepareQC`.
+    pub fn add(&mut self, signed_message: &Signed<ReplicaPrepare>, validator_index: usize) {
+        // TODO: refactor to cleaner code
+        if self.instance.map.contains_key(&signed_message.msg) {
+            self.instance
```
pompon0, line 194: shouldn't message signature verification happen here, so that the builder actually knows that it is building the correct thing?

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -169,10 +172,53 @@ pub struct LeaderCommit {
     pub justification: CommitQC,
 }
 
+/// Utility for incrementally constructing a `PrepareQC` instance .
+pub struct PrepareQCBuilder {
+    instance: PrepareQC,
+    val_set_size: usize,
```
pompon0, line 178: I suggest we actually keep an Arc<ValidatorSet> in the builder.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: Message signature was already verified prior to aggregating. Is there a reason why that should not suffice?

node/actors/bft/src/leader/state_machine.rs
moshababo: Right, but it’s still used to check if we got a secondary message from a validator, which isn’t easily extractable from the certificate. 

Obviously a better solution can be made other than storing all messages, but I think we should settle on that only after the transition to non-blocking processing (BFT-406).

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -169,10 +172,53 @@ pub struct LeaderCommit {
...
+            self.instance
+                .map
+                .insert(signed_message.msg.clone(), Signers(bit_vec));
+        }
+
+        self.instance.signature.add(&signed_message.sig);
+    }
+
+    /// Take the constructed `PrepareQC`, consuming the builder in the process.
+    pub fn take(&mut self) -> PrepareQC {
```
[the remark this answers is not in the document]
moshababo: I ended up removing the builder altogether.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: I ended up removing the builder altogether.

moshababo:
> A small nit, I don't see the need to have a separate Builder type. A QC is still a QC, even if it doesn't have enough signers.

Agree, I added the higher-level structs due to having additional fields. But I ended up removing all of them, delegating responsibility for correctness to the caller.

node/libs/roles/src/validator/messages/consensus.rs
pompon0: I meant to move verification here, not do it twice.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -227,6 +181,32 @@ pub struct PrepareQC {
...
+            .unwrap_or(ViewNumber(0))
+    }
+
+    /// Add a validator's signed message.
+    /// * `signed_message` - A signed message.
+    /// * `validator_index` - A tuple containing the index of the validator and the total size of the set.
+    pub fn add(
+        &mut self,
+        signed_message: &Signed<ReplicaPrepare>,
+        validator_index: (usize, usize),
```
pompon0, line 199: Why not &ValidatorSet? `(usize,usize)` is a very poor choice for an interface.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -332,6 +283,21 @@ pub struct CommitQC {
 }
 
 impl CommitQC {
+    /// Create a new empty instance for a given `ReplicaCommit` message and a validator set size.
+    pub fn new(message: ReplicaCommit, validator_set_size: usize) -> Self {
```
pompon0, line 287: &ValidatorSet

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -332,6 +283,21 @@ pub struct CommitQC {
...
+        Self {
+            message,
+            signers: Signers(BitVec::from_elem(validator_set_size, false)),
+            signature: validator::AggregateSignature::default(),
+        }
+    }
+
+    /// Add a validator's signature.
+    pub fn add(&mut self, sig: &Signature, validator_index: usize) {
```
pompon0, line 296: &ValidatorSet

node/libs/roles/src/validator/messages/consensus.rs
brunoffranca: +1

brunoffranca (approved):

node/libs/roles/src/validator/messages/consensus.rs
moshababo: I took the thin builder approach, delegating responsibility for correctness to the caller, to be used after all checks are done. Otherwise, there would be error checks duplications like previously, or confusion regarding who’s checking what. 

Also IMO signature should be verified even earlier than now.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: Was I overly excited to finally use built-in tuples syntax? 😁

Fixed.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: Fixed.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: This one is the already-known index value, so I left as is.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -190,46 +190,20 @@ impl PrepareQC {
...
-        }
-
-        // Aggregate the signatures.
-        let signature =
-            validator::AggregateSignature::aggregate(signed_messages.iter().map(|v| &v.sig));
+    /// Add a validator's signed message.
+    pub fn add(
+        &mut self,
+        signed_message: &Signed<ReplicaPrepare>,
+        validator_index: usize,
```
pompon0, line 197: this is implied by the validator_set.index(signed_message.key)

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -190,46 +190,20 @@ impl PrepareQC {
...
+        validator_set: &ValidatorSet,
+    ) {
+        self.map
+            .entry(signed_message.msg.clone())
+            .or_insert_with(|| Signers(BitVec::from_elem(validator_set.len(), false)))
+            .0
+            .set(validator_index, true);
```
pompon0, line 204: early exit if the signature for the given validator has been already added.

node/libs/roles/src/validator/testonly.rs
```diff
@@ -62,6 +63,78 @@ pub fn make_block<R: Rng>(
...
+
+            prepare_qc.add(signed_message, index, validators);
+        }
+
+        Ok(prepare_qc)
+    }
+}
+
+impl CommitQC {
+    /// Creates a new CommitQC from a list of *signed* replica Commit messages and the current validator set.
```
pompon0, line 111: ... panics if signed_messages is empty.

node/libs/roles/src/validator/messages/consensus.rs
pompon0: ack, please then annotate that the function expects the signature to be already verified.

node/libs/roles/src/validator/testonly.rs
pompon0: ... expects signatures to be already verified.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: I added it just because it was already retrieved and error-handled by the caller (for a different use case).

node/libs/roles/src/validator/messages/consensus.rs
moshababo: This wasn’t checked for in the previous version, and is rather the responsibility of the caller, which indeed checks it. 
So I’m not sure why to add it here, considering the thin/dumb builder approach, which doesn’t suppose to enforce the protocol rules.

node/libs/roles/src/validator/testonly.rs
moshababo: Fixed.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: Added.

pompon0 (approved):

PR #52, opened 2023-12-13
made bft actor send LeaderPrepare messages asynchronously 

pompon0:
## What ❔
Made bft actor send LeaderPrepare messages asynchronously.

## Why ❔
PayloadSource.propose() may be blocking (when no payload can be proposed at a time) but it shouldn't block message processing:
* other validators may be able to propose, so we should keep exchanging messages with them.
* validator should avoid keeping a growing backlog of unprocessed messages

### DIFF
--- a/node/actors/bft/src/leader/replica_commit.rs
+++ b/node/actors/bft/src/leader/replica_commit.rs
@@ -24,8 +24,8 @@ pub(crate) enum Error {
         signer: validator::PublicKey,
     },
     /// Unexpected proposal.
-    #[error("unexpected proposal")]
-    UnexpectedProposal,
+    //#[error("unexpected proposal")]
+    //UnexpectedProposal,
     /// Past view or phase.
     #[error("past view/phase (current view: {current_view:?}, current phase: {current_phase:?})")]
     Old {
@@ -44,7 +44,7 @@ pub(crate) enum Error {
         existing_message: validator::ReplicaCommit,
     },
     /// Number of received messages is below threshold.
-    #[error(
+    /*#[error(
         "number of received messages is below threshold. waiting for more (received: {num_messages:?}, \
          threshold: {threshold:?}"
     )]
@@ -53,7 +53,7 @@ pub(crate) enum Error {
         num_messages: usize,
         /// Threshold for message count.
         threshold: usize,
-    },
+    },*/
     /// Invalid message signature.
     #[error("invalid signature: {0:#}")]
     InvalidSignature(#[source] validator::Error),
@@ -117,14 +117,6 @@ impl StateMachine {
         // Check the signature on the message.
         signed_message.verify().map_err(Error::InvalidSignature)?;
 
-        // ----------- Checking the contents of the message --------------
-
-        // We only accept replica commit messages for proposals that we have cached. That's so
-        // we don't need to store replica commit messages for different proposals.
-        if self.block_proposal_cache != Some(message.proposal) {
-            return Err(Error::UnexpectedProposal);
-        }
-
         // ----------- All checks finished. Now we process the message. --------------
 
         // We store the message in our cache.
@@ -135,10 +127,10 @@ impl StateMachine {
 
         // Now we check if we have enough messages to continue.
         let mut by_proposal: HashMap<_, Vec<_>> = HashMap::new();
-        for msg in self.commit_message_cache.get(&message.view).unwrap().iter() {
-            by_proposal.entry(msg.proposal.clone()).or_default().push(msg);
+        for msg in self.commit_message_cache.get(&message.view).unwrap().values() {
+            by_proposal.entry(msg.msg.proposal.clone()).or_default().push(msg);
         }
-        let Some((proposal,replica_messages)) = by_proposal.into_iter().find(|(k,v)|v.len()>=self.inner.threshold()) else {
+        let Some((_,replica_messages)) = by_proposal.into_iter().find(|(_,v)|v.len()>=self.inner.threshold()) else {
             return Ok(());
         };
         debug_assert!(replica_messages.len() == self.inner.threshold());
@@ -157,7 +149,7 @@ impl StateMachine {
 
         // Create the justification for our message.
         let justification = validator::CommitQC::from(
-            &replica_messages.iter().cloned().collect::<Vec<_>>(),
+            &replica_messages.into_iter().cloned().collect::<Vec<_>>()[..],
             &self.inner.validator_set
         ).expect("Couldn't create justification from valid replica messages!");
 
--- a/node/actors/bft/src/leader/replica_commit.rs
+++ b/node/actors/bft/src/leader/replica_commit.rs
@@ -1,5 +1,5 @@
 use super::StateMachine;
-use crate::{inner::ConsensusInner, metrics, Consensus};
+use crate::{metrics};
 use tracing::instrument;
 use zksync_concurrency::{ctx, metrics::LatencyHistogramExt as _};
 use zksync_consensus_network::io::{ConsensusInputMessage, Target};
@@ -72,10 +72,10 @@ impl StateMachine {
         let author = &signed_message.key;
 
         // Check protocol version compatibility.
-        if !Consensus::PROTOCOL_VERSION.compatible(&message.protocol_version) {
+        if !crate::PROTOCOL_VERSION.compatible(&message.protocol_version) {
             return Err(Error::IncompatibleProtocolVersion {
                 message_version: message.protocol_version,
-                local_version: Consensus::PROTOCOL_VERSION,
+                local_version: crate::PROTOCOL_VERSION,
             });
         }
 
@@ -177,7 +177,7 @@ impl StateMachine {
                 .secret_key
                 .sign_msg(validator::ConsensusMsg::LeaderCommit(
                     validator::LeaderCommit {
-                        protocol_version: Consensus::PROTOCOL_VERSION,
+                        protocol_version: crate::PROTOCOL_VERSION,
                         justification,
                     },
                 )),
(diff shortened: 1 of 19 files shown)


node/actors/bft/src/leader/state_machine.rs
```diff
@@ -93,4 +93,104 @@ impl StateMachine {
...
+        for (vote, signers) in justification.map.iter() {
+            *count.entry(vote.high_vote.proposal).or_default() += signers.len();
+        }
+
+        let highest_vote: Option<validator::BlockHeader> = count
+            .iter()
+            // We only take one value from the iterator because there can only be at most one block with a quorum of 2f+1 votes.
+            .find(|(_, v)| **v > 2 * inner.faulty_replicas())
+            .map(|(h, _)| h)
+            .cloned();
```
slowli, line 141: Nit: This can be shortened to a single `Iterator` combining method:

```rust
.find_map(|(header, v)| {
    (*v > 2 * inner.faulty_replicas()).then_some(*header)
});
```

node/actors/bft/src/leader/state_machine.rs
```diff
@@ -93,4 +93,104 @@ impl StateMachine {
...
+        ctx: &ctx::Ctx,
+        inner: &ConsensusInner,
+        payload_source: &dyn PayloadSource,
+        justification: validator::PrepareQC,
+    ) -> ctx::Result<()> {
+        // Get the highest block voted for and check if there's a quorum of votes for it. To have a quorum
+        // in this situation, we require 2*f+1 votes, where f is the maximum number of faulty replicas.
+        let mut count: HashMap<_, usize> = HashMap::new();
+
+        for (vote, signers) in justification.map.iter() {
```
slowli, line 132: Nit: You can use `in &justification.map` instead of calling `iter()` (shorter and considered more idiomatic).

node/actors/bft/src/leader/tests.rs
```diff
@@ -133,16 +133,19 @@ async fn replica_prepare_during_commit() {
...
     let res = util.process_replica_prepare(ctx, replica_prepare).await;
     assert_matches!(
         res,
         Err(ReplicaPrepareError::Old {
-            current_view: ViewNumber(1),
+            current_view,
             current_phase: Phase::Commit,
-        })
+        }) => {
+            assert_eq!(current_view,util.replica.view);
```
slowli, line 147: Nit: Missing space after `,`.

node/actors/bft/src/testonly/make.rs
```diff
@@ -26,30 +20,19 @@ impl PayloadSource for RandomPayloadSource {
...
-    key: &validator::SecretKey,
-    validator_set: &validator::ValidatorSet,
-    genesis_block: &validator::FinalBlock,
-) -> (Consensus, DispatcherPipe<InputMessage, OutputMessage>) {
-    // Initialize the storage.
-    let storage = InMemoryStorage::new(genesis_block.clone());
-    // Create the pipe.
-    let (consensus_pipe, dispatcher_pipe) = pipe::new();
+/// Never provides a payload.
+pub struct InavailablePayloadSource;
```
slowli, line 24: Spelling: Isn't it "unavailable"?

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -274,23 +246,21 @@ impl UTHarness {
...
-                key.sign_msg(msg),
-            );
-            match (i + 1).cmp(&self.consensus_threshold()) {
+            let res = self.leader.process_replica_commit(ctx, key.sign_msg(msg));
+            let want_threshold = self.replica.inner.threshold();
+            match (i + 1).cmp(&want_threshold) {
                 Ordering::Equal => res.unwrap(),
-                Ordering::Less => assert_matches!(
+                Ordering::Less => res.unwrap(),
+                /*assert_matches!(
```
slowli, line 254: Should this commented out assertion be removed?

node/actors/bft/src/leader/state_machine.rs
```diff
@@ -93,4 +93,104 @@ impl StateMachine {
         metrics::METRICS.leader_processing_latency[&label].observe_latency(ctx.now() - now);
         Ok(())
     }
+
+    /// In a loop, receives a PrepareQC and sends a LeaderPrepare containing it.
+    /// Every subsequent PrepareQC has to be for a higher view than the previous one (otherwise it
```
slowli, line 98: Is it "higher" or "higher or equal"? Unless I'm missing something, the code below implements "higher or equal" check (i.e., the same view would pass the check). Not sure whether this matters, but it's at least a documentation nit 🙃

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -191,83 +171,19 @@ impl StateMachine {
...
-                    .await?;
-                metrics::METRICS
-                    .leader_proposal_payload_size
-                    .observe(payload.0.len());
-                let proposal =
-                    validator::BlockHeader::new(&highest_qc.message.proposal, payload.hash());
-                (proposal, Some(payload))
-            }
-        };
+        debug_assert!(num_messages == self.inner.threshold());
```
slowli, line 174: Nit: You can use `debug_assert_eq!` here.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -134,16 +112,22 @@ impl StateMachine {
...
-            return Err(Error::NumReceivedBelowThreshold {
-                num_messages,
-                threshold: consensus.threshold(),
-            });
+        let mut by_proposal: HashMap<_, Vec<_>> = HashMap::new();
+        for msg in self
+            .commit_message_cache
+            .get(&message.view)
+            .unwrap()
+            .values()
```
slowli, line 120: Nit: IIUC, this could be simplified by extracting an updated cache entry into a var:

```rust
let cache_entry = self.commit_message_cache
    .entry(message.view)
    .or_default();
cache_entry.insert(author.clone(), signed_message);

// Now we check if we have enough messages to continue.
let mut by_proposal: HashMap<_, Vec<_>> = HashMap::new();
for msg in cache_entry.values() {
    by_proposal.entry(msg.msg.proposal).or_default().push(msg);
}
```

slowli (commented):
Can't say I'm a particularly good reviewer for consensus logic, but overall the changes look solid.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -134,16 +112,22 @@ impl StateMachine {
...
         }
-
-        debug_assert!(num_messages == consensus.threshold());
+        let Some((_, replica_messages)) = by_proposal
+            .into_iter()
+            .find(|(_, v)| v.len() >= self.inner.threshold())
+        else {
+            return Ok(());
+        };
+        debug_assert!(replica_messages.len() == self.inner.threshold());
```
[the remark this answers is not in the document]
pompon0: we want ==, to make sure that we don't wait for too many signatures and that we don't include them in the cert (which basically would be a waste of space).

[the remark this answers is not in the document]
pompon0: this is a debug_assert, because it would be technically correct to include more signatures.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -157,37 +141,29 @@ impl StateMachine {
...
-            .unwrap()
-            .values()
-            .cloned()
-            .collect::<Vec<_>>();
-
         // Create the justification for our message.
-        let justification = validator::CommitQC::from(&replica_messages, &consensus.validator_set)
-            .expect("Couldn't create justification from valid replica messages!");
+        let justification = validator::CommitQC::from(
+            &replica_messages.into_iter().cloned().collect::<Vec<_>>()[..],
```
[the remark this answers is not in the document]
pompon0: done

node/actors/bft/src/leader/state_machine.rs
pompon0: done

node/actors/bft/src/leader/state_machine.rs
pompon0: done

node/actors/bft/src/leader/tests.rs
pompon0: done

node/actors/bft/src/testonly/make.rs
pompon0: done

node/actors/bft/src/testonly/ut_harness.rs
pompon0: done

node/actors/bft/src/leader/state_machine.rs
pompon0: Thanks for catching that. It matters, although not that much - it doesn't affect correctness, and we don't have slashing so nothing bad happens. At worst it would just send out useless messages.

node/actors/bft/src/leader/replica_prepare.rs
pompon0: done

brunoffranca (commented):
I'm not satisfied with this design. It's a bit of a quick fix, I think it does decrease the readability of the code.
Also, looking at the future here, besides wanting to send LeaderPrepare messages asynchronously, we probably want to send asynchronously ReplicaCommit messages (because we need to verify the proposals and that might block) and ReplicaPrepare (when we implement the consensus oracle we will need to fetch external information, again might block). The only that doesn't need to be asynchronous is LeaderCommit.
I would prefer then that each of these messages are handled by a separate background task, and ideally without separating the message processing.

brunoffranca:
@moshababo Take a look. You're a code owner for this actor now too.

brunoffranca (approved):

PR #53, opened 2023-12-19
Simplified block storage

pompon0:
* simplified the storage traits, by moving the inmem caching implementation to era-consensus
* extracted PayloadManager trait responsible for proposing/verifying payloads.
* clarified the expected persistence guarantees of the PersistentBlockStore trait
* simplified the block caching to maintain a continuous range of blocks
* simplified the sync_blocks implementation as a result of previous simplifications
* moved rocksdb storage implementation to tools crate, given that the executor binary is there anyway.
* Removed genesis block from config to simplify configuration
* Removed redundant header from FinalBlock.
* Made bft actor wait for past blocks to be stored before calling propose/verify to more accurately reflect the expected implementation of a stateful blockchain (and to simplify implementation of the PayloadManager trait)


node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+use zksync_concurrency::{ctx, sync};
+use zksync_consensus_roles::validator;
+
+/// State of the `BlockStore`: continuous range of blocks.
+#[derive(Debug, Clone)]
+pub struct BlockStoreState {
+    /// Stored block with the lowest number.
+    pub first: validator::CommitQC,
+    /// Stored block with the highest number.
+    pub last: validator::CommitQC,
```
slowli, line 13: Bikeshedding: There's some inconsistency regarding "first / last" vs "lowest / highest" block numbers. Personally, I like "first / last" terms slightly better.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                let block = sync::wait_for(ctx, inner, |inner| !inner.cache.is_empty())
+                    .await?
+                    .cache
+                    .first_key_value()
+                    .unwrap()
+                    .1
+                    .clone();
+                self.0.persistent.store_next_block(ctx, &block).await?;
+                self.0.inner.send_modify(|inner| {
+                    debug_assert!(inner.persisted.next() == block.header().number);
```
slowli, line 94: Nit: You can use `debug_assert_eq` here for more expressive panic messages.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+    // TODO: this data structure can be optimized to a VecDeque (or even just a cyclical buffer).
+    cache: BTreeMap<validator::BlockNumber, validator::FinalBlock>,
+    cache_capacity: usize,
+}
+
+/// A wrapper around a PersistentBlockStore which adds caching blocks in-memory
+/// and other useful utilities.
+#[derive(Debug)]
+pub struct BlockStore {
+    inner: sync::watch::Sender<Inner>,
```
slowli, line 71: It may be just me, but I find this `watch::Sender` within `watch::Sender` construction quite difficult to reason about; in my mind, values are usually *sent* over channels, which isn't the case here (e.g., correctness of `BlockStore::subscribe()` depends on the value being largely retained). I don't have specific suggestions how to improve it though.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                .block(ctx, number)
+                .await?
+                .context("block disappeared from storage")?,
+        ))
+    }
+
+    /// Inserts a block to cache.
+    /// Since cache is a continuous range of blocks, `queue_block()`
+    /// synchronously waits until all intermediate blocks are added
+    /// to cache AND cache is not full.
```
slowli, line 167: IMO, these constraints make it difficult to reason about store usage correctness; e.g., it's moderately easy to deadlock it by queuing blocks out of order. A somewhat artificial example: in `SyncBlocks` actor, if the number of concurrently downloaded blocks is more than cache capacity and the corresponding permits were (incorrectly) dropped too late, after queuing the received block.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+        ctx: &ctx::Ctx,
+        block: &validator::FinalBlock,
+    ) -> ctx::Result<()>;
+}
+
+#[derive(Debug)]
+struct Inner {
+    inmem: sync::watch::Sender<BlockStoreState>,
+    persisted: BlockStoreState,
+    // TODO: this data structure can be optimized to a VecDeque (or even just a cyclical buffer).
```
slowli, line 62: Would it be difficult to use a `VecDeque` here? AFAIU, `cache` only ever contains continuous ranges of blocks, so using a `VecDeque` could help make this clear on the type level.

node/libs/storage/src/testonly/in_memory.rs
```diff
@@ -0,0 +1,71 @@
+//! In-memory storage implementation.
+use crate::{BlockStoreState, PersistentBlockStore, ReplicaState};
+use std::{collections::BTreeMap, sync::Mutex};
+use zksync_concurrency::ctx;
+use zksync_consensus_roles::validator;
+
+/// In-memory block store.
+#[derive(Debug, Default)]
+pub struct BlockStore(Mutex<BTreeMap<validator::BlockNumber, validator::FinalBlock>>);
```
slowli, line 9: - Ditto as for `crate::BlockStore`; using a `VecDeque` could help clarifying an invariant that only continuous ranges of blocks can be stored.
- Bikeshedding: Subjectively, having multiple structs named `BlockStore` in the same crate is slightly confusing.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+            Err(ctx::Error::Internal(err)) => Err(err),
+        }
+    }
+}
+
+impl BlockStore {
+    /// Constructs a BlockStore.
+    /// BlockStore takes ownership of the passed PersistentBlockStore,
+    /// i.e. caller should modify the underlying persistent storage
+    /// (add/remove blocks) ONLY through the constructed BlockStore.
```
slowli, line 112: Dumb question: Is removing blocks planned in the future? AFAIU, it's not implemented right now.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,218 @@
...
+        persistent: Box<dyn PersistentBlockStore>,
+        cache_capacity: usize,
+    ) -> ctx::Result<(Arc<Self>, BlockStoreRunner)> {
+        if cache_capacity < 1 {
+            return Err(anyhow::anyhow!("cache_capacity has to be >=1").into());
+        }
+        let state = persistent
+            .state(ctx)
+            .await?
+            .context("storage empty, expected at least 1 block")?;
```
slowli, line 124: Dumb question: Is this requirement going to be relaxed in the future? Looks weird that on the `PersistentBlockStore` level, empty stores are allowed, but they are not actually supported.

node/actors/bft/src/lib.rs
```diff
@@ -32,86 +30,94 @@ pub mod testonly;
...
+        ctx: &ctx::Ctx,
+        number: validator::BlockNumber,
+        payload: &validator::Payload,
+    ) -> ctx::Result<()>;
 }
 
-/// Protocol version of this BFT implementation.
-pub const PROTOCOL_VERSION: validator::ProtocolVersion = validator::ProtocolVersion::EARLIEST;
+/// Pipe through which bft actor sends network messages.
+pub(crate) type OutputPipe = ctx::channel::UnboundedSender<OutputMessage>;
```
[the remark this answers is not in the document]
pompon0: done

node/actors/executor/src/testonly.rs
```diff
@@ -1,111 +1,46 @@
...
-/// Configuration for a full non-validator node.
-#[derive(Debug)]
-#[non_exhaustive]
-pub struct FullNodeConfig {
-    /// Executor configuration.
-    pub node_config: ExecutorConfig,
-    /// Secret key of the node used for identification in the gossip network.
-    pub node_key: node::SecretKey,
+/// Creates a new full node and configures this validator to accept incoming connections from it.
+pub fn connect_full_node(rng: &mut impl Rng, node: &mut Config) -> Config {
```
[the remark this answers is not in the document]
pompon0: Yes, I don't want a testonly public method on a public struct, unless we additionally mark it, for example by adding a "testonly_" prefix to the method name.

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -3,376 +3,219 @@
...
+        state: BlockStoreState,
+    ) -> anyhow::Result<()> {
+        let last = state.last.header().number;
+        anyhow::ensure!(state.first.header().number <= state.last.header().number);
+        state
+            .last
+            .verify(&self.config.validator_set, self.config.consensus_threshold)
+            .context("state.last.verify()")?;
+        let mut peers = self.peers.lock().unwrap();
+        let permits = self.config.max_concurrent_blocks_per_peer;
```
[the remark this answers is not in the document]
pompon0: done

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -3,376 +3,219 @@
...
+        use std::collections::hash_map::Entry;
```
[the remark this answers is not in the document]
pompon0: this use statement is to abbreviate the names in the match arms' patterns. It is a very local "use" in that sense. Anyway, moved to the start of the block.

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -3,376 +3,219 @@
...
-    peers: Mutex<HashMap<node::PublicKey, PeerState>>,
-    pending_blocks: Mutex<PendingBlocks>,
-    message_sender: channel::UnboundedSender<io::OutputMessage>,
-    storage: Arc<dyn WriteBlockStore>,
     config: Config,
+    storage: Arc<BlockStore>,
+    message_sender: channel::UnboundedSender<io::OutputMessage>,
+
+    peers: Mutex<HashMap<node::PublicKey, PeerState>>,
+    highest: sync::watch::Sender<BlockNumber>,
```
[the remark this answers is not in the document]
pompon0: done

node/actors/sync_blocks/src/peers/tests/basics.rs
```diff
@@ -241,33 +246,45 @@ impl Test for DisconnectingPeer {
...
+            .update(&peer_key, test_validators.sync_state(1))
+            .unwrap();
 
         // Drop the response sender emulating peer disconnect.
-        message_receiver.recv(ctx).await?;
-
-        let peer_event = events_receiver.recv(ctx).await?;
-        assert_matches!(peer_event, PeerStateEvent::PeerDisconnected(key) if key == peer_key);
+        let msg = message_receiver.recv(ctx).await?;
+        assert_matches!(&msg, io::OutputMessage::Network(network::io::SyncBlocksInputMessage::GetBlock{
```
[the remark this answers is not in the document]
pompon0: done

node/actors/sync_blocks/src/peers/tests/multiple_peers.rs
```diff
@@ -79,10 +90,16 @@ impl Test for RequestingBlocksFromTwoPeers {
...
         );
+        tracing::info!(%first_peer_block_number,"received requrest");
 
         test_validators.send_block(first_peer_block_number, first_peer_response);
-        let peer_event = events_receiver.recv(ctx).await?;
-        assert_matches!(peer_event, PeerStateEvent::GotBlock(num) if num == first_peer_block_number);
+        wait_for_event(
+            ctx,
+            &mut events_receiver,
+            |ev| matches!(ev,PeerStateEvent::GotBlock(num) if num == first_peer_block_number),
```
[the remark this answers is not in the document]
pompon0: done

node/libs/storage/src/block_store.rs
pompon0: wdym by inconsistency? first block in storage is the one with the lowest block number by definition (last analogically)

node/libs/storage/src/block_store.rs
pompon0: done

node/libs/storage/src/block_store.rs
pompon0: ack, indeed use of watch::Sender here is non-trivial. "Sender" naming however IMO is the issue of the watch module, since it is the notification about update that is "sent", not the value stored in the watch.

node/libs/storage/src/block_store.rs
pompon0: Can you clarify your example? As I understand it:
* N = download permits in SyncBlocks actor
* M = max cache size
* N > M
* first N blocks are being downloaded in parallel
* (what happens here?)

I can remove the block cache size by replacing queue_block with store_block (i.e. make fetching tasks until the block is stored persistently), which actually simplify the api.

node/libs/storage/src/block_store.rs
pompon0: done

node/libs/storage/src/testonly/in_memory.rs
pompon0: * done
* ack, that's why we don't use fully qualified imports for these. We would need to prefix the struct names with the module name, which would cause unnecessary stuttering

node/libs/storage/src/block_store.rs
pompon0: we will most likely gc blocks at some point in the future. Then we will need the consensus to be aware that old blocks are not available any more. How it will be implemented is tbd. I've dropped the "add/remove" from the comment for now.

node/libs/storage/src/block_store.rs
pompon0: bft algorithm implementation assumes that the storage is non-empty, as fetching genesis block from a trusted source is out of scope of bft. I've allowed PersistentBlockStore to be empty to simplify implementations which operate on real storage and need to have a way of inserting a genesis block and a way of checking whether genesis is already there (and it looked weird to me, having special methods all over the place just for that). We can later reconsider whether restricting PersistentBlockStore more would be beneficial.

node/libs/storage/src/block_store.rs
pompon0: done

node/actors/bft/src/replica/state_machine.rs
```diff
@@ -7,14 +7,15 @@ use tracing::instrument;
...
-use zksync_consensus_storage::ReplicaStore;
 
 /// The StateMachine struct contains the state of the replica. This is the most complex state machine and is responsible
 /// for validating and voting on blocks. When participating in consensus we are always a replica.
 #[derive(Debug)]
 pub(crate) struct StateMachine {
     /// Consensus configuration and output channel.
-    pub(crate) inner: Arc<ConsensusInner>,
+    pub(crate) config: Arc<Config>,
+    /// Pipe through with replica sends network messages.
```
brunoffranca, line 17:
```suggestion
    /// Pipe through which replica sends network messages.
```

brunoffranca (approved):
LGTM.

brunoffranca (approved):

node/actors/sync_blocks/src/peers/mod.rs
```diff
@@ -3,376 +3,217 @@
...
+                }
+                Err(err) => {
+                    tracing::info!(%err, %number, ?peer, "get_block() failed");
+                    if let Some(send) = &self.events_sender {
+                        send.send(PeerStateEvent::RpcFailed {
+                            peer_key: peer.clone(),
+                            block_number: number,
+                        });
+                    }
+                    self.drop_peer(&peer);
```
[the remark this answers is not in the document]
pompon0: yes, it is a weak version of a ban.

node/libs/storage/src/block_store.rs
pompon0: Space of block numbers is fixed - there are block numbers without a block, so it is hard to say that last block number is the same as the number of the last block. And you could hardly call it a definition then.

node/actors/sync_blocks/src/peers/tests/multiple_peers.rs
```diff
@@ -11,25 +12,26 @@ impl Test for RequestingBlocksFromTwoPeers {
 
     fn tweak_config(&self, config: &mut Config) {
         config.sleep_interval_for_get_block = BLOCK_SLEEP_INTERVAL;
-        config.max_concurrent_blocks = 2;
+        config.max_concurrent_blocks = 5;
         config.max_concurrent_blocks_per_peer = 1;
         // ^ Necessary for blocks numbers in tests to be deterministic
```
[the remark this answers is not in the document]
pompon0: yes, it is intentional. This way we put a bound on both traffic to peers and the cache size. It makes sense, because on average there is no point for fetching data faster than it can be stored. We need a buffer only because both storage speed and fetching speed may vary over time.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,214 @@
...
+        &self,
+        ctx: &ctx::Ctx,
+        block: &validator::FinalBlock,
+    ) -> ctx::Result<()>;
+}
+
+#[derive(Debug)]
+struct Inner {
+    queued: sync::watch::Sender<BlockStoreState>,
+    persisted: BlockStoreState,
```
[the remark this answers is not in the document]
pompon0: done

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,214 @@
...
+/// and other useful utilities.
+#[derive(Debug)]
+pub struct BlockStore {
+    inner: sync::watch::Sender<Inner>,
+    persistent: Box<dyn PersistentBlockStore>,
+}
+
+/// Runner of the BlockStore background tasks.
+#[must_use]
+pub struct BlockStoreRunner(Arc<BlockStore>);
```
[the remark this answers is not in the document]
pompon0: As far as I'm concerned, allowing BlockStoreRunner to persist block even after all other references to BlockStore are dropped may be a valid use case. We do not use this property though.

node/libs/storage/src/block_store.rs
```diff
@@ -0,0 +1,214 @@
...
+        ctx: &ctx::Ctx,
+        block: &validator::FinalBlock,
+    ) -> ctx::Result<()>;
+}
+
+#[derive(Debug)]
+struct Inner {
+    queued: sync::watch::Sender<BlockStoreState>,
+    persisted: BlockStoreState,
+    queue: VecDeque<validator::FinalBlock>,
```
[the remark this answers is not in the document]
pompon0: As per https://github.com/matter-labs/era-consensus/pull/53#discussion_r1441540001 I've moved the responsibility of keeping cache bounded to the user, so that both cache size limit and inflight request limit are enforced in the same place.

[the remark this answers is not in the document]
pompon0: added metrics

node/libs/storage/src/block_store/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+        number: validator::BlockNumber,
+    ) -> ctx::Result<Option<validator::FinalBlock>> {
+        {
+            let inner = self.inner.borrow();
+            if !inner.queued_state.borrow().contains(number) {
+                return Ok(None);
+            }
+            if !inner.persisted_state.contains(number) {
+                let idx = number.0 - inner.persisted_state.next().0;
+                return Ok(Some(inner.queue[idx as usize].clone()));
```
slowli, line 158: Nit: I'd expect checked subtraction and `get(_)` instead of indexing `queue`; or a comment why these ops are safe (because `inner.queued_state.borrow().contains(number)` was checked earlier, right?).

node/libs/storage/src/testonly/in_memory.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    async fn store_next_block(
+        &self,
+        _ctx: &ctx::Ctx,
+        block: &validator::FinalBlock,
+    ) -> ctx::Result<()> {
+        let mut blocks = self.0.lock().unwrap();
+        let got = block.header().number;
+        if !blocks.is_empty() {
+            let want = blocks.back().unwrap().header().number.next();
```
slowli, line 55: Nit: This could be more idiomatically written as

```rust
if let Some(last_block) = blocks.back() {
    let want = last_block.header().number.next();
    // ...
}
```

slowli (approved):

node/tools/src/store.rs
```diff
@@ -0,0 +1,200 @@
+//! This module contains the methods to handle an append-only database of finalized blocks. Since we only store finalized blocks, this forms a
+//! chain of blocks, not a tree (assuming we have all blocks and not have any gap). It allows for basic functionality like inserting a block,
+//! getting a block, checking if a block is contained in the DB. We also store the head of the chain. Storing it explicitly allows us to fetch
+//! the current head quickly.
```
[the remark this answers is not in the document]
pompon0: done

node/libs/storage/src/traits.rs
```diff
@@ -1,79 +1,37 @@
...
-    /// of the genesis block.
-    fn subscribe_to_block_writes(&self) -> watch::Receiver<BlockNumber>;
+    /// Verify that `payload` is a correct proposal for the block `block_number`.
+    async fn verify_payload(&self, ctx: &ctx::Ctx, block_number: BlockNumber, payload: &Payload) -> ctx::Result<()>;
 }
 
-/// Mutable storage of L2 blocks.
-///
-/// Implementations **must** propagate context cancellation using [`ctx::Error::Canceled`].
+/*
```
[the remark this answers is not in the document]
pompon0: this file is not used any more, removed.

node/actors/sync_blocks/src/tests/end_to_end.rs
```diff
@@ -70,67 +53,81 @@ impl fmt::Debug for Node {
...
+
 impl Node {
-    fn new(mut network: NetworkInstance) -> (Self, InitialNodeHandle) {
-        let (create_block_sender, create_block_receiver) = channel::unbounded();
-        let (sync_state_subscriber_sender, sync_state_subscriber_receiver) = oneshot::channel();
+    async fn new_network(
+        ctx: &ctx::Ctx,
+        node_count: usize,
+        gossip_peers: usize,
+    ) -> (Vec<NodeHandle>, Vec<Node>) {
```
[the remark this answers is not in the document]
pompon0: done

node/libs/storage/src/testonly/in_memory.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    async fn block(
+        &self,
+        _ctx: &ctx::Ctx,
+        number: validator::BlockNumber,
+    ) -> ctx::Result<validator::FinalBlock> {
+        let blocks = self.0.lock().unwrap();
+        let front = blocks.front().context("not found")?;
+        let idx = number.0 - front.header().number.0;
```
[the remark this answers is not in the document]
pompon0: thx

node/libs/storage/src/block_store/mod.rs
pompon0: done

node/libs/storage/src/testonly/in_memory.rs
pompon0: done

PR #55, opened 2024-01-08
Improvements on standalone node execution

IAvecilla:
## What ❔

Documented a guide on setting up and running a local consensus node, along with a new Makefile containing commands for easy execution. Also modifies the database directory path for each node.

## Why ❔

The README and the Makefile would streamline generating the configuration for all nodes with a single command and simplify running the desired node. About the modification for the database directory path, it seems beneficial for running multiple nodes on a single machine, especially for local testing. However, in the broader architecture of the Test Framework, each node is expected to operate within a distinct Docker container or a similar setup. In that context, this change might not be useful.


node/tools/src/main.rs
```diff
@@ -100,11 +100,13 @@ async fn main() -> anyhow::Result<()> {
...
-    let storage = RocksdbStorage::new(
-        ctx,
-        &configs.executor.genesis_block,
-        Path::new("./database"),
-    );
+    // Creates a database directory for each node, this allows to run multiple nodes on the same machine.
+    let database_path = if let Some((consensus_config, _)) = &configs.consensus {
+        Path::new("database").join(consensus_config.public_addr.to_string())
+    } else {
+        Path::new("database").join(configs.executor.server_addr.to_string())
```
pompon0, line 107: fyi, "database" is a flag now

node/tools/Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: run-node node-configs clean
+IP?=127.0.0.1:8000
```
pompon0, line 2: is this a good default? This port has some canonical uses https://www.speedguide.net/port.php?port=8000 , so perhaps something more random would be preferred.

node/tools/Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+run-node:
+	export RUST_LOG=INFO && cargo run -- --config-file node-configs/${IP}/config.json --node-key node-configs/${IP}/node_key --validator-key node-configs/${IP}/validator_key --database ./database/${IP} 
+
```
pompon0, line 6: nit: note that if you execute binary from "node-configs/${IP}" working directory, you don't have to set any flags.

node/tools/README.md
```diff
@@ -0,0 +1,5 @@
+# Running a test node
+
+1. Generate a file named `config.txt` in the root directory of the tools crate, containing node addresses in the format `IP:PORT`, with each address on a separate line.
```
pompon0, line 3: nit: this is a metaconfig in a sense (used to generate configs), so perhaps giving it a different name than "config" would be helpful.

pompon0 (approved):

node/tools/src/main.rs
pompon0: that was intentional, thx for updating the comment

node/tools/Makefile
IAvecilla: Updated here [9fe5437](https://github.com/matter-labs/era-consensus/pull/55/commits/9fe5437f3bda51f8010a2264f4aa912ca4c1229c)!

node/tools/README.md
IAvecilla: Sounds reasonable, I'm using `addresses` now to be more specific

PR #56, opened 2024-01-10
(WIP) Make consensus state machine asynchronous (BFT-406)

moshababo:
Goals:
* `leader` being unable to block colocated `replica` (and vise versa)
* Having an inbound message queue with the capability to invalidate/remove specific pending items

Notes:
* `Deduper` struct has been added to wrap a FIFO queue and provide functionality for removal of specific pending items
* The initial solution, which is currently implemented only for `leader`, involves a single loop that consumes all corresponding message from the FIFO queue. At the top-level, processing involves only the `enqueue()` operation and does not await the processing response.

TODOs:
* `Deduper` to support `dequeue()` to await/block until next item is available, through some reliable signaling mechanism. Both `enqueue()` and `dequeue()` should continue to require immutable reference only, for bridging between producer and consumer
* Current work doesn’t pass the in-process p2p tests; need to further debug
* Implement a similar mechanism for `replica`
* Re-evaluate caching and transitioning within the state machines 
* Re-evaluate message processing lifecycle and results reporting considering the new asynchronous flow

Follow-up:
* Consider additional asynchronous pipelines within each role

### DIFF
--- a/node/actors/bft/src/leader/state_machine.rs
+++ b/node/actors/bft/src/leader/state_machine.rs
@@ -4,10 +4,15 @@ use std::{
     sync::Arc,
     unreachable,
 };
+use std::time::Duration;
+
 use tracing::instrument;
+
 use zksync_concurrency::{ctx, error::Wrap as _, metrics::LatencyHistogramExt as _, sync, time};
 use zksync_consensus_network::io::{ConsensusInputMessage, Target};
 use zksync_consensus_roles::validator;
+
+use crate::deduper::Deduper;
 
 /// The StateMachine struct contains the state of the leader. This is a simple state machine. We just store
 /// replica messages and produce leader messages (including proposing blocks) when we reach the threshold for
@@ -59,6 +64,48 @@ impl StateMachine {
             prepare_qc: sync::watch::channel(None).0,
             commit_qcs: BTreeMap::new(),
         }
+    }
+
+    pub async fn run_process_queue(&mut self, ctx: &ctx::Ctx, queue: &Deduper<validator::Signed<validator::ConsensusMsg>>) -> ctx::Result<()> {
+        loop {
+            let signed_message = queue.dequeue().await;
+            let Some(signed_message) = signed_message else {
+                // println!("None");
+                tokio::time::sleep(Duration::from_millis(100)).await; // temp solution for rate limiting until a blocking solution will be implemented for `queue.dequeue()`
+                continue;
+            };
+
+            println!("{:?}", signed_message);
+
+            match signed_message.msg {
+                validator::ConsensusMsg::ReplicaPrepare(_) => {
+                    let _ = self.process_replica_prepare(ctx, signed_message.cast().unwrap()).await;
+                }
+                validator::ConsensusMsg::ReplicaCommit(_) => {
+                    let _ = self.process_replica_commit(ctx, signed_message.cast().unwrap());
+                }
+                _ => unreachable!(),
+            }
+        }
+    }
+
+    // temp: minimal debuggable version of `process_input`
+    pub(crate) async fn process_input_debug(
+        &mut self,
+        ctx: &ctx::Ctx,
+        input: validator::Signed<validator::ConsensusMsg>,
+    ) -> ctx::Result<()> {
+        println!("{:?}", input);
+        match &input.msg {
+            validator::ConsensusMsg::ReplicaPrepare(_) => {
+                let _ = self.process_replica_prepare(ctx, input.cast().unwrap()).await;
+            }
+            validator::ConsensusMsg::ReplicaCommit(_) => {
+                let _ = self.process_replica_commit(ctx, input.cast().unwrap());
+            }
+            _ => unreachable!(),
+        };
+        Ok(())
     }
 
     /// Process an input message (leaders don't time out waiting for a message). This is the
@@ -201,7 +248,7 @@ impl StateMachine {
                 message: msg,
                 recipient: Target::Broadcast,
             }
-            .into(),
+                .into(),
         );
         Ok(())
     }
--- a/node/actors/bft/src/lib.rs
+++ b/node/actors/bft/src/lib.rs
@@ -14,11 +14,18 @@
 //! - [Notes on modern consensus algorithms](https://timroughgarden.github.io/fob21/andy.pdf)
 //! - [Blog post comparing several consensus algorithms](https://decentralizedthoughts.github.io/2023-04-01-hotstuff-2/)
 //! - Blog posts explaining [safety](https://seafooler.com/2022/01/24/understanding-safety-hotstuff/) and [responsiveness](https://seafooler.com/2022/04/02/understanding-responsiveness-hotstuff/)
-use crate::io::{InputMessage, OutputMessage};
-use std::sync::Arc;
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+
+pub use config::Config;
 use zksync_concurrency::{ctx, scope};
 use zksync_consensus_roles::validator;
+use zksync_consensus_roles::validator::{ConsensusMsg, Signed};
+use zksync_consensus_storage::ReplicaStore;
 use zksync_consensus_utils::pipe::ActorPipe;
+
+use crate::deduper::Deduper;
+use crate::io::{InputMessage, OutputMessage};
 
 mod config;
 pub mod io;
@@ -29,8 +36,7 @@ mod replica;
 pub mod testonly;
 #[cfg(test)]
 mod tests;
-
-pub use config::Config;
+mod deduper;
 
 /// Protocol version of this BFT implementation.
 pub const PROTOCOL_VERSION: validator::ProtocolVersion = validator::ProtocolVersion::EARLIEST;
@@ -65,15 +71,23 @@ impl Config {
         mut pipe: ActorPipe<InputMessage, OutputMessage>,
     ) -> anyhow::Result<()> {
         let cfg = Arc::new(self);
+        let mut leader = leader::StateMachine::new(ctx, cfg.clone(), pipe.send.clone());
+        let leader_queue = Deduper::new(
+            Mutex::new(VecDeque::<Signed<ConsensusMsg>>::new()),
+            Box::new(|_, _| { true }), // TODO: apply actual dedup predicate
+        );
+
         let res = scope::run!(ctx, |ctx, s| async {
-            let mut replica =
-                replica::StateMachine::start(ctx, cfg.clone(), pipe.send.clone()).await?;
-            let mut leader = leader::StateMachine::new(ctx, cfg.clone(), pipe.send.clone());
+            let mut replica = replica::StateMachine::start(ctx, cfg.clone(), pipe.send.clone()).await?;
+
+            let prepare_qc_receiver = leader.prepare_qc.subscribe();
+
+            s.spawn_bg(leader.run_process_queue(ctx, &leader_queue));
 
             s.spawn_bg(leader::StateMachine::run_proposer(
                 ctx,
                 &cfg,
-                leader.prepare_qc.subscribe(),
+                prepare_qc_receiver,
                 &pipe.send,
             ));
 
@@ -102,19 +116,22 @@ impl Config {
                 use validator::ConsensusMsg as Msg;
                 let res = match &req.msg.msg {
                     Msg::ReplicaPrepare(_) | Msg::ReplicaCommit(_) => {
-                        leader.process_input(ctx, req.msg).await
+                        leader_queue.enqueue(req.msg);
+                        //leader.process_input_debug(ctx, req.msg).await;
+                        ()
                     }
                     Msg::LeaderPrepare(_) | Msg::LeaderCommit(_) => {
-                        replica.process_input(ctx, req.msg).await
+                        let _ = replica.process_input(ctx, req.msg).await;
+                        ()
                     }
                 };
                 // Notify network actor that the message has been processed.
                 // Ignore sending error.
                 let _ = req.ack.send(());
-                res?;
+               // res?;
             }
         })
-        .await;
+            .await;
         match res {
             Ok(()) | Err(ctx::Error::Canceled(_)) => Ok(()),
             Err(ctx::Error::Internal(err)) => Err(err),
--- a/node/actors/bft/src/deduper.rs
+++ b/node/actors/bft/src/deduper.rs
@@ -1,0 +1,85 @@
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+pub(crate) struct Deduper<T> {
+    queue: Mutex<VecDeque<T>>, // TODO: don't bound to a concrete type
+    predicate: Box<dyn Sync + Send + Fn(&T, &T) -> bool>,
+}
+
+impl<T> Deduper<T> {
+    pub fn new(queue: Mutex<VecDeque<T>>, predicate: Box<dyn Sync + Send + Fn(&T, &T) -> bool>) -> Self {
+        Self {
+            queue,
+            predicate,
+        }
+    }
+
+    pub fn enqueue(&self, item: T) {
+        let mut queue = self.queue.lock().unwrap();
+        queue.retain(|existing_item| (self.predicate)(existing_item, &item));
+        queue.push_back(item)
+    }
+
+    pub async fn dequeue(&self) -> Option<T> {
+        // TODO: await/block until next item is available
+        let mut queue = self.queue.lock().unwrap();
+        queue.pop_front()
+    }
+}
+
+// TODO: this is not a proper test yet, just a temp debugging tool
+#[tokio::test]
+async fn test_deduper() {
+    #[derive(PartialEq, Debug)]
+    struct MockType {
+        id: usize,
+    }
+    let mut deduper = Arc::new(Deduper::new(
+        Mutex::new(VecDeque::<MockType>::new()),
+        Box::new(|a: &MockType, b: &MockType| {
+            a.id != b.id
+        }),
+    ));
+
+    let mut deduper_clone = deduper.clone();
+    let handle1 = tokio::spawn(async move {
+        for i in 1..=90 {
+            deduper_clone.enqueue(MockType { id: i });
+            tokio::task::yield_now().await;
+            // println!("{:?}", i)
+        }
+        tokio::time::sleep(Duration::from_millis(1)).await;
+
+        for i in 100..=190 {
+            deduper_clone.enqueue(MockType { id: i })
+        }
+        tokio::time::sleep(Duration::from_millis(1)).await;
+
+        for i in 200..=290 {
+            deduper_clone.enqueue(MockType { id: i })
+        }
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    });
+
+    let mut deduper_clone = deduper.clone();
+    let handle2 = tokio::spawn(async move {
+        let mut counter = 1;
+        loop {
+            let item = deduper_clone.dequeue().await;
+            match item {
+                Some(item) => {
+                    println!("{:?} {:?}", counter, item);
+                }
+                None => {
+                    println!("{:?} {:?}", counter, item);
+                }
+            }
+
+            counter += 1;
+            tokio::task::yield_now().await
+        }
+    });
+
+    let _ = tokio::join!(handle1);
+}


node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+//! The pruning takes place whenever a new item is sent, based on a specified predicate.
+//!
+//! The queue facilitates the asynchronous return of processing result for each added item.
+//!
+//! The separation of `Sender` and `Receiver` is employed primarily because `Receiver` requires
+//! a mutable reference to the signaling channel, unlike `Sender`, hence making it undesirable to
+//! be used in conjunction.
+
+use crate::{
+    ctx::{self, Ctx},
```
pompon0, line 12: please don't import Ctx directly, but rather use ctx::Ctx.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    oneshot,
+    sync::{self, watch},
```
pompon0, line 14: same as with ctx: use sync::watch rather than importing watch directly.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+};
+use std::{collections::VecDeque, sync::Arc};
+use tokio::sync::Mutex;
```
pompon0, line 17: tokio::sync::Mutex is available as crate::sync::Mutex.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+
+pub fn new<T, U>(
```
pompon0, line 19: nit: the convention is to name it as "channel"

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    pruning_predicate: Box<dyn Sync + Send + Fn(&T, &T) -> bool>,
```
pompon0, line 20: document what it is expected to do.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+impl<T, U> Sender<T, U> {
+    pub async fn enqueue(&self, item: T) -> oneshot::Receiver<U> {
+        // Create oneshot channel for returning result asynchronously.
+        let (res_send, res_recv) = oneshot::channel();
+
+        let mut queue = self.shared.queue.lock().await;
+        queue.retain(|existing_item| !(self.pruning_predicate)(&existing_item.0, &item));
```
pompon0, line 60: as far as I'm concerned, with our use case we are good with running the predicate over all messages in the queue when a new message arrives, but please remember to document that pushing a new message takes O(n).

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    has_items_send: watch::Sender<bool>,
+}
+
+pub struct Sender<T, U> {
+    shared: Arc<Shared<T, U>>,
+    pruning_predicate: Box<dyn Sync + Send + Fn(&T, &T) -> bool>,
+}
+
+impl<T, U> Sender<T, U> {
+    pub async fn enqueue(&self, item: T) -> oneshot::Receiver<U> {
```
pompon0, line 55: nit: by convention it should be called "send".

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        // Create oneshot channel for returning result asynchronously.
+        let (res_send, res_recv) = oneshot::channel();
+
+        let mut queue = self.shared.queue.lock().await;
+        queue.retain(|existing_item| !(self.pruning_predicate)(&existing_item.0, &item));
+        queue.push_back((item, res_send));
```
pompon0, line 61: what if the new message is older than a message already in the queue?

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+pub struct Receiver<T, U> {
+    shared: Arc<Shared<T, U>>,
+    has_items_recv: watch::Receiver<bool>,
+}
+
+impl<T, U> Receiver<T, U> {
+    pub async fn dequeue(&mut self, ctx: &Ctx) -> ctx::OrCanceled<(T, oneshot::Sender<U>)> {
```
pompon0, line 76: nit: recv

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        sync::wait_for(ctx, &mut self.has_items_recv, |has_items| *has_items).await?;
```
pompon0, line 77: nit: you can obtain the same effect if you wrapped the VecDeque into a watch, no?
Sth like `Arc<watch::Sender<VecDeque<T>>>`.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut queue = self.shared.queue.lock().await;
+        // `None` is unexpected because we waited for new items.
```
pompon0, line 79: ... and there is only 1 receiver.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        let item = queue.pop_front().unwrap();
+
+        if queue.len() == 0 {
+            // Send error is unexpected because `self` holds the receiver.
+            self.shared.has_items_send.send(false).unwrap();
```
pompon0, line 84: use send_replace which is infallible.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        Ok(item)
+    }
+}
+
+// Test scenario:
+// Enqueue two sets of 0..1000 items, in conjunction, while pruning
+// so that only one 0..1000 set is expected to remain in queue.
+// Then, dequeue to assert the queue's content.
+#[tokio::test]
```
pompon0, line 95: nit: please move tests to a separate module in a separate file.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    });
+    let handle2 = tokio::spawn(async move {
+        let set = 2;
+        let items = (0..1000).map(|i| ItemType(i, set));
+        for item in items {
+            let _ = sender2.enqueue(item).await;
+            tokio::task::yield_now().await;
+        }
+    });
+    tokio::try_join!(handle1, handle2).unwrap();
```
pompon0, line 125: please use structured concurrency for writing tests, unless infeasible.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
+//! Prunable queue mimics a producer-consumer channel, in enabling
```
pompon0, line 1: nit: if you want to present it as mpsc with pruning, maybe "prunable_mpsc" would be more accurate name? Just a suggestion.

pompon0: also document that it is unbounded.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    let sender1 = Arc::new(sender);
+    let sender2 = sender1.clone();
+
+    let handle1 = tokio::spawn(async move {
+        let set = 1;
+        let items = (0..1000).map(|i| ItemType(i, set));
+        for item in items {
+            let _ = sender1.enqueue(item).await;
+            tokio::task::yield_now().await;
```
pompon0, line 114: I'd suggest to manually implement a deterministic random interleaving, since it is feasible here (rather than rely on yield_now).

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: also note that using enqueue/dequeue like this requires guaranteeing cancel safety. Cancel safety when accepting ctx as an argument makes little sense (cancel safety: dropping a future is ok; accepting a ctx: to cancel, cancel the context and wait for the future to complete by itself)

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    tokio::spawn(async move {
+        let mut i = 0;
+        loop {
+            let (item, sender) = receiver.dequeue(&ctx).await.unwrap();
+            assert_eq!(item.0, i);
+            let _ = sender.send(());
+
+            i = i + 1;
+            if i == 1000 {
+                match timeout(Duration::from_secs(0), receiver.dequeue(&ctx)).await {
```
pompon0, line 136: afair tokio::time::timeout doesn't specify the semantics when timeout is 0 AND future completes immediately. IMO it would be more beneficial to run receiver and sender (adding additional items) in parallel (on a prepopulated queue, which you already prepared) to make sure that consumer doesn't overtake the producer. Alternatively you can add a "try_recv" method for checking that the queue is actually empty.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+                    Ok(_) => assert!(
```
pompon0, line 137: nit: assert!(timeout(...).await.is_err(), "... is expected to hang ...")

pompon0 (commented):
approach lg in general, I've added comments about conventions and expected usage of structured concurrency. The important comment is the one about pruning semantics which won't work as is, but should be easy to fix.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    let receiver = Receiver {
+        shared: shared.clone(),
+        has_items_recv: recv,
+    };
+
+    return (sender, receiver);
+}
+
+pub struct Shared<T, U> {
+    queue: Mutex<VecDeque<(T, oneshot::Sender<U>)>>,
```
pompon0, line 45: synchronous mutex should be good enough here (or watch::Sender, as mentioned in another comment). It will also make the enqueue sync.

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: nit: consider accepting impl Send + Sync + Fn(...) and wrapping inside the constructor.

node/actors/bft/src/replica/state_machine.rs
```diff
@@ -59,58 +58,70 @@ impl StateMachine {
             high_vote: backup.high_vote,
             high_qc: backup.high_qc,
             block_proposal_cache,
-            timeout_deadline: time::Deadline::Infinite,
+            timeout_deadline: sync::watch::channel(time::Deadline::Infinite).0,
         };
         // We need to start the replica before processing inputs.
         this.start_new_view(ctx).await.wrap("start_new_view()")?;
         Ok(this)
```
brunoffranca, line 65: Nit: Maybe we move this part to the run method. Rename this method to `new`.

node/actors/bft/src/lib.rs
```diff
@@ -65,15 +66,28 @@ impl Config {
...
     ) -> anyhow::Result<()> {
         let cfg = Arc::new(self);
+        let leader = leader::StateMachine::new(ctx, cfg.clone(), pipe.send.clone());
+        let replica = replica::StateMachine::start(ctx, cfg.clone(), pipe.send.clone()).await?;
+
+        let (leader_send, leader_recv) =
+            sync::prunable_queue::new(Box::new(leader::StateMachine::queue_pruning_predicate));
+
+        let (replica_send, replica_recv) =
+            sync::prunable_queue::new(Box::new(replica::StateMachine::queue_pruning_predicate));
```
brunoffranca, line 76: Nit: Couldn't this be inside the `new` methods?

brunoffranca (approved):
Looks good other than a few nits. I'll approve it.

node/actors/bft/src/lib.rs
```diff
@@ -82,36 +96,46 @@ impl Config {
...
                     return Ok(());
                 }
 
+                // Check if an asynchronous error was received.
+                if let Ok(res) = err_recv.try_recv() {
+                    return Err(res.err().unwrap());
+                }
+
                 let Some(InputMessage::Network(req)) = input else {
-                    replica.start_new_view(ctx).await?;
```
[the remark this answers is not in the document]
brunoffranca: +100

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/actors/bft/src/lib.rs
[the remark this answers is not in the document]
moshababo: I moved it to the management of the replica's event loop [here](https://github.com/matter-labs/era-consensus/blob/8099e4d912327b8b5e1dbcb654472b697c0e61e7/node/actors/bft/src/replica/state_machine.rs#L82).

(to clarify—it was moved there because `replica.run()` now consumes `self`, or at least should permanently own the mutable reference, so it's not possible to call `replica.start_new_view()` from here anymore. I used the `Option::None` mechanism which was already in place to communicate with the event loop through the queue. This is indeed awkward.)

Do you mean to move the timeout management under `replica`? It would not be logically identical to the previous behavior, where messages to `leader` would invalidate the timeout. Was this a shortcoming of the previous implementation?

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: It might be possible, but I don't think it's a good thing to have a `T` for `watch` that needs to be mutated.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Doesn't sync `Mutex` halt async execution environment?

moshababo: Doesn't sync `Mutex` halts async execution environments?

node/actors/bft/src/replica/state_machine.rs
moshababo: We can still separate `new()` from `start()` (in addition to `run()`), but it's probably too verbose. 

If it's too confusing, I can rename `run()` with something more specific. I chose not to in order to emphasize that it serves as the main loop.

node/actors/bft/src/lib.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/0ac07073f1481289df1d7f6dbbaf31c15f10cd45

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: I didn't call it a channel because it doesn't cover all the functionality of one, but it might be less confusing option at this point. 

I renamed as `prunable_mpsc`.

moshababo: I didn't call it a channel because it doesn't cover all the functionality of one, but it might be the less confusing option at this point. 

I renamed as `prunable_mpsc`.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Can you clarify what you mean in the context of the test code? are you referring to being cancel-safe in the loops?

Regarding `recv()` (and unrelated to this test code)—it accepts `ctx` as an argument for cancellation watch. Watching on the loop level isn't sufficient, because `recv()` can hang indefinitely.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: I left this code because I planned to further implement something non-deterministic, but it's probably a bad idea. 
I'll do something more reasonably-deterministic as you suggested, without relying on `yield_now`.

node/libs/concurrency/src/sync/prunable_queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        let set = 2;
+        let items = (0..1000).map(|i| ItemType(i, set));
+        for item in items {
+            let _ = sender2.enqueue(item).await;
+            tokio::task::yield_now().await;
+        }
+    });
+    tokio::try_join!(handle1, handle2).unwrap();
+
+    tokio::spawn(async move {
```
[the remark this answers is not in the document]
moshababo: It's indeed useless for runtime, I put it there as an example and for compile time guarantees that the receiver is usable in that context (basically ensuring that sender and receiver are decoupled).
I can remove if you think it's redundant as well.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: > afair tokio::time::timeout doesn't specify the semantics when timeout is 0 AND future completes immediately.

> Alternatively you can add a "try_recv" method for checking that the queue is actually empty.

I came this solution due to the no-guarantees in `tokio::select!` which were noticeable, while this solution seems stable. Looking at the code, it indeed tries first to poll the future, and only if not ready proceeds to the deadline. So it might be better than adding a "peek"-style method which isn't needed otherwise. 

> IMO it would be more beneficial to run receiver and sender (adding additional items) in parallel (on a prepopulated queue, which you already prepared) to make sure that consumer doesn't overtake the producer.

I agree that the test case should be improved. It was just for having something basic until getting a review ack.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Currently, won't be pruned. I wasn't sure about this.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Test code changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/020f25b2049f9f4bb2dbbddc9a0e876eadaf5389

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Test code changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/020f25b2049f9f4bb2dbbddc9a0e876eadaf5389

It doesn't involve anything sophisticated, just covers the basic features.

node/libs/concurrency/src/sync/prunable_queue.rs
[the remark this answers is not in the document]
moshababo: Outdated in the new test version, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/020f25b2049f9f4bb2dbbddc9a0e876eadaf5389

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Test code changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/020f25b2049f9f4bb2dbbddc9a0e876eadaf5389

node/actors/bft/src/lib.rs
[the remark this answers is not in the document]
moshababo: I went ahead and made the fix, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/3620997e2cdf9fd7c4e926230d261992c68c0a01

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: it should be pruned, as it opens an attack vector of getting spammed with messages in decreasing view order.

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: why not though?

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: Sync mutex blocks the thread, so it is one thread less for other tasks to use. However same applies to just running a task on a thread which takes nontrivial amount of time. Fast operations are ok to be executed under a blocking mutex on a runtime. No deadlock is possible, because you cannot hold a lock when a task yields (MutexGuard is !Send).

node/actors/bft/src/leader/state_machine.rs
```diff
@@ -46,10 +48,19 @@ pub(crate) struct StateMachine {
 impl StateMachine {
     /// Creates a new StateMachine struct.
```
pompon0, line 49: document how to use the result, which is nontrivial now.

node/actors/bft/src/lib.rs
```diff
@@ -96,34 +90,27 @@ impl Config {
...
+                let InputMessage::Network(req) = input.unwrap();
                 let res_recv;
                 match &req.msg.msg {
                     ConsensusMsg::ReplicaPrepare(_) | ConsensusMsg::ReplicaCommit(_) => {
-                        res_recv = leader_send.enqueue(req.msg).await;
+                        res_recv = leader_send.send(req.msg).await;
                     }
                     ConsensusMsg::LeaderPrepare(_) | ConsensusMsg::LeaderCommit(_) => {
-                        res_recv = replica_send.enqueue(Some(req.msg)).await;
+                        res_recv = replica_send.send(req.msg).await;
```
pompon0, line 113: can you pass the response channel together with the message to the replica/leader, so that you don't have to have handle_result at all?

node/actors/bft/src/replica/state_machine.rs
```diff
@@ -36,8 +41,11 @@ impl StateMachine {
     pub(crate) async fn start(
         ctx: &ctx::Ctx,
         config: Arc<Config>,
-        pipe: OutputSender,
-    ) -> ctx::Result<Self> {
+        outbound_pipe: OutputSender,
+    ) -> ctx::Result<(
+        Self,
+        sync::prunable_mpsc::Sender<Signed<ConsensusMsg>, ctx::Result<()>>,
```
pompon0, line 47: ditto: please document the result.

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+#[cfg(test)]
+mod tests;
+
+/// Creates a channel and returns the [`Sender`] and [`Receiver`] handles.
+/// All values sent on [`Sender`] will become available on [`Receiver`] in the same order as it was sent,
+/// unless will be pruned before received.
+/// The Sender can be cloned to send to the same channel from multiple code locations. Only one Receiver is supported.
+///
+/// * [`T`]: The type of data that will be sent through the channel.
+/// * [`U`]: The type of the asynchronous returning of result associated with the processing of values received from the channel.
```
pompon0, line 24: T,U naming is not very informative at this point.
do you mean that T is a Request and U is a Response? If so I'd suggest Req/Resp naming.
And given that it supports bidirectional communication now (which I think is a nice idea) it is no longer a mpsc but rather just an async service, and it would benefit from changing the api altogether (and the name, of course):
* easy: instead of `send(Req) -> Receiver<Resp>`, just `call(Req) -> OrPruned<Result<Resp>>` would be better.
* not-so-easy: instead of `recv() -> (Req,Sender<Resp>)`, you can make channel() accept a Server as an argument, where Server just implements a function `handle(&mut self, Req) -> Result<Resp,Err>`

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
...
+#[tokio::test]
+async fn test_prunable_mpsc() {
+    let ctx = ctx::test_root(&ctx::RealClock);
+
+    #[derive(Debug, Clone)]
+    struct ValueType(usize, usize);
+
+    #[allow(clippy::type_complexity)]
+    let (send, mut recv): (
+        super::Sender<ValueType, Result<(), usize>>,
```
pompon0, line 17: nit, test modules are allowed to "use super::*;"

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    let res: Result<(), ctx::Canceled> = crate::scope::run!(&ctx, |ctx, s| async move {
+        // Pre-send sets 0 and 1, 1000 values each.
+        // Set 0 is expected to be pruned and dropped.
+        let values = (0..2000).map(|i| ValueType(i / 1000, i % 1000));
+        for val in values {
+            let res_recv = send.send(val.clone()).await;
+            s.spawn(async move {
+                let res = res_recv.recv_or_disconnected(ctx).await;
```
pompon0, line 31: .await.unwrap(), since outermost Result is always expected to be Ok.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
+use crate::ctx;
+use assert_matches::assert_matches;
+use tokio::time::{timeout, Duration};
+
+// Test scenario:
+// 1. Pre-send two sets of 1000 values, so that the first set is expected to be pruned.
+// 2. Send a third set of 1000 values in parallel to receiving.
+#[tokio::test]
+async fn test_prunable_mpsc() {
```
pompon0, line 9: abort_on_panic()

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
...
+        a.1 == b.1
+    });
+
+    let res: Result<(), ctx::Canceled> = crate::scope::run!(&ctx, |ctx, s| async move {
+        // Pre-send sets 0 and 1, 1000 values each.
+        // Set 0 is expected to be pruned and dropped.
+        let values = (0..2000).map(|i| ValueType(i / 1000, i % 1000));
+        for val in values {
+            let res_recv = send.send(val.clone()).await;
+            s.spawn(async move {
```
pompon0, line 30: nit: "async move" is rarely applicable in scope::run/spawn/spawn_bg. The convention is to always use just "async { .. }" which is way more flexible than "async move". If you need to move x explicitly inside the closure, use "let x = x;". If x implements Copy, first wrap it in NoCopy then move it as previously.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
...
+        s.spawn(async move {
+            let values = (1000..2000).map(|i| ValueType(2, i));
+            for val in values {
+                let res_recv = send.send(val.clone()).await;
+                s.spawn(async move {
+                    let res = res_recv.recv_or_disconnected(ctx).await;
+                    let i = val.1;
+                    match val.0 {
+                        // set 2 values are expected to return `Err(i)`.
+                        2 => assert_matches!(res, Ok(Ok(Err(err))) => {
```
pompon0, line 52: nit: just `assert_eq(2,val.0)` directly.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
```diff
@@ -0,0 +1,94 @@
...
+                    // Return `Ok(())` for set 1.
+                    1 => res_send.send(Ok(())).unwrap(),
+                    // Return `Err(i)` for set 2.
+                    2 => res_send.send(Err(i)).unwrap(),
+                    _ => unreachable!(),
+                };
+                i += 1;
+                if i == 2000 {
+                    assert!(
+                        timeout(Duration::from_secs(0), recv.recv(ctx))
```
pompon0, line 80: nit: as previously, `timeout()` doesn't guarantee the expected semantics here.

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: Cancel safety is defined here: https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety.
tokio::timeout() requires the passed future to be cancel safe, because it drops it in case of timeout (https://docs.rs/tokio/latest/tokio/time/fn.timeout.html#cancellation). Additionally for this assert to work as intended, the `Receiver::recv(ctx)` here would need to guarantee that it completes immediately in case the queue is non-empty (which is not documented as of now). `Receiver::recv(ctx)` calls  `sync::wait_for` which doesn't provide such guarantee, so it would be hard to provide this guarantee. In general, none of the methods in concurrency crate do, as it is a very fragile property.

node/actors/bft/src/lib.rs
```diff
@@ -96,34 +90,27 @@ impl Config {
...
-                let input = pipe.recv.recv(&ctx.with_deadline(deadline)).await.ok();
+                let input = pipe.recv.recv(ctx).await;
 
                 // We check if the context is active before processing the input. If the context is not active,
                 // we stop.
                 if !ctx.is_active() {
                     return Ok(());
                 }
 
-                // Check if an asynchronous error was received.
```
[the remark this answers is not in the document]
moshababo: The error sender is within a loop within a background task, so regular error propagation isn't feasible AFAIU. Also when received. 
However, as in the previous implementation, errors are not returned, only reported locally in tracing/metrics.
If we are to keep this, meaning to have no errors which would lead to termination, then the error channel and handling of non-cancel, internal errors is indeed not needed. 
Is that the case?

node/actors/bft/src/lib.rs
moshababo: Yes, especially if we don't need to be able to return errors (https://github.com/matter-labs/era-consensus/pull/56#discussion_r1462249244).

node/actors/bft/src/lib.rs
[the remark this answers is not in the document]
pompon0: if the background task fails with an error, then the whole scope gets cancelled and also fails with an error. Background tasks and regular tasks behave the same in that aspect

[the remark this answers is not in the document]
pompon0: so, if the error is critical then the background task can just fail, and if it is not then it can log the error directly.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Is there another way you'd recommend doing this, while only having `recv()`?

node/actors/bft/src/lib.rs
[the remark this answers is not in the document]
moshababo: I see, so since we don't need non-cancel error propagation support, this would be good enough. 

I removed the err channel and the bidirectional communication, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/2209d5dc32e7c6ce4195c19101faf4290b68535c

node/actors/bft/src/lib.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/2209d5dc32e7c6ce4195c19101faf4290b68535c

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
moshababo: Nice idea, but I removed the bidirectional communication, since it is not needed. Changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/2209d5dc32e7c6ce4195c19101faf4290b68535c

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Outdated now.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Fixed.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Outdated now.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Thanks, working now as you suggested.

node/libs/protobuf_build/src/lib.rs
```diff
@@ -183,6 +183,7 @@ impl Config {
     }
 
     /// Generates rust code from the proto files according to the config.
+    #[allow(clippy::map_identity)]
```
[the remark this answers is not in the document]
moshababo: Removed.

node/actors/bft/src/replica/state_machine.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/500b17f0ba75d3edebc77a04ae04d7de5a47fc5a

node/actors/bft/src/leader/state_machine.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/500b17f0ba75d3edebc77a04ae04d7de5a47fc5a

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Changed here: https://github.com/matter-labs/era-consensus/pull/56/commits/677b4677e24ebc4d9624214808492b32ff3b2b1a.
It disallow messages from the same type from the same validator altogether.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: It doesn't seem straightforward, but more like a hack.

moshababo: It doesn't seem straightforward, but more like a hack. The standard usage is to send new values.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: I understand the problem. Follow-up here: https://github.com/matter-labs/era-consensus/pull/56#discussion_r1462277267

moshababo: I understand the problem, assuming you only refer to the timeout in the test code. Follow-up here: https://github.com/matter-labs/era-consensus/pull/56#discussion_r1462277267

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Mutating instead of sending new values doesn't seem straightforward, but more like a hack.

moshababo: Mutating, instead of sending new values, doesn't seem straightforward, but more like a hack.

pompon0: watch is not a channel. In fact it just a name for a mutex with notifications.

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
pompon0: I'd suggest inspecting directly the channel state. But it wouldn't hurt to have a public try_recv() and/or empty() methods.

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: that's also not ok, because the validator should be allowed to override their previous message if our node is stuck processing. You may argue that it should be allowed to override it unconditionally, but that complicates reasoning about the bft algorithm.

moshababo: Sorry if it wasn't clear, the pruning/disallowing always refer to the pending messages. New messages are always accepted, hence overriding is supported. 

Not sure I understood your last sentence though.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/84a919cf0c41ca900b4c5a5f8bebba1d733d0133

moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/6d02da048b899ed944966d2c69d8b8230e036121

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Maybe I'm missing something, but I didn't find any `&mut` access from the sender.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/ba22e7400b699d7daf847bea570c36f65f9cd69f

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: This would test the buffer's state, but not the waiting functionality.

moshababo: This would test the channel's state, but not the waiting functionality.

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_if_modified

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
pompon0: well, recv() signature prevents it from returning without providing an element of T, which is not required to have any constructor, so you might say that a test is not needed. But if you really want to have a test of that you can add a small realtime timeout, i.e. call `recv.recv(&ctx.with_timeout(time::Duration::milliseconds(10))).await`

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Creates a channel and returns the [`Sender`] and [`Receiver`] handles.
+/// All values sent on [`Sender`] will become available on [`Receiver`] in the same order as it was sent,
+/// unless will be pruned before received.
+/// The Sender can be cloned to send to the same channel from multiple code locations. Only one Receiver is supported.
+///
+/// * [`T`]: The type of data that will be sent through the channel.
+/// * [`pruning_predicate`]: A function that determines whether an unreceived, pending value in the buffer should be pruned based on a newly sent value.
+///
+pub fn channel<T>(
+    pruning_predicate: impl 'static + Sync + Send + Fn(&T, &T) -> bool,
```
pompon0, line 30: it still isn't documented what is the semantics of the predicate: if predicate(a,b) is called and it returns true/false which element will be pruned? a or b?

node/libs/concurrency/src/sync/prunable_queue.rs
pompon0: ok,that's acceptable semantics then.

node/libs/concurrency/src/sync/prunable_queue.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/dd4f659e59ced72f76d0d190c667fb8bedd73a07

node/libs/concurrency/src/sync/prunable_mpsc/tests.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/era-consensus/pull/56/commits/ab5f4cb87a3bb95b024f5a9f80a0f6e43e2c43f9

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
moshababo: Added clarification here: https://github.com/matter-labs/era-consensus/pull/56/commits/8e4543cb1d2963472b466226d977c6972f9dbfec

PR #57, opened 2024-01-10
Consensus node Dockerfile for test framework 

IAvecilla:
## What ❔

Add Dockerfile to create a container with a consensus node running in it.

## Why ❔

This introduces a Dockerfile for the test framework proposal, enabling the execution of a node in isolation for potential use in future consensus among N nodes in the network. Additionally, a compose file is included for testing purposes, facilitating the simultaneous launch of two nodes that can initiate consensus, each running in a separate container. This file might be replaced in the future with a script that initializes N nodes, allowing for easier management, potentially with Kubernetes or a similar tool.


node/compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+    build: .
+    image: test-node
+    networks:
+      node_net:
+        # This allow us to know the ip of the node-1 container to fill the address in the config file
+        # Only for test purposes, may be removed in the future
+        ipv4_address: 172.12.0.10
+  node-2:
+    image: test-node
+    depends_on: [node-1]
```
pompon0, line 14: do we need this asymmetry?

node/tools/Makefile
```diff
@@ -1,12 +1,34 @@
...
+	echo 172.12.0.11:3054 >> addresses.txt
+	cd .. && docker-compose up -d
+
+clean: clean_docker
 	rm -rf node-configs
 	rm -rf database
+
+clean_docker:
+	docker rm test-consensus-node
+	docker rm node-node-1-1
```
pompon0, line 31: AFAIU these names are automatically assigned by docker. Can we make them more stable by specifying them explicitly somewhere?

node/tools/Makefile
```diff
@@ -1,12 +1,34 @@
...
+node:
 	export RUST_LOG=INFO && cd node-configs/${IP} && cargo run -- --database ./database/${IP} 
 
-node-configs:
+node_configs:
 	cargo run --bin localnet_config -- --input-addrs addresses.txt --output-dir node-configs
 
-clean:
+node_docker:
+	rm -rf addresses.txt
```
pompon0, line 12: can we make Makefile do not touch files in the source directory (do this stuff in /tmp or sth, I don't use makefiles much)? For example node_docker command overwrites the addresses.txt file that was required to be provided in the local node setup according to README.txt and I think it is not nice.

node/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Build Stage
+FROM rust:1.72.1 as build
```
pompon0, line 2: any reason why you put docker file in the node/ while makefile lives in node/tools ?

node/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY . /node/
```
pompon0, line 3: this is a bad idea, because node/target contains the local build artifacts which can be enormous.

node/tools/Makefile
```diff
@@ -1,12 +1,34 @@
...
+	echo 172.12.0.10:3054 >> addresses.txt
+	echo 172.12.0.11:3054 >> addresses.txt
+	cd .. && docker-compose up -d
+
+clean: clean_docker
 	rm -rf node-configs
 	rm -rf database
+
+clean_docker:
+	docker rm test-consensus-node
```
pompon0, line 30: docker rm -f ?

node/Dockerfile
pompon0: it would be nice to have it rebuilt only if the code changes

node/tools/Makefile
```diff
@@ -1,12 +1,34 @@
...
 
-node-configs:
+node_configs:
 	cargo run --bin localnet_config -- --input-addrs addresses.txt --output-dir node-configs
 
-clean:
+node_docker:
+	rm -rf addresses.txt
+	touch addresses.txt
+	echo ${DOCKER_IP}:3054 >> addresses.txt
```
pompon0, line 14: for consistency, how about
```
rm -rf addresses.txt
echo 172.12.0.10:3054 >> addresses.txt
cd .. && docker-compose up -d node-1
```

node/compose.yaml
```diff
@@ -0,0 +1,27 @@
+version: "3.9"
```
pompon0, line 1: docker compose will poorly generalize to N node setup, no?
Or are you planning to generate compose.yaml programmatically?

node/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt-get update && apt-get install -y libclang-dev
+RUN cargo build --release
+RUN cd tools && make node_configs
+
+# Runtime Stage
+FROM debian:stable-slim as runtime
+
+COPY --from=build /node/target/release/localnet_config /node/
+COPY --from=build /node/target/release/executor /node/
+COPY --from=build /node/tools/addresses.txt /node/
```
pompon0, line 14: this is not needed once you have the configs generated.

pompon0: same for localnet_config

node/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Build Stage
+FROM rust:1.72.1 as build
+COPY . /node/
+WORKDIR /node
+RUN apt-get update && apt-get install -y libclang-dev
+RUN cargo build --release
+RUN cd tools && make node_configs
```
pompon0, line 7: make node_configs builds with dev profile which unnecesarily increases the compilation time, since we have already everything built with release profile.

node/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Runtime Stage
+FROM debian:stable-slim as runtime
+
+COPY --from=build /node/target/release/localnet_config /node/
+COPY --from=build /node/target/release/executor /node/
+COPY --from=build /node/tools/addresses.txt /node/
+COPY --from=build /node/tools/node-configs /node/
```
pompon0, line 15: how does the container know which config from node-configs to use?

node/tools/Makefile
IAvecilla: Oh sure, I forgot to change them. Fixed!

node/tools/Makefile
IAvecilla: Good point. I've added another directory with the config specifically for the nodes running in Docker. This way, Docker will only use this config, and the original `addresses.txt` will remain untouched.

node/compose.yaml
IAvecilla: Certainly, this `compose` file was intended for testing purposes, allowing quick deployment of two nodes to verify their communication within containers. With Kubernetes, we could potentially remove this without much issue, as the orchestration of containers using the node image would be handled there. I can remove it and not keep track of it if you consider it, but for now, it serves as a quick test.

node/tools/Makefile
IAvecilla: Fixed it!

node/tools/Makefile
IAvecilla: Nice! already done

node/Dockerfile
IAvecilla: I'm not entirely sure if I grasp this correctly. I completely agree that copying the entire node crate into Docker could lead to problems. Would moving everything outside the `node` directory help alleviate this? Currently, the entire crate is built within a Docker image, and the final image only includes the generated binaries. I attempted to compile everything outside Docker and only copy the `target` directory, but that requires cross-compilation to x86_64 architecture if run on a Mac or another ARM machine. It's technically feasible, but I'm uncertain if we intend to pursue this, considering potential additional dependencies. Let me know if my understanding is accurate or if there's something I might be overlooking.

pompon0: I would recommend copying everything EXCEPT the target directory to avoid pulling in local artifacts. Alternatively you can try to pull in only files tracked by git (git ls-tree) but I don't know if that's an easy thing to do.

node/compose.yaml
pompon0: ack, that's fine

node/Dockerfile
pompon0: moving your docker-related configs outside of the node directory will prevent rebuilding the image when you change the config. Ideally it would be great to have the "cargo build" command split into building dependencies and the local code, in a way that the image layer with dependencies' artifacts is not rebuilt every time you update the local code (I don't know how to do that though without changing the build system, so this is just a thought for the future).

IAvecilla: Okay got it. I added the `node/target` dir to the `.dockerignore` file so it should not be cloned inside the containers. Also I moved every docker related file and the makefile to the root of the project and adapted every script and command to work with the new paths. Let me know if I missed something

IAvecilla: Regarding incorporating dependencies and local code in different Docker layers to cache the initial ones, it seems feasible. However, as you mentioned, this would entail altering the structure of the `Cargo.toml`. My initial thought on achieving this is to include only external dependencies in the `Cargo.toml` of the `node` crate. Then, within the `Cargo.toml` of each distinct crate, build every member of the workspace. This way, we could initially build all the external dependencies to cache them and subsequently build the actual node code. I plan to explore this in a separate pull request to avoid making unnecessary changes for this one.

node/tools/README.md
```diff
@@ -1,5 +1,27 @@
...
+
+To get up a standalone consensus node running in a docker container just run the following command inside the tools crate:
+
+`make node_docker`
+
+This will create a container running a single node advancing views and finalizing blocks.
+
+To set up a simple example with two different nodes communicating with each other in running in different containers run the following command:
+
+`make consenus_docker_example`
```
brunoffranca, line 19:
```suggestion
`make consensus_docker_example`
```

Dockerfile
```diff
@@ -1,19 +1,17 @@
 # Build Stage
 FROM rust:1.72.1 as build
-COPY . /node/
+COPY /node/ Makefile /node/
```
pompon0, line 3: do I understand correctly that this way Makefile will be executed from a different location from within docker and locally? Please avoid that either by copying Makefile to the root, or by moving it to node/ directory.

pompon0 (approved):

Dockerfile
IAvecilla: Yeah, I fixed it to be in the root of the container, just like on the local machine.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Build Stage
+FROM rust:1.72.1 as build
+COPY /node/ /node/
+COPY Makefile .
+WORKDIR /node
+RUN apt-get update && apt-get install -y libclang-dev
+RUN cargo build --release
```
moshababo, line 7: Following the discussion you had about caching:
The image build runs line 6 and 7 on every code change, which takes very long time. 
I think it's possible to better use layer caching so that at least line 6 won't be repeated (I'm not sure why it does).

As for line 7, as you already mentioned, it should be possible to avoid the reinstalling of dependencies, by only copying `Cargo.toml` and `Cargo.lock` files, then create a dummy `main.rs` (`RUN echo "fn main() {println!(\"dummy\")}" > src/main.rs`) and do cargo build to cache the dependencies. Then, copy the rest of the source code and build again. 

This is not critical for now, but might worth trying.

moshababo (commented):
Looks good in the overall!

Dockerfile
pompon0: Cargo.toml is a workspace manifest, so it would need to be converted to a crate manifest.

node/tools/README.md
```diff
@@ -1,5 +1,27 @@
-# Running a test node
+# Running a test consensus node
+
+## Local
 
 1. Generate a file named `addresses.txt` in the root directory of the tools crate, containing node addresses in the format `IP:PORT`, with each address on a separate line.
-2. Run `make node-configs`. This command will establish a directory named `node-configs` and create a folder for each address listed in the `.txt` file, providing the necessary configuration files for the respective node.
```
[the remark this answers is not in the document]
IAvecilla: You're right, with new changes in the directories and overall structure I forgot to update the README. Now it should be updated including the last changes added from this comments. Let me know if it's more clear now.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+node:
+	export RUST_LOG=INFO && cd ${EXECUTABLE_NODE_DIR}/node-configs/${IP} && cargo run -- --database ../../database/${IP} 
+
+node_configs:
+	cd ${EXECUTABLE_NODE_DIR} && cargo run --bin localnet_config -- --input-addrs addresses.txt --output-dir node-configs
+
+# Docker commands
+
+docker_node_configs:
```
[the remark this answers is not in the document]
IAvecilla: This command shouldn't be important for the user. It is used by Docker in the initial phase of the image building process to generate the configuration. However, I added a comment above the command with this disclaimer.

compose.yaml
```diff
@@ -0,0 +1,28 @@
...
+    image: consensus-node
+    container_name: consensus-node-1
+    networks:
+      node_net:
+        # This allow us to know the ip of the node-1 container to fill the address in the config file
+        # Only for test purposes, may be removed in the future
+        ipv4_address: 172.12.0.10
+  node-2:
+    image: consensus-node
+    container_name: consensus-node-2
```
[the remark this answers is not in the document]
IAvecilla: I checked it, and it was my initial choice. The thing about that feature is that it duplicates the exact same container, and I needed to manually assign IPs to the containers to run the example in a single command where I can write the IPs in the config file, knowing them before running the container. Perhaps I overlooked something, and it's still feasible to use it.

entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd $(hostname -i):3054
+export RUST_LOG=INFO
+../executor
```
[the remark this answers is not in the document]
IAvecilla: I renamed the file to `docker-entrypoint` and be more clear about what is its purpose. I added a comment anyways to explain how it is used

Dockerfile
IAvecilla: Exactly, that was my initial approach for caching external dependencies but since it works as a workspace manifest all the local code would be necessary to build along them.

brunoffranca (approved):

PR #59, opened 2024-01-19
Test RPC layer for Standalone Node

IAvecilla:
## What ❔

Introduce an RPC layer for the Standalone node, including a test server featuring an endpoint for health-checking the node, and a CLI client for making requests to this server.

## Why ❔

This layer becomes necessary for communication with nodes within a cluster, especially for running tests that involve communication with these nodes. Additionally, it could serve as the initial implementation for a signaling or communication system within the test module, facilitating the determination of test success or failure.


IAvecilla:
@pompon0 I think there's some issues with `cargo-deny` probably because of the new `jsonrpsee` crate. LMK if you want me to change the `deny.toml` of you want another way to solve this kind of issues.

brunoffranca:
The offending crate seems to be `ring` actually, not `jsonrpsee`. It has a non-standard license so cargo deny can't read it. It's enough to just clarify the license in the deny.toml file. See `licenses.clarify` in the [configuration](https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html) (funnily enough, the example they give is the ring crate).

pompon0:
please fix the license warning, just like Bruno said. Please fix the "multiple crate versions" warning by upgrading the crate versions in Cargo.toml where feasible. For the rest of packages, add the old versions to the deny.toml following the examples that we already have there.

IAvecilla:
I managed to fix it by only adding the older versions of the conflicting crates. I didn't include the license for the `ring` crate since I didn't see any warning for that. Additionally, I removed an old version of `itertools` that was causing a warning.

brunoffranca (approved):

pompon0 (approved):

PR #60, opened 2024-01-22
made max_payload_size configurable

pompon0:
The max_payload_size is checked in bft replica and also implies buffering limits in the network actor (both gossip and consensus networks). It turned out that hardcoding it was too optimistic strategy, so I've extracted it to a config, so that it is settable in runtime. Given the lesson learned, I'll be also deprecating the `ProtoFmt::max_size()` function soon.
Additional changes:
* added some logging useful for debugging the zksync-era tests and improved existing logging to always show error the cause, not just the leaf message.
* removed `tracing::instrument` annotations which were logging non-critical/not-a-bug errors on level ERROR
* made network crate depend on the storage crate to avoid the workaround way of implementing the get_block rpc
* cleaned up the implementations of the rpcs to be more consistent
* deduplicated the testonly helpers and moved them to the crates that they belong to


brunoffranca (approved):
Great work as always. No notes.

PR #61, opened 2024-01-25
Added `--seed` cli argument to provide IP address, port and key of seed peers on startup

ElFantasma:
## What ❔

Adds a `--seed` command line argument to support seed peer address and key injection when starting the node

Example:
```
executor --seed 10.123.12.21:3043,node:public:ed25519:3059a9ae1de665f2e5dfcb64d5d3a044b2b3617436ddd56c2baf896878b13961 --seed 10.23.22.0:1010,node:public:ed25519:07928c09f2828cbbe1bacc9dc00271c4517c3af1dad034054591e6af10450338
```
will add two IP and key tuples to the `gossipStaticOutbound` in the `config.json` file when starting the node (and it will override the address if the same key is present in the file).

## Why ❔

This will be necessary for the test framework to be able to start nodes within a k8s cluster when initially the IP addresses assigned are unknown. In that sense, we will deploy an initial set of Seed Peers, retrieve their IP addresses and then deploy the rest of the nodes injecting the seed peer addresses and keys via command line arguments.

### DIFF
--- a/node/tools/src/config.rs
+++ b/node/tools/src/config.rs
@@ -118,13 +118,6 @@ impl ProtoFmt for AppConfig {
     }
 }
 
-impl AppConfig {
-    pub fn add_seed_peers(&mut self, seed_peers: Vec<(std::net::SocketAddr, node::PublicKey)>) {
-        for (addr, key) in seed_peers.iter() {
-            self.gossip_static_outbound.insert(key.clone(), *addr);
-        }
-    }
-}
 /// This struct holds the file path to each of the config files.
 #[derive(Debug)]
 pub struct ConfigPaths<'a> {
--- a/node/tools/src/config.rs
+++ b/node/tools/src/config.rs
@@ -5,7 +5,6 @@ use std::{
     collections::{HashMap, HashSet},
     fs,
     path::{Path, PathBuf},
-    str::FromStr,
 };
 use zksync_concurrency::ctx;
 use zksync_consensus_bft as bft;
@@ -13,12 +12,12 @@ use zksync_consensus_crypto::{read_optional_text, read_required_text, Text, Text
 use zksync_consensus_executor as executor;
 use zksync_consensus_roles::{node, validator};
 use zksync_consensus_storage::{BlockStore, BlockStoreRunner, PersistentBlockStore};
-use zksync_protobuf::{required, ProtoFmt};
+use zksync_protobuf::{required, serde::Serde, ProtoFmt};
 
 /// Decodes a proto message from json for arbitrary ProtoFmt.
-fn decode_json<T: ProtoFmt>(json: &str) -> anyhow::Result<T> {
+pub fn decode_json<T: serde::de::DeserializeOwned>(json: &str) -> anyhow::Result<T> {
     let mut d = serde_json::Deserializer::from_str(json);
-    let p: T = zksync_protobuf::serde::deserialize(&mut d)?;
+    let p = T::deserialize(&mut d)?;
     d.end()?;
     Ok(p)
 }
@@ -34,8 +33,8 @@ impl ProtoFmt for NodeAddr {
     type Proto = proto::NodeAddr;
 
     fn read(r: &Self::Proto) -> anyhow::Result<Self> {
-        let key = read_required_text(&r.key)?;
-        let addr = read_required_text(&r.addr)?;
+        let key = read_required_text(&r.key).context("key")?;
+        let addr = read_required_text(&r.addr).context("addr")?;
         Ok(Self { addr, key })
     }
 
@@ -44,14 +43,6 @@ impl ProtoFmt for NodeAddr {
             key: Some(TextFmt::encode(&self.key)),
             addr: Some(TextFmt::encode(&self.addr)),
         }
-    }
-}
-
-impl FromStr for NodeAddr {
-    type Err = anyhow::Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        decode_json(s)
     }
 }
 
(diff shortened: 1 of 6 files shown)


node/tools/src/config.rs
```diff
@@ -118,6 +118,13 @@ impl ProtoFmt for AppConfig {
     }
 }
 
+impl AppConfig {
+    pub fn add_seed_peers(&mut self, seed_peers: Vec<(std::net::SocketAddr, node::PublicKey)>) {
```
pompon0, line 122: from the pov of the config, seed peers do not differ at all from regular peers afaiu. No special method is required and you can just use HashMap::extend method to insert these. Instead of "seed" for a flag name use the name of the field that you are appending to, and prefix it with "append" or sth to indicate that the flag is for appending rather than overriding for example. keep the (key,ip) order (rather than (ip,key) in the flag spec to be consistent with the config format.

node/tools/src/main.rs
```diff
@@ -41,6 +46,19 @@ impl Args {
     }
 }
 
+/// Parse a single (SocketAddr,node key) pair
+fn parse_seed_peer(
+    s: &str,
+) -> Result<(std::net::SocketAddr, node::PublicKey), Box<dyn Error + Send + Sync>> {
+    let pos = s
+        .find(',')
+        .ok_or_else(|| format!("expected format: <IP>:<Port>,<key>"))?;
```
pompon0, line 55: nit: consider using json encoding rather than manual parsing for the flag values. You can even reuse the Addr struct from the config.
nit: alternatively please consider whether passing just the whole json encoded config as a flag wouldn't be simpler and more flexible

node/tools/src/config.rs
ElFantasma: 👍 Makes sense

node/tools/src/main.rs
ElFantasma: I think it will be good to use json encoding for these pair (SocketAddr, PublicKey) values. It will make the command line argument more consistent and easy to build. 

It may be also useful to pass the whole config as a json encoded, but I don't think it will be necessary for this case and it may overcomplicate the k8s entrypoint call, as well as the file-config vs. cli-arg overriding code.

So if it is ok with you I think I will just implement the json encoding for the `gossip_static_outbound` field.

In order to do this, I'm not sure I understand what Addr struct you are referring to. Is it proto::NodeAddr? Not sure if it is straightforward to deserialize it (maybe I have to make it implement some trait) and even if I do, it will return strings instead of PublicKey and SocketAddr.

Maybe it's cleaner to use a new custom struct for this?

pompon0: Yes, I  meant `proto::NodeAddr`. Currently it is decoded directly in the `config::AppConfig` deserializer: https://github.com/matter-labs/era-consensus/blob/a766bd2b981e08eb106c3c787cca023f7d8f48e4/node/tools/src/config.rs#L65 , but you can create a dedicated `config::NodeAddr` struct and implement `ProtoFmt` trait for it (by just extracting the related code from `ProtoFmt` implementation for `config::AppConfig`). The json decoding function is here: https://github.com/matter-labs/era-consensus/blob/a766bd2b981e08eb106c3c787cca023f7d8f48e4/node/tools/src/config.rs#L18

node/tools/src/main.rs
```diff
@@ -2,14 +2,41 @@
...
 use vise_exporter::MetricsExporter;
 use zksync_concurrency::{ctx, scope};
+use zksync_consensus_crypto::read_required_text;
+use zksync_consensus_roles::node;
 use zksync_consensus_tools::ConfigPaths;
 use zksync_consensus_utils::no_copy::NoCopy;
 
+/// Utility struct to parse json value from cli arg
+#[derive(Debug, Clone)]
+struct NodeAddresses(Vec<(node::PublicKey, std::net::SocketAddr)>);
```
pompon0, line 17: please move to config crate and implement `ProtoFmt` for it instead. Deduplicate the code with `ProtoFmt` implementation for `AppConfig`.

node/tools/src/main.rs
```diff
@@ -2,14 +2,41 @@
...
+
+impl FromStr for NodeAddresses {
+    type Err = String;
```
pompon0, line 20: use anyhow::Error for errors which don't need to be distinguished in runtime.

node/tools/src/main.rs
```diff
@@ -26,6 +53,9 @@ struct Args {
     /// Path to the rocksdb database of the node.
     #[arg(long, default_value = "./database")]
     database: PathBuf,
+    /// IP address and key of the seed peers.
+    #[arg(long = "add_gossip_static_outbound")]
+    gossip_static_outbound: NodeAddresses,
```
pompon0, line 58: add_gossip_static_outbound

ElFantasma:
Thanks for the review @pompon0. Implemented most of your suggestions. 

The only thing I could not get to work is to be able to provide the `--gossip_static_outbound` as a single json list.
like:
```
--add-gossip-static-outbound '[{"key": "node:public:ed25519:...", "addr": "..."}, {"key": "node:public:ed25519:...", "addr": "..."}]'
```
Instead, now we have to do:
```
--add-gossip-static-outbound '{"key": "node:public:ed25519:...", "addr": "..."} --add-gossip-static-outbound '{"key": "node:public:ed25519:...", "addr": "..."}'
```

Any idea how to address this? Found [this](https://users.rust-lang.org/t/clap-derive-arg-to-parse-json-array-of-objects-into-vec-mystruct/97975) but I'm not happy with none of that "solutions"

pompon0:
Yeah, that seems annoying. I've pushed a commit with my suggestion, which combines this and previous approach.

pompon0 (approved):

ElFantasma:
Nice, @pompon0! thank you

PR #63, opened 2024-02-06
Sanity Test for nodes running in k8s (BFT-394)

IAvecilla:
## What ❔

Add a first sanity test to check node liveness running with the new Kubernetes script.

## Why ❔

We need this basic test to see how to address the test suite development and have a layer used to communicate a local environment (like CI) with the nodes running in kubernetes.


node/tests/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+    GenerateConfig,
+    /// Set up the test pod.
+    StartPod,
+    /// Deploy the nodes.
+    Run,
+}
+
+/// Get the path of the node ips config file.
+/// This way we can run the test from every directory and also inside kubernetes pod.
+fn get_config_path() -> String {
```
pompon0, line 33: std::path::Path

node/tools/src/k8s.rs
```diff
@@ -21,6 +21,25 @@ pub async fn get_client() -> anyhow::Result<Client> {
     Ok(Client::try_default().await?)
 }
 
+/// Get a kube client
+pub async fn get_consensus_node_ips(client: &Client) -> anyhow::Result<Vec<String>> {
```
pompon0, line 25: std::net::SocketAddr

node/tools/src/k8s.rs
```diff
@@ -21,6 +21,25 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
+    let lp = ListParams::default();
+    let pod = pods.list(&lp).await?;
+    let a: Vec<String> = pod
+        .into_iter()
+        .filter(|pod| {
+            pod.clone()
+                .metadata
+                .name
+                .unwrap()
+                .starts_with("consensus-node")
```
pompon0, line 36: Is this the most reliable way of selecting targets? Perhaps we can do an exact match on the docker image name?
Can we pass the prefix/image name as a command line argument?

node/tools/src/k8s.rs
```diff
@@ -63,6 +82,59 @@ pub async fn create_or_reuse_namespace(client: &Client, name: &str) -> anyhow::R
     }
 }
 
+pub async fn create_tests_deployment(client: &Client) -> anyhow::Result<()> {
+    let deployment: Deployment = serde_json::from_value(json!({
```
pompon0, line 86: Why hardcoding json? It should be feasible to construct a Deployment object directly, no?

node/tests/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+    let client = k8s::get_client().await.unwrap();
+    k8s::create_tests_deployment(&client).await.unwrap();
+}
+
+/// Sanity test for the RPC server.
+pub async fn sanity_test() {
+    let config_file_path = get_config_path();
+    let node_ips = fs::read_to_string(config_file_path).unwrap();
+    for ip in node_ips.lines() {
+        let url: String = format!("http://{}:3154", ip);
```
pompon0, line 70: do we need to hardcode the port? Or can it be fetched from pod metadata? If not, please make it command line argument.

node/tests/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+    let pods_ip = k8s::get_consensus_node_ips(&client).await.unwrap();
+    let config_file_path: String = get_config_path();
+    for ip in pods_ip {
+        let mut config_file = fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(&config_file_path)
+            .unwrap();
+        config_file.write_all(ip.as_bytes()).unwrap();
```
pompon0, line 55: if it can fail - propagate the error. If it cannot, document why unwrap is safe.

node/tools/src/k8s.rs
IAvecilla: Yeah, I don't like the idea of hardcoded JSON either, but I was avoiding the other approach using the `kube` crate because it adds a lot of boilerplate, at least for this specific deployment. I changed it anyway since it's a pretty straightforward change. It adds a lot of `Default::default()` for the fields that aren't needed and a lot of `Options` for the optional fields. Maybe we can explore other options or decide on whichever you think is better.

node/tests/src/main.rs
IAvecilla: I changed it to fetch directly from the pod, but it's still a bit of a suboptimal solution. I'm currently getting all the exported ports of each pod and selecting the second one, which is supposed to be the port for the RPC server (as the first one is for node communication). While this approach is functional, specifying it as a command-line argument might be a better solution.

node/tests/src/main.rs
IAvecilla: All the `unwrap` calls should be handled now. I've left the ones in the `sanity_test` function since they should function as intended for testing purposes. I'll leave a comment explaining this for clarity.

node/tools/src/k8s.rs
```diff
@@ -22,22 +25,38 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
-                .unwrap()
-                .starts_with("consensus-node")
+            if let Some(pod) = pod.clone().metadata.name {
+                pod.contains("consensus-node")
+            } else {
+                false
+            }
+        })
+        .map(|pod| {
+            let pod_ip = pod.status.and_then(|status: PodStatus| status.pod_ip);
```
pompon0, line 42: `Option` implements `anyhow::Context`, so you can just use `pod.status.context("status")?.context("pod_ip")?` etc.

node/tools/src/k8s.rs
```diff
@@ -83,43 +102,54 @@ pub async fn create_or_reuse_namespace(client: &Client, name: &str) -> anyhow::R
...
-              }
-            ]
-          }
-        }
-      }
-    }))?;
+        spec: Some(DeploymentSpec {
+            selector: LabelSelector {
+                match_labels: Some(
+                    [("app".to_string(), "test-node".to_string())]
```
pompon0, line 120: `[("app".to_string(),"test-node".to_string())].into()` should work, no?

node/tools/src/k8s.rs
```diff
@@ -22,22 +25,38 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
+            let port = pod.spec.and_then(|spec: PodSpec| {
+                spec.containers[0].clone().ports.and_then(|ports| {
+                    ports
+                        .iter()
+                        .find(|port| port.container_port != config::NODES_PORT as i32)
+                        .map(|port| port.container_port)
+                })
+            });
+            if let Some((pod_ip, port)) = pod_ip.zip(port) {
+                let pod_addr = format!("{}:{}", pod_ip, port);
```
pompon0, line 52: SocketAddr::new(pod_ip.parse()?,port)

node/tools/src/k8s.rs
pompon0, line 42: argument type annotations in these closures shouldn't be necessary, no?

node/tools/src/k8s.rs
```diff
@@ -22,22 +25,38 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
-pub async fn get_consensus_node_ips(client: &Client) -> anyhow::Result<Vec<String>> {
+pub async fn get_consensus_node_ips(client: &Client) -> anyhow::Result<Vec<SocketAddr>> {
     let pods: Api<Pod> = Api::namespaced(client.clone(), "consensus");
     let lp = ListParams::default();
     let pod = pods.list(&lp).await?;
-    let a: Vec<String> = pod
+    let a: Result<Vec<SocketAddr>, _> = pod
         .into_iter()
         .filter(|pod| {
-            pod.clone()
```
pompon0: don't clone, use `pod.metadata.name.as_ref().unwrap().starts_with()`

node/tools/src/k8s.rs
```diff
@@ -22,22 +25,38 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
-                .metadata
-                .name
-                .unwrap()
-                .starts_with("consensus-node")
+            if let Some(pod) = pod.clone().metadata.name {
+                pod.contains("consensus-node")
```
pompon0, line 36: `if let Some(pod) = &pod.metadata.name`

node/tools/src/k8s.rs
IAvecilla: I liked the idea of matching the `consensus-node` Docker image, so I went with that approach.

node/tools/src/k8s.rs
IAvecilla: Nah, I didn't like the JSON either, so we can go with this solution. Also, I think it has more room for improvements or refactors in the future if needed.

node/tests/src/main.rs
```diff
@@ -85,11 +89,11 @@ async fn main() {
 
     match args.command {
         TesterCommands::GenerateConfig => {
-            generate_config().await;
+            let _ = generate_config().await;
```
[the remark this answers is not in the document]
IAvecilla: Right, I added context for errors in each of the commands and made the main function return a `Result`. This way, if one of the commands fails, it won't continue with the following commands when using the `Makefile` target.

node/tools/src/k8s.rs
```diff
@@ -22,22 +25,38 @@ pub async fn get_client() -> anyhow::Result<Client> {
...
+            } else {
+                false
+            }
+        })
+        .map(|pod| {
```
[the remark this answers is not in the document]
IAvecilla: I needed to tweak the code a bit, but it's working well now

node/tools/src/k8s.rs
IAvecilla: The `context` function casts the `Option` to `Result`, and the `filter_map` closure has to return an Option, so I use `context("status").ok()?.pod_ip`. I also replicated this block for all the options that the `kube` library returns.

IAvecilla: Right, I already removed them

node/tests/src/main.rs
```diff
@@ -30,44 +31,52 @@ enum TesterCommands {
...
-fn get_config_path() -> String {
+fn get_config_path() -> PathBuf {
     // This way we can run the test from every directory and also inside kubernetes pod.
     let manifest_path = std::env::var("CARGO_MANIFEST_DIR");
     if let Ok(manifest) = manifest_path {
-        format!("{}/config.txt", manifest)
+        PathBuf::from(&format!("{}/config.txt", manifest))
     } else {
-        "config.txt".to_owned()
+        PathBuf::from("config.txt")
```
pompon0, line 40: nit: `"config.txt".into()`

node/tests/src/main.rs
```diff
@@ -30,44 +31,52 @@ enum TesterCommands {
 
 /// Get the path of the node ips config file.
 /// This way we can run the test from every directory and also inside kubernetes pod.
-fn get_config_path() -> String {
+fn get_config_path() -> PathBuf {
     // This way we can run the test from every directory and also inside kubernetes pod.
     let manifest_path = std::env::var("CARGO_MANIFEST_DIR");
     if let Ok(manifest) = manifest_path {
-        format!("{}/config.txt", manifest)
+        PathBuf::from(&format!("{}/config.txt", manifest))
```
pompon0, line 38: prefer `Path::join` to `format`

node/tests/src/main.rs
```diff
@@ -30,44 +31,52 @@ enum TesterCommands {
 
 /// Get the path of the node ips config file.
 /// This way we can run the test from every directory and also inside kubernetes pod.
-fn get_config_path() -> String {
+fn get_config_path() -> PathBuf {
     // This way we can run the test from every directory and also inside kubernetes pod.
     let manifest_path = std::env::var("CARGO_MANIFEST_DIR");
     if let Ok(manifest) = manifest_path {
-        format!("{}/config.txt", manifest)
```
pompon0: nit: `Path::new(manifest_path.unwrap_or("")).join("config.txt")`

node/tests/src/main.rs
```diff
@@ -30,44 +31,52 @@ enum TesterCommands {
...
+    let config_file_path = get_config_path();
+    let mut config_file = fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(&config_file_path)?;
+    for addr in pods_ip {
+        config_file
+            .write_all(format!("{}\n", addr.to_string()).as_bytes())
+            .with_context(|| "Failed to write to config file")?;
```
pompon0, line 59: just `context("...")`

node/tests/src/main.rs
```diff
@@ -30,44 +31,52 @@ enum TesterCommands {
...
 
 /// Sanity test for the RPC server.
+/// We use unwraps here because this function is intended to be used like a test.
 pub async fn sanity_test() {
     let config_file_path = get_config_path();
-    let node_ips = fs::read_to_string(config_file_path).unwrap();
-    for ip in node_ips.lines() {
-        let url: String = format!("http://{}:3154", ip);
+    let nodes_socket = fs::read_to_string(config_file_path).unwrap();
+    for socket in nodes_socket.lines() {
```
pompon0, line 78: first parse the ips, rather than blindly turning them into urls

node/tools/src/k8s.rs
```diff
@@ -1,43 +1,82 @@
...
+                None
+            }
         })
-        .map(|pod| pod.status.unwrap().pod_ip.unwrap())
         .collect();
-    Ok(a)
+    ensure!(
+        !pod_addresses.is_empty(),
+        "No consensus pods found in the k8s cluster"
+    );
```
pompon0, line 78: nit: that should be rather checked by the caller, who will inspect the result

node/tools/src/k8s.rs
```diff
@@ -1,43 +1,82 @@
...
-                .metadata
-                .name
-                .unwrap()
-                .starts_with("consensus-node")
+        .filter_map(|pod| {
+            let pod_spec = pod.spec.context("Failed to get pod spec").ok()?;
+            let pod_running_container = pod_spec
+                .containers
+                .first()
+                .context("Failed to get pod container")
```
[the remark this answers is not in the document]
IAvecilla: Okay, you're correct that returning a `Result` from this function is no longer useful when using `filter_map`. I made the change using `filter` followed by `map`, but the errors that arise in the `filter` closure are still being ignored because it requires a boolean and using `?` to return a `Result` is not possible. Let me know if this is acceptable. In the `map` closure, the errors should now be handled correctly.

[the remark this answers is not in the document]
pompon0: if you need to return an error from filter, then just use a regular loop over the elements instead.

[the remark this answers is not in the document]
IAvecilla: I ended up including everything in the for loop instead of having the for loop only for filtering and then mapping. I chose this solution mainly because:

- I can reuse some pod variables instead of retrieving them twice, once for filtering and then again for mapping and address parsing.
- This way, I only iterate over all the pods once. With the separate map approach, I had to create an empty vector, push all the filtered pods into it, and then convert it to an iterator for the address parsing part.

Let me know what you think.

node/tools/src/k8s.rs
```diff
@@ -39,61 +44,54 @@ pub async fn get_consensus_nodes_address(client: &Client) -> anyhow::Result<Vec<
...
-                    let port = port.container_port.try_into().ok()?;
+    let mut node_rpc_addresses: Vec<SocketAddr> = Vec::new();
+    for pod in pods.into_iter() {
+        let pod_container = pod
+            .clone()
+            .spec
+            .context("Failed to get pod spec")?
+            .containers
+            .first()
+            .cloned()
```
pompon0, line 55: first().as_ref().context()

PR #67, opened 2024-02-21
Node state for the RPC server

IAvecilla:
## What ❔

Make the `ReplicaState` struct and additional information about the node state available on the RPC server side.

## Why ❔

For future tests, it would be necessary to have access to the state of different running nodes in order to assess the behavior of the entire consensus network. This could involve getting different metrics such as the last saved view for each node, proposals, highest vote or quorum certificate (QC), etc.


node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -417,7 +418,7 @@ impl ValidatorSet {
 }
 
 /// A struct that represents a view number.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
```
pompon0, line 421: Don't do that. Just extract the int

node/tools/src/rpc/methods/last_view.rs
```diff
@@ -0,0 +1,22 @@
+//! Peers method for RPC server.
+use jsonrpsee::core::RpcResult;
+use std::sync::Arc;
+use zksync_consensus_storage::{BlockStore, ReplicaState};
+
+/// Config response for /config endpoint.
```
pompon0, line 6: fix comments

pompon0:
You seem to expose just the view number from the replica state. Is that intended?

node/tools/src/rpc/methods/peers.rs
```diff
@@ -1,41 +1,33 @@
...
-        let peers: Vec<String> = node_config
-            .gossip_static_inbound
-            .iter()
-            .map(|x| x.encode())
-            .collect();
-        Ok(serde_json::json!({
-            "peers": peers
-        }))
-    }
+/// Peers response for /peers endpoint.
```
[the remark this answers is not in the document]
IAvecilla: Actually, we don't need it anymore. With the recent changes in the initial configuration for the consensus node, the `config` endpoint is also no longer necessary.

IAvecilla:
> You seem to expose just the view number from the replica state. Is that intended?

Yes, it is intended. In the future, we may require additional information from the storage. I've introduced another endpoint that could prove useful, especially considering the basic tests planned with the `chaos-mesh` tool. The intention is to incorporate more endpoints as we develop further tests and additional information from the node is needed.

pompon0:
I mean, why not just dump the whole ReplicaState?

node/tools/src/k8s.rs
```diff
@@ -39,61 +44,53 @@ pub async fn get_consensus_nodes_address(client: &Client) -> anyhow::Result<Vec<
...
+                .context("Failed to get pod status")?
+                .pod_ip
+                .context("Failed to get pod ip")?;
+            let pod_rpc_port = pod_container
+                .ports
+                .as_ref()
+                .context("Failed to get ports of container")?
+                .iter()
+                .find_map(|port| {
+                    let port: u16 = port.container_port.try_into().ok()?;
```
pompon0, line 71: shouln't failed conversion return an error?

IAvecilla: The overall operation will fail due to the `.context("Failed parsing container port")?;` below. If the conversion fails then a `None` is returned, if every port returns `None` in the `find_map` operation, the context will raise an error because there was no port that met the criteria. While it's true that if the conversion succeeds but there's no port different from `NODE_PORT`, the error saying `Failed parsing container port` will also be thrown. I can separate the `find` and `map` operations, and that should fix it, but let me know what you think.

IAvecilla:
> I mean, why not just dump the whole ReplicaState?

Is there a way to obtain the replica state from the executor? With the latest changes to the `ReplicaState`, I don't see a clear way to do this. That's why I went with the solution to obtain the `BlockStore` and then pick the view and the last message through the `CommitQC`. If obtaining the replica state is indeed feasible, that would be a better approach.

node/tools/src/k8s.rs
pompon0: wrt correctness, this should fail if anything in the response received is different than expected, even if the given part of the response is unused. Said that, it is not that important in testing code, so up to you.

pompon0:
> Is there a way to obtain the replica state from the executor? With the latest changes to the ReplicaState, I don't see a clear way to do this. That's why I went with the solution to obtain the BlockStore and then pick the view and the last message through the CommitQC. If obtaining the replica state is indeed feasible, that would be a better approach.

Right, you would need to pass ReplicaStore to the rpc server, which is not cloneable as of today. It is fine as is then.

pompon0 (approved):

PR #69, opened 2024-02-27
Improve get addrs

ElFantasma:
## What ❔
Fixes https://linear.app/matterlabs/issue/BFT-429/improve-get-seed-node-addrs-function
Created a `ConsensusNode` struct to handle the internals of a k8s running node as well as its temporary configuration

## Why ❔

The idea is to have a more object oriented code as stated in a comment in main [k8s deployment script PR](https://github.com/matter-labs/era-consensus/pull/62#discussion_r1490241441)
Also, this `ConsensusNode` struct will be useful to handle the whole node configurations set for https://linear.app/matterlabs/issue/BFT-427/provide-full-config-as-a-cli-argument


node/tools/src/k8s.rs
```diff
@@ -172,30 +187,32 @@ pub async fn deploy_node(
     Ok(())
 }
 
-/// Returns a HashMap with mapping: node_id -> IP address
-pub async fn get_seed_node_addrs(
+/// Waits for the pods to start in order to complete each ConsensusNode with its assigned NodeAddr
+pub async fn find_node_addrs(
```
moshababo, line 191: This should become a method which can be applied to a given `ConsensusNode` instance. Also, shouldn't be associated with the notion of seed nodes.

node/tools/src/k8s.rs
```diff
@@ -16,11 +16,29 @@ use kube::{
...
+use std::str::FromStr;
 use tokio_retry::strategy::FixedInterval;
 use tokio_retry::Retry;
 use tracing::log::info;
+use zksync_consensus_roles::node;
 use zksync_protobuf::serde::Serde;
 
+/// Consensus Node Representation
+#[derive(Debug)]
+pub struct ConsensusNode {
```
moshababo, line 29: This should eventually contain the RPC client and all other means of communicating with / using the node programmatically in the test framework.

node/tools/src/k8s.rs
ElFantasma: On one hand, method may be useful, but on the other this way is more efficient as it queries the cluster for a set of nodes in a single request (and waits for all of them to be available and running on an assigned IP address).

If this was a method in a given `ConsensusNode` we will have to query and wait for each pod individually instead

node/tools/src/k8s.rs
ElFantasma: Yes, and I'm also refactoring it in another PR to contain the full configuration and keys to be provided by command line in pod entrypoints, in order to build the image without configuration files.

node/tools/src/k8s.rs
moshababo: Optimizing for efficiency isn't very important here. (and it can be done later, if needed)
There's a need for a basic functionality in which you can `.waitPod()` on a given node until ready, to perform various operations thereafter. (obtain IP address, connect RPC client, etc.).
This is needed for all nodes, not just seed nodes.

node/tools/src/k8s.rs
```diff
@@ -66,145 +220,19 @@ pub async fn create_or_reuse_namespace(client: &Client, name: &str) -> anyhow::R
...
+        .await?
+        .iter()
+        .next()
+        .context(format!("Pod not found: {label}"))
+        .cloned()?;
+    if is_pod_running(&pod) {
+        Ok(pod)
     } else {
-        Err(anyhow!("Pods are not ready"))
+        Err(anyhow!("Pod not ready"))
```
pompon0, line 235: anyhow::format_err!

node/tools/src/k8s.rs
```diff
@@ -12,15 +12,169 @@ use k8s_openapi::{
...
+        let ip = self
+            .await_running_pod(client, namespace)
+            .await?
+            .status
+            .context("Status not present")?
+            .pod_ip
+            .context("Pod IP address not present")?;
+        self.node_addr = Some(NodeAddr {
+            key: self.key.public(),
+            addr: SocketAddr::from_str(&format!("{}:{}", ip, config::NODES_PORT))?,
```
pompon0, line 73: SocketAddr::new(ip.parse()?,config::NODES_PORT)

node/tools/src/k8s.rs
```diff
@@ -66,145 +219,19 @@ pub async fn create_or_reuse_namespace(client: &Client, name: &str) -> anyhow::R
...
-    if p.items.len() == amount && p.iter().all(is_pod_running) {
-        Ok(p)
+async fn get_running_pod(pods: &Api<Pod>, label: &str) -> anyhow::Result<Pod> {
+    let lp = ListParams::default().labels(&format!("app={label}"));
+    let pod = pods
+        .list(&lp)
+        .await?
+        .iter()
+        .next()
+        .context(format!("Pod not found: {label}"))
```
pompon0, line 229: Use `with_context(||...)` in case context is not a constant (so that it is not evaluated if not needed)

node/tools/src/k8s.rs
```diff
@@ -12,15 +12,168 @@ use k8s_openapi::{
...
+    /// Wait for a deployed consensus node to be ready and have an IP address
+    pub async fn await_running_pod(
+        &mut self,
+        client: &Client,
+        namespace: &str,
+    ) -> anyhow::Result<Pod> {
+        let pods: Api<Pod> = Api::namespaced(client.clone(), namespace);
+        // Wait until the pod is running, otherwise we get 500 error.
+        let pod = Retry::spawn(FixedInterval::from_millis(1000).take(15), || {
+            get_running_pod(&pods, &self.id)
```
pompon0, line 51: that tooling looks like an overkill for a simple retry loop with constant delay.

node/tools/src/k8s.rs
```diff
@@ -167,145 +317,19 @@ pub async fn create_tests_deployment(client: &Client) -> anyhow::Result<()> {
...
+    let lp = ListParams::default().labels(&format!("app={label}"));
+    let pod = pods
+        .list(&lp)
+        .await?
+        .iter()
+        .next()
+        .with_context(|| format!("Pod not found: {label}"))
+        .cloned()?;
+    if is_pod_running(&pod) {
+        Ok(pod)
```
pompon0, line 330: if !is_pod_running(&pod) {
  anyhow::bail!("Pod is not running"));
}
Ok(pod)

node/tools/src/k8s.rs
```diff
@@ -167,145 +317,19 @@ pub async fn create_tests_deployment(client: &Client) -> anyhow::Result<()> {
...
-        Ok(p)
+async fn get_running_pod(pods: &Api<Pod>, label: &str) -> anyhow::Result<Pod> {
+    let lp = ListParams::default().labels(&format!("app={label}"));
+    let pod = pods
+        .list(&lp)
+        .await?
+        .iter()
+        .next()
+        .with_context(|| format!("Pod not found: {label}"))
+        .cloned()?;
```
pompon0, line 328: items.pop().with_context(...)

pompon0 (approved):

node/tools/src/k8s.rs
```diff
@@ -359,3 +354,23 @@ fn get_cli_args(peers: &[NodeAddr]) -> Vec<String> {
         .to_vec()
     }
 }
+
+async fn retry<T, Fut, F>(retries: usize, delay: usize, mut f: F) -> anyhow::Result<T>
```
pompon0, line 358: pass `std::time::Duration`, usize is not a unit of time.

node/tools/src/k8s.rs
```diff
@@ -359,3 +354,23 @@ fn get_cli_args(peers: &[NodeAddr]) -> Vec<String> {
...
+where
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = anyhow::Result<T>>,
+{
+    let delay = std::time::Duration::from_millis(delay.try_into()?);
```
pompon0, line 363: https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#impl-Interval

node/tools/src/k8s.rs
```diff
@@ -359,3 +354,23 @@ fn get_cli_args(peers: &[NodeAddr]) -> Vec<String> {
...
+    let mut count = 0;
+    loop {
```
pompon0, line 365: for count in 0.. {
   ...
}

node/tools/src/k8s.rs
```diff
@@ -359,3 +354,23 @@ fn get_cli_args(peers: &[NodeAddr]) -> Vec<String> {
...
+        let result = f().await;
+        if result.is_ok() {
+            return result;
+        }
```
pompon0, line 369: if result.is_ok() || count > retries { return result; }

node/tools/src/k8s.rs
ElFantasma: This would have been really clean
```
{
    let mut interval = time::interval(delay);
    for count in 0.. {
        interval.tick().await;
        let result = f().await;
        if result.is_ok() || count > retries {
            return result
        }
    }
}
```
but it seems rust compiler can't figure out that it won't have zero iterations:
```
mismatched types
  expected enum `std::result::Result<T, anyhow::Error>`
found unit type `()`
return a value for the case when the loop has zero elements to iterate on, otherwise consider changing the return type to account for that possibility
```

pompon0: just put Ok(()) at the end (or unreachable).

PR #70, opened 2024-02-27
Add `chaos-mesh` for testing (BFT-392)

IAvecilla:
## What ❔

Incorporate `chaos-mesh` tools into the consensus testing framework.

## Why ❔

To validate the network's behavior in response to non-canonical node behavior, we need tests. `Chaos-mesh` enables us to introduce delays or node crashes, simulating real-world network scenarios for thorough testing.


node/tests/src/main.rs
```diff
@@ -83,6 +83,15 @@ pub async fn sanity_test() {
     }
 }
 
+/// Sanity test for the RPC server.
+/// We use unwraps here because this function is intended to be used like a test.
+pub async fn delay_test() {
```
moshababo, line 88: This is not a delay test per se, but rather applying delay behavior on a running pod.

node/tools/src/k8s.rs
```diff
@@ -121,6 +128,38 @@ pub async fn create_or_reuse_namespace(client: &Client, name: &str) -> anyhow::R
     }
 }
 
+pub async fn add_chaos_delay_for_node(client: &Client, node_name: &str) -> anyhow::Result<()> {
```
moshababo, line 131: Nit: technically it's applied to a pod, not a node.

moshababo, line 131: The chaos presets should be defined in a separate module.

moshababo, line 131: Other chaos actions to be applied: pod failure & network partition.

moshababo, line 131: This needs to support teardown/cleanup.

node/tests/src/main.rs
moshababo, line 88: This needs to support the scheduling of the chaos action, meaning to apply & remove it at specific views.

IAvecilla:
@moshababo Thanks for the comments! I'll address and incorporate those suggestions.

Regarding the overall design, I'm encountering some challenges with defining the workflow for these tests. Currently, the workflow involves:

- Spinning up multiple consensus nodes.
- Running tests within a pod, which includes a sanity test along with basic delay and network failure tests.
- Chaos mesh tests require a `NetworkChaos` deployment, a component of the additional resources in `chaos-mesh`. After a set duration, we check the network state somehow.

The issue arises when we attempt to deploy the `NetworkChaos` component. Since the tester is already running inside a Kubernetes pod, deploying resources within the Kubernetes cluster from within a pod isn't a feasible solution. I'm exploring alternative approaches to address this challenge. Do you have any suggestions or ideas on how to achieve that?

moshababo:
It should be feasible to deploy `NetworkChaos` from within the cluster. Can you describe the problem you encounter?

Maybe it's because you're not using `kube::config::incluster()` to create the test's kube client. Only deploying the test module should run outside the cluster, locally on dev / CI machine. Everything else (initiated by the test module) happens within the cluster.

moshababo:
Can you respond to my previous reviews so I'll know what changes to expect?

node/tests/src/main.rs
IAvecilla: Right, it was initially a rough draft just to observe how the delay functions. It should now operate as a proper test since it verifies a couple of conditions after the delay.

IAvecilla: We can make request to the node for the its latest view and start the delay after a certain number. Right now the delay starts in the beginning of the test with a set duration of 10 seconds, is it necessary to check for a certain view?

node/tools/src/k8s.rs
IAvecilla: I updated the function signature to accept the duration of the delay as a parameter. Once the delay completes, it will not remove the chaos action from the cluster but it will stop. Perhaps we could introduce another function to delete that action, which we can then call at the end of the test. What do you think about this approach?

IAvecilla:
@moshababo Done!

node/tools/src/k8s.rs
moshababo: It will be better to further split it, but it can be done later.

moshababo: Yes, full cleanup need to be supported, to be called prior to the end of the test, in accordance with the test scenario, which might not rely on clock time. And in addition, for not introducing test code bugs due to unpredictable cluster state.

node/tests/src/main.rs
moshababo: Yes, most byzantine behavior duration will need to be "denominated" in views, not clock time, for actually testing liveness during the disruption (and not just quick recovery once it's over). 

It might be needed only for the removal, while applying it always at the beginning of the test, but it's better to just have the test framework supports chaos scheduling with  `from_view`,`to_view` args.

node/tests/src/main.rs
```diff
@@ -15,91 +31,149 @@ struct TesterCLI {
...
+        .unwrap();
+    let url: String = format!("http://{}", ip);
+    let rpc_client = HttpClientBuilder::default().build(url).unwrap();
+    let response: serde_json::Value = rpc_client
+        .request(last_commited_block::method(), rpc_params!())
+        .await
+        .unwrap();
+    let last_voted_view: u64 =
+        serde_json::from_value(response.get("last_commited_block").unwrap().to_owned()).unwrap();
+    for _ in 0..5 {
```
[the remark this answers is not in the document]
IAvecilla: At first my idea was to check a few times that the node was indeed going through the delay and not being able to vote for the views, now I'm not sure if that is useful at all. Maybe there's a better way to do achieve this.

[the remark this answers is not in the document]
moshababo: I don't think it adds much value, while adding non-determinism to the scenario (for being dependent on the response latency). 
It is good to check that status while the chaos resources are deployed, but it's better done via test controlling the teardown explicitly.

brunoffranca:
@moshababo given the developments with AttackNet, should we close this?

PR #71, opened 2024-02-29
Full config in cli (BFT-427)

ElFantasma:
## What ❔

Consensus nodes can receive the full configuration via cli args, as well as node and validator keys
That, in turn is used to start the nodes in k8s pods without requiring any configuration in local filesystem

## Why ❔

This is cleaner and easier. Now the docker image built is independent of the number of nodes to deploy, as every configuration parameter is provided via command line


node/tools/src/config.rs
```diff
@@ -152,13 +152,19 @@ impl ProtoFmt for AppConfig {
...
 #[derive(Debug)]
-pub struct ConfigPaths<'a> {
+pub struct ConfigArgs<'a> {
+    /// Node configuration from command line.
+    pub config: Option<AppConfig>,
     /// Path to a JSON file with node configuration.
-    pub app: &'a Path,
+    pub config_file: &'a Path,
+    /// Validator key as a string.
+    pub validator_key: Option<String>,
```
pompon0, line 161: why not `validator::SecretKey`?

node/tools/src/config.rs
```diff
@@ -152,13 +152,19 @@ impl ProtoFmt for AppConfig {
...
     /// Path to a validator key file.
-    pub validator_key: Option<&'a Path>,
+    pub validator_key_file: &'a Path,
```
pompon0, line 163: this field is now mutually exclusive with validator_key, no? Make it an enum.

node/tools/src/config.rs
```diff
@@ -170,34 +176,53 @@ pub struct Configs {
     pub database: PathBuf,
 }
 
-impl<'a> ConfigPaths<'a> {
+impl<'a> ConfigArgs<'a> {
     // Loads configs from the file system.
-    pub fn load(self) -> anyhow::Result<Configs> {
+    pub fn load(self) -> anyhow::Result<Configs, anyhow::Error> {
```
pompon0, line 181: what happened here? :)

node/tools/src/config.rs
pompon0: or actually you might be better off defining a separate ConfigArgs IN ADDITION to ConfigPath. They don't seem to have much overlap.

node/tools/src/main.rs
```diff
@@ -45,22 +44,74 @@ struct Args {
     add_gossip_static_outbound: Option<NodeAddrs>,
 }
 
-impl Args {
+/// Node key definitions:
+/// If both are present, node-key option has precedence.
+#[derive(Debug, Args)]
+#[group(required = false, multiple = true)]
+struct NodeKey {
+    /// Set Node key in command line
```
pompon0, line 52: secret key

node/tools/src/main.rs
```diff
@@ -45,22 +44,74 @@ struct Args {
...
+
+    /// Set a path to the node key file
+    #[arg(long, default_value = "./node_key")]
+    node_key_file: PathBuf,
+}
+
+/// Validator key definitions:
+/// If validator key is not set, and key file is not present,
+/// a node will be initialized as a non-validator node.
+/// If both are present, validator-key option has precedence.
```
pompon0, line 64: secret key

node/tools/src/main.rs
```diff
@@ -45,22 +44,74 @@ struct Args {
...
+#[derive(Debug, Args)]
+#[group(required = false, multiple = true)]
```
pompon0, line 66: clap supports annotating mutually exclusive flags and enforces exclusivity

node/tools/src/config.rs
```diff
@@ -152,13 +152,19 @@ impl ProtoFmt for AppConfig {
 
 /// This struct holds the file path to each of the config files.
```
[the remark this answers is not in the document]
ElFantasma: Done

node/tools/src/main.rs
ElFantasma: Done

node/tools/src/main.rs
ElFantasma: Done

node/tools/src/config.rs
ElFantasma: Not sure I understand your comments here. I made them mutually exclusive [here](https://github.com/matter-labs/era-consensus/blob/b63f987057fce56731cc7ebd6098a651fbb88930/node/tools/src/main.rs#L54)

In any case I intend to keep previous behaviour: If no `--validator_key`, nor `--validator_key_file` are defined, still assume `./validator_key` as default if that file is present in root dir. 

If not, then run in non-validator mode

moshababo: If they are mutually exclusive on the CLI arg level, you are currently losing that restriction here and that what makes `ConfigArgs::load` more complicated.

node/tools/src/main.rs
```diff
@@ -1,66 +1,109 @@
...
-    /// IP address and key of the seed peers.
-    #[arg(long)]
-    add_gossip_static_outbound: Option<NodeAddrs>,
 }
 
-impl Args {
+/// Node key definitions:
+/// If both are present, node-key option has precedence.
+#[derive(Debug, Args)]
+#[group(required = false, multiple = false)]
```
moshababo, line 38: The best would be if these input options won't be mutually exclusive, but follow the pattern mentioned here: https://github.com/matter-labs/era-consensus/pull/62#discussion_r1493585570 (so that every field can be overridden independently).

You can see a similar example from a different project here: [config.go#L716](https://github.com/lightningnetwork/lnd/blob/fba6fdaf94be92063f19e8d6e92c7722278e156f/config.go#L716).

If it's too complicated we can leave it as is for now.

node/tools/src/config.rs
ElFantasma: I'm not following here. What restriction I'm losing?

To be fair, this functionality had a bug previously: if no `--validator-key` was defined, nor `./validator_key` file was present, the executable raised an error instead of running in non-validator mode.

`ConfigArgs::load` got a bit more complicated, that's true, but I'm not sure how to improve it.

node/tools/src/main.rs
moshababo: Ok, I think it's good enough, because the the node running stand-alone isn't its main use case.

moshababo: Ok, I think it's good enough, because the the node running stand-alone isn't the main use case.

node/tools/src/config.rs
moshababo: > I'm not following here. What restriction I'm losing?

The restriction of mutual exclusivity, and so the ability to represent it as as enum, instead of 2 separate fields which can both exist. 
However, because default values are used as well, it might be less applicable.

pompon0: I mean, can't you just make "specifying config via command line arg" and "specifying config via file(s)" mutually exclusive as a whole? We don't have a use case for combining those, or do we?

node/tools/src/k8s.rs
```diff
@@ -361,41 +317,16 @@ pub async fn deploy_node(
...
+async fn get_running_pod(pods: &Api<Pod>, label: &str) -> anyhow::Result<Pod> {
+    let lp = ListParams::default().labels(&format!("app={label}"));
+    let pod = pods
+        .list(&lp)
+        .await?
+        .items
+        .pop()
+        .with_context(|| format!("Pod not found: {label}"))?;
+    if !is_pod_running(&pod) {
+        anyhow::bail!("Pod is not running");
```
pompon0, line 329: `anyhow::ensure!(is_pod_running(&pod), "Pod is not running");`

node/tools/src/config.rs
moshababo: Combining them is the normal use-case for actual users, where static config files are the "default", and the nested CLI args are the dynamic override. 
However, this is not mandatory, especially when the node standalone use-case is test-only.

pompon0 (approved):

node/tools/src/config.rs
pompon0: I was pretty sure that we have the configuration using CLI args only because of kubernetes/docker limitations, i.e. that it is impossible to run a pre-built docker image with access to a dynamically constructed file.

ElFantasma: Now I made this set of options mutually exclusive: 
You'll need full cli configuration or full file configuration (if no configuration args are provided at all, default files are expected)
Also, if no `--validator-key` or `--validator-key-file` is provided, nor a `validator-key` file exists in root, it won't complain but it should run in non-validator mode

PR #75, opened 2024-03-13
support for running a node from a state snapshot (BFT-418)

pompon0:
Support for a situation in which a node is started with a blockchain state at block X > 0. Such a node doesn't want to fetch and store blocks <X. I've adjusted syncing logic and validator logic (since validator also might start from state snapshot).
Added tests for the new cases.

I've removed `parent` from the block struct, because the cost of maintaining it exceeds the gains at this point (in zksync-era payload implementation we have a commitment to the chain history anyway).


node/actors/bft/src/replica/leader_prepare.rs
```diff
@@ -110,6 +111,14 @@ impl StateMachine {
             });
         }
 
+        // Replica MUSN'T vote for blocks which have been already pruned for storage.
+        // (because it won't be albe to persist and broadcast them once finalized).
```
brunoffranca, line 115:
```suggestion
        // Replica MUSTN'T vote for blocks which have been already pruned for storage.
        // (because it won't be able to persist and broadcast them once finalized).
```

brunoffranca (approved):
Looks good. Removing parent hash does simplify quite a bit.

PR #76, opened 2024-03-14
Broadcast validators signature and collect QC [BFT-414]

IAvecilla:
## What ❔

Implement a mechanism for validators to broadcast their signatures to other nodes in the network and collect them in a new certificate.

## Why ❔

This is essential for signing L1 batches and sending them to L1 for verification. Validators need to broadcast their signatures and gather them in a certificate in case the majority signs the new batch.

### DIFF
--- a/node/libs/roles/src/proto/validator.proto
+++ b/node/libs/roles/src/proto/validator.proto
@@ -144,11 +144,16 @@ message NetAddress {
   optional std.Timestamp timestamp = 3; // required
 }
 
+message L1BatchSignature {
+  optional bytes signature = 1; // required
+}
+
 message Msg {
   oneof t { // required
     ConsensusMsg consensus = 1;
     bytes session_id = 2;
     NetAddress net_address = 3;
+    L1BatchSignature l1_batch_signature = 4;
   }
 }
 
--- a/node/libs/roles/src/validator/messages/msg.rs
+++ b/node/libs/roles/src/validator/messages/msg.rs
@@ -1,6 +1,9 @@
 //! Generic message types.
-use super::{ConsensusMsg, NetAddress};
-use crate::{node::SessionId, validator, validator::Error};
+use super::{ConsensusMsg, L1BatchSignature, NetAddress};
+use crate::{
+    node::SessionId,
+    validator::{self, Error},
+};
 use std::fmt;
 use zksync_consensus_crypto::{keccak256, ByteFmt, Text, TextFmt};
 use zksync_consensus_utils::enum_util::{BadVariantError, Variant};
@@ -14,6 +17,8 @@ pub enum Msg {
     SessionId(SessionId),
     /// validator discovery
     NetAddress(NetAddress),
+    /// l1 batch signature
+    L1BatchSignature(L1BatchSignature),
 }
 
 impl Msg {
@@ -53,6 +58,18 @@ impl Variant<Msg> for NetAddress {
     }
     fn extract(msg: Msg) -> Result<Self, BadVariantError> {
         let Msg::NetAddress(this) = msg else {
+            return Err(BadVariantError);
+        };
+        Ok(this)
+    }
+}
+
+impl Variant<Msg> for L1BatchSignature {
+    fn insert(self) -> Msg {
+        Msg::L1BatchSignature(self)
+    }
+    fn extract(msg: Msg) -> Result<Self, BadVariantError> {
+        let Msg::L1BatchSignature(this) = msg else {
             return Err(BadVariantError);
         };
         Ok(this)
--- a/node/libs/roles/src/validator/conv.rs
+++ b/node/libs/roles/src/validator/conv.rs
@@ -1,8 +1,9 @@
 use super::{
     AggregateSignature, BlockHeader, BlockHeaderHash, BlockNumber, CommitQC, ConsensusMsg,
-    FinalBlock, Fork, ForkNumber, Genesis, GenesisHash, LeaderCommit, LeaderPrepare, Msg, MsgHash,
-    NetAddress, Payload, PayloadHash, Phase, PrepareQC, ProtocolVersion, PublicKey, ReplicaCommit,
-    ReplicaPrepare, Signature, Signed, Signers, ValidatorSet, View, ViewNumber,
+    FinalBlock, Fork, ForkNumber, Genesis, GenesisHash, L1BatchSignature, LeaderCommit,
+    LeaderPrepare, Msg, MsgHash, NetAddress, Payload, PayloadHash, Phase, PrepareQC,
+    ProtocolVersion, PublicKey, ReplicaCommit, ReplicaPrepare, Signature, Signed, Signers,
+    ValidatorSet, View, ViewNumber,
 };
 use crate::{node::SessionId, proto::validator as proto};
 use anyhow::Context as _;
@@ -355,6 +356,20 @@ impl ProtoFmt for NetAddress {
     }
 }
 
+impl ProtoFmt for L1BatchSignature {
+    type Proto = proto::L1BatchSignature;
+
+    fn read(r: &Self::Proto) -> anyhow::Result<Self> {
+        Ok(Self(ByteFmt::decode(required(&r.signature)?)?))
+    }
+
+    fn build(&self) -> Self::Proto {
+        Self::Proto {
+            signature: Some(self.0.encode()),
+        }
+    }
+}
+
 impl ProtoFmt for Msg {
     type Proto = proto::Msg;
 
@@ -364,6 +379,9 @@ impl ProtoFmt for Msg {
             T::Consensus(r) => Self::Consensus(ProtoFmt::read(r).context("Consensus")?),
             T::SessionId(r) => Self::SessionId(SessionId(r.clone())),
             T::NetAddress(r) => Self::NetAddress(ProtoFmt::read(r).context("NetAddress")?),
+            T::L1BatchSignature(r) => {
+                Self::L1BatchSignature(ProtoFmt::read(r).context("L1BatchSignature")?)
+            }
         })
     }
 
@@ -374,6 +392,7 @@ impl ProtoFmt for Msg {
             Self::Consensus(x) => T::Consensus(x.build()),
             Self::SessionId(x) => T::SessionId(x.0.clone()),
             Self::NetAddress(x) => T::NetAddress(x.build()),
+            Self::L1BatchSignature(x) => T::L1BatchSignature(x.build()),
         };
 
         Self::Proto { t: Some(t) }
--- a/node/libs/roles/src/validator/messages/l1_batch_signature.rs
+++ b/node/libs/roles/src/validator/messages/l1_batch_signature.rs
@@ -1,0 +1,6 @@
+use crate::validator::Signature;
+
+/// A message to send by validators to the gossip network.
+/// It contains the validators signature to sign the block batches to be sent to L1.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct L1BatchSignature(pub Signature);
--- a/node/libs/roles/src/validator/messages/mod.rs
+++ b/node/libs/roles/src/validator/messages/mod.rs
@@ -3,6 +3,7 @@
 mod block;
 mod consensus;
 mod discovery;
+mod l1_batch_signature;
 mod leader_commit;
 mod leader_prepare;
 mod msg;
@@ -12,6 +13,7 @@ mod replica_prepare;
 pub use block::*;
 pub use consensus::*;
 pub use discovery::*;
+pub use l1_batch_signature::*;
 pub use leader_commit::*;
 pub use leader_prepare::*;
 pub use msg::*;


node/actors/network/src/consensus/mod.rs
```diff
@@ -59,13 +67,27 @@ impl rpc::Handler<rpc::consensus::Rpc> for &Network {
     }
 }
 
+#[async_trait::async_trait]
+impl rpc::Handler<rpc::signature::Rpc> for &L1BatchSignatureServer<'_> {
+    /// Here we bound the buffering of incoming consensus messages.
+    fn max_req_size(&self) -> usize {
+        self.0.gossip.cfg.max_block_size.saturating_add(kB)
```
pompon0, line 74: Block size is unrelated to the signature size. Please provide relevant bound.

node/actors/network/src/consensus/mod.rs
```diff
@@ -59,13 +67,27 @@ impl rpc::Handler<rpc::consensus::Rpc> for &Network {
...
+    }
+
+    async fn handle(&self, _ctx: &ctx::Ctx, req: rpc::signature::Req) -> anyhow::Result<()> {
+        self.0.l1_batch_qc.clone().add(req.0.sig);
```
pompon0, line 78: the signature should be verified before inserted anywhere.

node/libs/roles/src/validator/messages/msg.rs
```diff
@@ -14,6 +17,8 @@ pub enum Msg {
     SessionId(SessionId),
     /// validator discovery
     NetAddress(NetAddress),
+    /// l1 batch signature
+    L1BatchSignature(L1BatchSignatureMsg),
```
pompon0, line 21: Something went wrong here. It is the Msg what is being signed, so you want Msg::L1Batch variant not L1BatchSignature, otherwise validator will be signing its own signature which doesn't make sense in this case.

brunoffranca:
@IAvecilla I think it will be better if we create another validator key just for batch signing. So, this way we can use different signatures schemes for block signing and for batch signing.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -103,7 +103,11 @@ impl StateMachine {
             .expect("Could not add message to CommitQC");
 
         // Calculate the CommitQC signers weight.
-        let weight = self.config.genesis().committee.weight(&commit_qc.signers);
+        let weight = self
+            .config
+            .genesis()
+            .validators_committee
```
pompon0, line 109: let's rename it back to "validators" then.

node/actors/executor/src/lib.rs
```diff
@@ -36,6 +42,14 @@ impl fmt::Debug for Validator {
     }
 }
 
+impl fmt::Debug for Attester {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.debug_struct("AttesterExecutor")
+            .field("key", &self.key)
```
pompon0, line 48: this can be just derived. Also "ValidatorExecutor" should be just "Validator" - we forgot to change that.

node/actors/network/src/config.rs
```diff
@@ -42,6 +44,10 @@ impl Default for RpcConfig {
                 burst: 10,
                 refresh: time::Duration::ZERO,
             },
+            push_l1_batch_signature_rate: limiter::Rate {
+                burst: 10,
```
pompon0, line 48: this should be low. Let's start with burst = 2, and refresh = 500ms. In tests we can use a higher/infinite rate limit of course.

node/actors/network/src/consensus/mod.rs
```diff
@@ -9,7 +14,7 @@ use std::{
 };
 use tracing::Instrument as _;
 use zksync_concurrency::{ctx, oneshot, scope, sync, time};
-use zksync_consensus_roles::validator;
+use zksync_consensus_roles::validator::{self};
```
pompon0, line 17: `::{self}` is redundant

node/libs/roles/src/proto/attester.proto
```diff
@@ -0,0 +1,49 @@
...
+  optional AggregateSignature sig = 3; // required
+}
+
+message Msg {
+  oneof t { // required
+    L1Batch l1_batch = 4;
+  }
+}
+
+message SignedBatch {
```
pompon0, line 24: SignedMsg

node/libs/roles/src/proto/validator.proto
```diff
@@ -3,19 +3,21 @@ syntax = "proto3";
...
   optional uint64 fork_number = 6; // required; ForkNumber
   optional uint64 first_block = 7; // required; BlockNumber
+  repeated attester.WeightedAttester attesters = 4;
  
   // These properties are expected to be overwritten each epoch.
   // We will either remove them entirely, or keep them for the initial epoch.
   optional uint32 protocol_version = 8; // required; ProtocolVersion
   repeated WeightedValidator validators_v1 = 3;
-  optional LeaderSelectionMode leader_selection = 4; // required
+  optional LeaderSelectionMode leader_selection = 9; // required
```
pompon0, line 20: the whole point of numeric tags is that they are not allowed to change to maintain backward/forward compatibility. Just use the next unused number for the new field.

node/libs/roles/src/proto/validator.proto
```diff
@@ -3,19 +3,21 @@ syntax = "proto3";
...
 import "zksync/std.proto";
+import "zksync/roles/attester.proto";
 
 message Genesis {
   reserved 1,2;
   reserved "fork","validators";
   optional uint64 chain_id = 5; // required
   optional uint64 fork_number = 6; // required; ForkNumber
   optional uint64 first_block = 7; // required; BlockNumber
+  repeated attester.WeightedAttester attesters = 4;
```
pompon0, line 14: attesters are not supposed to be fixed per genesis, but rather might change each epoch, so move this field to the section below

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -241,7 +240,9 @@ pub struct GenesisRaw {
     /// First block of a fork.
     pub first_block: BlockNumber,
     /// Set of validators of the chain.
-    pub committee: Committee,
+    pub validators_committee: Committee,
+    /// Set of attesters of the chain.
+    pub attesters_committee: attester::Committee,
```
pompon0, line 245: just "validators" and "attesters"

node/libs/roles/src/proto/attester.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package zksync.roles.attester;
+
+import "zksync/std.proto";
+
+message L1Batch {
+  optional uint64 number = 1; // required
+  optional std.Timestamp timestamp = 2; // required
```
pompon0, line 9: add a todo to add the hash field

node/libs/roles/src/validator/messages/msg.rs
```diff
@@ -1,6 +1,9 @@
 //! Generic message types.
 use super::{ConsensusMsg, NetAddress};
-use crate::{node::SessionId, validator};
+use crate::{
+    node::SessionId,
+    validator::{self},
```
pompon0, line 5: `::{self}` is redundant

node/tools/src/config.rs
```diff
@@ -210,6 +214,7 @@ impl Configs {
                         self.app.max_payload_size,
                     )),
                 }),
+            attester: None,
```
pompon0, line 217: if it is supposed to be unused for now, then don't add the attester_secret_key to the config at all.

node/actors/network/src/gossip/testonly.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(dead_code)]
 use super::*;
 use crate::{frame, mux, noise, preface, rpc, Config, GossipConfig};
-use anyhow::Context as _;
+use anyhow::Context;
```
pompon0, line 4: ?

node/actors/network/src/gossip/tests/mod.rs
```diff
@@ -1,6 +1,6 @@
 use super::*;
 use crate::{metrics, preface, rpc, testonly};
-use anyhow::Context as _;
```
pompon0, line 3: ditto

node/actors/network/src/gossip/tests/mod.rs
```diff
@@ -14,7 +14,10 @@ use zksync_concurrency::{
     testonly::{abort_on_panic, set_timeout},
     time,
 };
-use zksync_consensus_roles::validator;
+use zksync_consensus_roles::{
+    attester::{L1Batch, SignedBatchMsg},
```
pompon0, line 18: nit: please prefer to use qualified names.

node/actors/network/src/gossip/tests/mod.rs
```diff
@@ -99,6 +105,14 @@ fn mk_netaddr(
     })
 }
 
+fn mk_batch(
+    key: &attester::SecretKey,
+    number: BatchNumber,
+    timestamp: time::Utc,
+) -> SignedBatchMsg<L1Batch> {
+    key.sign_batch_msg(L1Batch { number, timestamp })
```
pompon0, line 113: drop "batch" infix from names

node/actors/network/src/gossip/tests/mod.rs
```diff
@@ -111,6 +125,17 @@ fn random_netaddr<R: Rng>(
     ))
 }
 
+fn random_signature<R: Rng>(
```
pompon0, line 128: nit: and here in turn, a "batch" infix would be meaningful :P

node/actors/network/src/proto/gossip.proto
```diff
@@ -17,6 +18,11 @@ message PushValidatorAddrs {
   repeated roles.validator.Signed net_addresses = 1;
 }
 
+message PushSignature {
```
pompon0, line 21: This name also would benefit from L1Batch infix.

node/actors/network/src/proto/signature.proto
pompon0, line 1: is this file needed?

node/actors/network/src/rpc/push_signature.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::{mux, proto::gossip as proto};
+use anyhow::Context;
+use zksync_consensus_roles::attester::{self, L1Batch};
+use zksync_protobuf::ProtoFmt;
+
+/// Signature RPC.
+pub(crate) struct Rpc;
+
+impl super::Rpc for Rpc {
+    const CAPABILITY_ID: mux::CapabilityId = 5;
```
pompon0, line 13: add it to test_capability_rpc_correspondence test - it currently acts as a register for all RPCs to prevent reuse of CAPABILITY_IDs

node/actors/network/src/rpc/mod.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) mod get_block;
 mod metrics;
 pub(crate) mod ping;
 pub(crate) mod push_block_store_state;
+pub(crate) mod push_signature;
```
pompon0, line 29: push_l1batch_signature

node/actors/network/src/testonly.rs
```diff
@@ -77,27 +77,32 @@ pub fn new_configs(
...
-            max_block_size: usize::MAX,
-            tcp_accept_rate: limiter::Rate::INF,
-            rpc: RpcConfig::default(),
-            max_block_queue_size: 10,
-        }
-    });
+    let configs = setup
+        .validator_keys
+        .iter()
+        .zip(setup.attester_keys.iter())
```
pompon0, line 83: do not assume that setup contains the same number of validators and attesters

node/deny.toml
```diff
@@ -76,7 +76,7 @@ skip = [
 
     # Old versions required by kube.
     { name = "strsim", version = "0.10.0" },
-
+
```
pompon0, line 79: ?

node/libs/roles/src/attester/keys/signature.rs
```diff
@@ -0,0 +1,56 @@
...
+impl fmt::Debug for Signature {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.write_str(&TextFmt::encode(self))
+    }
+}
+
+impl std::hash::Hash for Signature {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        ByteFmt::encode(self).hash(state)
+    }
```
pompon0, line 55: nit: make bn254::Signature implement Hash and just derive Hash for attester::Signature (and same for validator::Signature and node::Signature while you are at it).

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::{
+    attester::{self, AggregateSignature},
+    validator::Genesis,
+};
+
+use super::{SignedBatchMsg, Signers};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default, PartialOrd)]
+/// A batch number.
+pub struct BatchNumber(pub u64);
```
pompon0, line 12: use "L1Batch" and "l1batch" names consistently (i.e. do not abbreviate to "batch").

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl BatchNumber {
+    /// Increment the batch number.
+    pub fn next_batch_number(&self) -> BatchNumber {
```
pompon0, line 16: just "next()" - "batch_number" is redundant.

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+            message,
+            signers: Signers::new(genesis.attesters_committee.len()),
+            signature: attester::AggregateSignature::default(),
+        }
+    }
+
+    /// Add a attester's signature.
+    /// Signature is assumed to be already verified.
+    pub fn add(&mut self, msg: &SignedBatchMsg<L1Batch>, genesis: &Genesis) {
+        if self.message != msg.msg {
```
pompon0, line 82: nit: ptal at CommitQC - `add()` returns an error now, rather than silently ignoring useless signatures.

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.signature.add(&msg.sig);
+    }
+
+    /// Verifies the signature of the L1BatchQC.
+    pub fn verify(&self, genesis: &Genesis) -> Result<(), L1BatchQCVerifyError> {
+        use L1BatchQCVerifyError as Error;
+        if self.signers.len() != genesis.attesters_committee.len() {
+            return Err(Error::BadSignersSet);
+        }
+        // Verify the signers' weight is enough.
```
pompon0, line 101: verify that ...

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Get the index of a attester in the committee.
+    pub fn index(&self, attester: &PublicKey) -> Option<usize> {
+        self.indexes.get(attester).copied()
+    }
+
+    /// Computes the leader for the given view.
+    pub fn view_leader(&self, view_number: ViewNumber) -> PublicKey {
+        let index = view_number.0 as usize % self.len();
+        self.get(index).unwrap().key.clone()
+    }
```
pompon0, line 154: there is no such a thing as leader in an attester committee

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,279 @@
...
+
+/// Calculate the consensus threshold, the minimum votes' weight for any consensus action to be valid,
+/// for a given committee total weight.
+pub fn threshold(total_weight: u64) -> u64 {
+    total_weight - max_faulty_weight(total_weight)
+}
+
+/// Calculate the maximum allowed weight for faulty replicas, for a given total weight.
+pub fn max_faulty_weight(total_weight: u64) -> u64 {
+    // Calculate the allowed maximum weight of faulty replicas. We want the following relationship to hold:
```
pompon0, line 191: add a comment that technically we need just n > 2*f, but for now we use a threshold consistent with the validator committee.

node/libs/roles/src/attester/tests.rs
```diff
@@ -0,0 +1,234 @@
...
+use zksync_consensus_crypto::{ByteFmt, Text, TextFmt};
+use zksync_protobuf::testonly::test_encode_random;
+
+#[test]
+fn test_byte_encoding() {
+    let ctx = ctx::test_root(&ctx::RealClock);
+    let rng = &mut ctx.rng();
+
+    let sk: SecretKey = rng.gen();
+    assert_eq!(
```
pompon0, line 16: SecretKey for all roles now implement PartialEq, so you can just directly compare sk with decode(encode(sk)). Please also update tests for other roles.

node/libs/roles/src/attester/tests.rs
```diff
@@ -0,0 +1,234 @@
...
+#[test]
+fn test_text_encoding() {
+    let ctx = ctx::test_root(&ctx::RealClock);
+    let rng = &mut ctx.rng();
+
+    let sk: SecretKey = rng.gen();
+    let t = TextFmt::encode(&sk);
+    assert_eq!(
+        sk.public(),
+        Text::new(&t).decode::<SecretKey>().unwrap().public()
```
pompon0, line 45: ditto

node/libs/roles/src/attester/tests.rs
```diff
@@ -0,0 +1,234 @@
...
+        .verify_hash([(msg1, &key2.public()), (msg2, &key1.public())].into_iter())
+        .is_err());
+}
+
+fn make_l1_batch_msg(rng: &mut impl Rng) -> L1Batch {
+    L1Batch {
+        number: BatchNumber(rng.gen()),
+        timestamp: time::UNIX_EPOCH + time::Duration::seconds(rng.gen_range(0..1000000000)),
+    }
+}
```
pompon0, line 142: isn't it the same as rng.gen::<L1Batch>() ?

node/libs/roles/src/attester/tests.rs
```diff
@@ -0,0 +1,234 @@
...
+            .unwrap();
+    let genesis3 = genesis3.with_hash();
+    let attester_weight = setup1.genesis.attesters_committee.total_weight() / 6;
+
+    for i in 0..setup1.attester_keys.len() + 1 {
+        let mut qc = L1BatchQC::new(make_l1_batch_msg(rng), &setup1.genesis);
+        for key in &setup1.attester_keys[0..i] {
+            qc.add(&key.sign_batch_msg(qc.message.clone()), &setup1.genesis);
+        }
+        let expected_weight = i as u64 * attester_weight;
```
pompon0, line 164: I asked Esteban to fix this, but I think he forgot: don't unnecessarily assume that all attesters have the same weight here - just sum the weights of first i attesters in the committee.

node/actors/network/src/gossip/runner.rs
```diff
@@ -26,7 +29,25 @@ impl rpc::Handler<rpc::push_validator_addrs::Rpc> for PushValidatorAddrsServer<'
...
+            .await?;
+        Ok(())
+    }
+}
+
+struct L1BatchServer<'a>(&'a Network);
+
+#[async_trait::async_trait]
+impl rpc::Handler<rpc::push_signature::Rpc> for L1BatchServer<'_> {
+    /// Here we bound the buffering of incoming consensus messages.
```
pompon0, line 42: these are not consensus messages

node/actors/network/src/gossip/runner.rs
```diff
@@ -26,7 +29,25 @@ impl rpc::Handler<rpc::push_validator_addrs::Rpc> for PushValidatorAddrsServer<'
...
+            .update(&self.0.genesis().validators_committee, &req.0)
+            .await?;
+        Ok(())
+    }
+}
+
+struct L1BatchServer<'a>(&'a Network);
+
+#[async_trait::async_trait]
+impl rpc::Handler<rpc::push_signature::Rpc> for L1BatchServer<'_> {
```
pompon0, line 41: use the full name of the rpc here. Like `PushBatchSignaturesServer`.

node/actors/network/src/rpc/push_signature.rs
pompon0, line 1: push_batch_signatures?

pompon0: now I'm thinking that perhaps a better name would be "push_batch_votes"? Like: quorum consists of votes, so that would be a more meaningful naming IMO.

node/actors/network/src/gossip/mod.rs
```diff
@@ -13,12 +13,23 @@
 //! eclipse attack. Dynamic connections are supposed to improve the properties of the gossip
 //! network graph (minimize its diameter, increase connectedness).
 use crate::{gossip::ValidatorAddrsWatch, io, pool::PoolWatch, Config};
-use std::sync::{atomic::AtomicUsize, Arc};
+use anyhow::Context;
```
pompon0, line 16: `as _` since we just use the interface methods.

node/actors/network/src/gossip/mod.rs
```diff
@@ -38,12 +49,22 @@ pub(crate) struct Network {
...
+    /// Current state of knowledge about batch signatures.
+    pub(crate) batch_signatures: L1BatchSignaturesWatch,
     /// Block store to serve `get_block` requests from.
     pub(crate) block_store: Arc<BlockStore>,
     /// Output pipe of the network actor.
     pub(crate) sender: channel::UnboundedSender<io::OutputMessage>,
     /// Queue of block fetching requests.
     pub(crate) fetch_queue: fetch::Queue,
+    /// L1 batches.
+    // pub(crate) l1_batches: sync::watch::Receiver<Option<L1Batch>>,
```
pompon0, line 61: remove?

node/actors/network/src/gossip/mod.rs
```diff
@@ -63,6 +84,11 @@ impl Network {
             ),
             outbound: PoolWatch::new(cfg.gossip.static_outbound.keys().cloned().collect(), 0),
             validator_addrs: ValidatorAddrsWatch::default(),
+            batch_signatures: L1BatchSignaturesWatch::default(),
+            attester_key: cfg.attester_key.clone(),
+            l1_batch_qc: HashMap::new(),
+            // l1_batches: sync::watch::channel(None).1,
```
pompon0, line 90: remove?

node/actors/network/src/gossip/mod.rs
```diff
@@ -103,4 +129,51 @@ impl Network {
...
+            let mut sub = self.batch_signatures.subscribe();
+            let signatures = sync::changed(ctx, &mut sub)
+                .await
+                .context("batch signatures")?
+                .clone();
+            for (_, sig) in signatures.0 {
+                self.l1_batch_qc
+                    .clone()
+                    .entry(sig.msg.number.clone())
+                    .or_insert_with(|| L1BatchQC::new(sig.msg.clone(), self.genesis()))
```
pompon0, line 148: attester might be malicious and sign invalid batches. We shouldn't construct the batch using the msg from the first signature we encounter, but rather collect the signatures for a expected batch that the node KNOWS of (node knows the expected batch because it has all the blocks of the batch).

node/actors/network/src/gossip/batch_signatures.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::watch::Watch;
+use std::{collections::HashSet, sync::Arc};
+use zksync_concurrency::sync;
+use zksync_consensus_roles::attester::{self, L1Batch};
+
+/// Mapping from attester::PublicKey to a signed attester::L1Batch message.
+/// Represents the currents state of node's knowledge about the attester signatures.
+#[derive(Clone, Default, PartialEq, Eq)]
+pub(crate) struct L1BatchSignatures(
+    pub(super) im::HashMap<attester::PublicKey, Arc<attester::SignedBatchMsg<L1Batch>>>,
```
pompon0, line 11: We need to be a bit more subtle with batch signatures retention than with network addresses: node should keep collecting and broadcasting signatures for batch N until it has a certificate for batch N (then it can just broadcast the whole certificate until the certificate is sent to L1). It shouldn't discard signatures for batch N just because an attester has sent a signature for batch N+1. IMO before we implement this broadcasting logic, we should design the API through which the node will receive the the batches to sign (from the business logic) and send back the commitqcs collected, so that we know exactly what properties the batch broadcasting should have.

pompon0 (changes requested):
Mostly naming suggestions.

node/actors/bft/src/leader/replica_commit.rs
IAvecilla: I deleted the prefix from the attesters too.

node/libs/roles/src/attester/messages/l1_batch.rs
brunoffranca: +1

node/libs/roles/src/attester/messages/msg.rs
brunoffranca: +1

node/libs/roles/src/attester/messages/msg.rs
brunoffranca: Technically, we just need f+1.

node/actors/network/src/rpc/push_signature.rs
brunoffranca: +1

node/actors/network/src/gossip/testonly.rs
IAvecilla: Sorry, that wasn't intentional. It's back to its original state

node/actors/network/src/gossip/tests/mod.rs
IAvecilla: I think that `context` is not being used in this module.

node/libs/roles/src/attester/messages/msg.rs
brunoffranca: I would add said comment to the threshold() method above.

node/actors/network/src/gossip/tests/mod.rs
IAvecilla: I think that `context` is no used here, should we keep it?

node/libs/roles/src/attester/messages/msg.rs
brunoffranca: I would add said comment to the `threshold` method above.

node/actors/network/src/gossip/batch_signatures.rs
brunoffranca: I propose [this interface](https://www.notion.so/matterlabs/L1-batches-and-Consensus-7348aef31a514940b7779e2293ba0da0?pvs=4#52f1bd16fe2f4588b88bc36ea16d0db4). @pompon0 @IAvecilla PTAL

node/libs/roles/src/proto/attester.proto
IAvecilla: Should we modify the `SignedBatchMsg` struct as well? I named it that way to make it clear that the only message the attesters could sign would be an `L1Batch`, but perhaps we can use `Signed` just like the validators and document it accordingly.

node/actors/network/src/proto/signature.proto
IAvecilla: Seems like it was created for an old implementation, deleting it!

node/actors/network/src/testonly.rs
IAvecilla: Since the attesters are not yet used in the config, this change is stalled. However, I'll keep it in mind when implementing it.

node/libs/roles/src/attester/messages/l1_batch.rs
IAvecilla: Yeah, I totally agree. It's weird because `next()` was my first choice, and I remember Clippy saying that it could clash with an `iter()` implementation for `BatchNumber`, but it's not raising any issues anymore, so I assume it's okay now.

node/tools/src/config.rs
IAvecilla: I forgot to delete it sry. I'll move all this unused code to a new branch beacuse it would be useful

node/actors/network/src/rpc/mod.rs
IAvecilla: I changed it to `push_batch_signature` to mantain consistency and go with just `batch` instead of `l1batch`

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl BatchNumber {
+    /// Increment the batch number.
+    pub fn next_batch_number(&self) -> BatchNumber {
+        BatchNumber(self.0.checked_add(1).unwrap_or(0))
+    }
+}
+
+/// A message to send by attesters to the gossip network.
+/// It contains the attester signature to sign the block batches to be sent to L1.
```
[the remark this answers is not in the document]
IAvecilla: Upon reviewing it again, I realize that the description of the struct seems incorrect. This struct holds information about the batch itself, and it's signed by the attesters before being propagated as a signed message in the gossip network. I revised it to better reflect this. Let me know if it's okay.

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,279 @@
...
+        self.total_weight
+    }
+
+    /// Maximal weight of faulty replicas allowed in this attester committee.
+    pub fn max_faulty_weight(&self) -> u64 {
+        max_faulty_weight(self.total_weight())
+    }
+}
+
+/// Calculate the consensus threshold, the minimum votes' weight for any consensus action to be valid,
```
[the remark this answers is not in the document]
IAvecilla: Should we delete it? I changed to make sense in the attesters context but I don't know if its strictly necessary

node/actors/executor/src/lib.rs
```diff
@@ -28,6 +28,12 @@ pub struct Validator {
     pub payload_manager: Box<dyn bft::PayloadManager>,
 }
 
+/// Validator-related part of [`Executor`].
```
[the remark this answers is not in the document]
IAvecilla: I actually deleted that struct because is not used at this time, we'll need to introduce it in a future test

node/libs/roles/src/attester/messages/l1_batch.rs
IAvecilla: Got it, I made some extra changes to the code, incorporating `anyhow` to utilize some helpful macros. Let me know if you prefer the previous code, or if you prefer to mimic this approach in the validators' code as well.

node/actors/network/src/rpc/push_signature.rs
IAvecilla: sgtm, I changed the name of the rpc method and the server to match with that.

node/libs/roles/src/attester/tests.rs
IAvecilla: I believe so, the timestamp will no longer be used anyway.

node/actors/network/src/gossip/batch_signatures.rs
IAvecilla: Seems good for now. We'll probably need a way to block while waiting for new batches, as you already mentioned. The State Keeper struct was added in the network actor; in the future it probably would be better to have it in a new actor that takes care of the related work with L1 batches. LMK what you think

node/actors/network/src/gossip/mod.rs
IAvecilla: Okay, that sounds reasonable. I've updated it to calculate the new QC from the last one viewed by that node. This way, the node will only collect signatures for its subsequent QC. However, I'm not entirely sure if this prevents the attack from a malicious attester. In the future, should we also check hashes and batch numbers?

node/actors/network/src/gossip/tests/mod.rs
pompon0: if it is unused, then yeah, I thought it was accidental.

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -0,0 +1,122 @@
...
+use super::{SignedBatchMsg, Signers};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default, PartialOrd)]
+/// A batch number.
+pub struct BatchNumber(pub u64);
+
+impl BatchNumber {
+    /// Increment the batch number.
+    pub fn next_batch_number(&self) -> BatchNumber {
+        BatchNumber(self.0.checked_add(1).unwrap_or(0))
```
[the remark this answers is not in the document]
pompon0: It is ok to panic here. Nothing can be done if we get out of numbers.

node/libs/roles/src/attester/messages/msg.rs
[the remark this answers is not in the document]
pompon0: well, some documentation is needed ;)

node/actors/network/src/rpc/push_signature.rs
pompon0: @IAvecilla I'm sorry for the inconsistency across my comments, I meant to replace "batch_signature(s)" to "batch_vote(s)" everywhere (except for the Signature struct itself of course).

node/actors/network/src/lib.rs
```diff
@@ -26,6 +30,20 @@ mod watch;
...
+    /// Get the L1 batch from storage with the highest number.
+    fn last_batch(&self) -> attester::BatchNumber;
+    /// Get the L1 batch QC from storage with the highest number.
+    fn last_batch_qc(&self) -> attester::BatchNumber;
+    /// Returns the batch with the given number.
+    fn get_batch(&self, number: attester::BatchNumber) -> Option<attester::Batch>;
+    /// Returns the QC of the batch with the given number.
+    fn get_batch_qc(&self, number: attester::BatchNumber) -> Option<attester::BatchQC>;
+    /// Store the given QC in the storage.
+    fn store_qc(&self, qc: attester::BatchQC);
```
pompon0, line 44: Please move this trait to the storage crate.

node/actors/network/src/proto/gossip.proto
```diff
@@ -17,6 +18,11 @@ message PushValidatorAddrs {
   repeated roles.validator.Signed net_addresses = 1;
 }
 
+message PushBatchVotes {
+  // Signed roles.validator.Msg.net_address.
```
pompon0, line 22: fix comment

node/actors/network/src/rpc/push_batch_votes.rs
```diff
@@ -0,0 +1,42 @@
+//! Defines RPC for passing consensus messages.
+use std::sync::Arc;
+
+use crate::{mux, proto::gossip as proto};
+use anyhow::Context as _;
+use zksync_consensus_roles::attester::{self, Batch};
+use zksync_protobuf::ProtoFmt;
+
+/// Signature RPC.
```
pompon0, line 9: push_batch_votes

node/actors/network/src/rpc/push_batch_votes.rs
```diff
@@ -0,0 +1,42 @@
...
+pub(crate) struct Rpc;
+
+impl super::Rpc for Rpc {
+    const CAPABILITY_ID: mux::CapabilityId = 5;
+    const INFLIGHT: u32 = 1;
+    const METHOD: &'static str = "push_signature";
```
pompon0, line 15: push_batch_votes

node/libs/roles/src/attester/messages/msg.rs
pompon0: in that case, let's just use validator::max_faulty_weight function instead for now.

node/libs/roles/src/validator/tests.rs
```diff
@@ -232,8 +223,14 @@ fn test_commit_qc() {
             qc.add(&key.sign_msg(qc.message.clone()), &setup1.genesis)
                 .unwrap();
         }
-        let expected_weight = i as u64 * validator_weight;
-        if expected_weight >= setup1.genesis.validators_committee.threshold() {
+        let expected_weight: u64 = setup1
+            .genesis
+            .attesters
+            .iter()
```
pompon0, line 229: get(setup1.validator_keys[i].public()).unwrap().weight

node/actors/network/src/gossip/mod.rs
```diff
@@ -141,39 +129,59 @@ impl Network {
                 .await
                 .context("batch signatures")?
                 .clone();
+
+            // Check next QC to collect signatures for.
+            let new_qc = self
+                .last_viewed_qc
+                .clone()
+                .and_then(|qc| {
+                    Some(attester::BatchQC::new(
```
pompon0, line 138: map?

node/actors/network/src/gossip/mod.rs
pompon0: Once we have the hash provided via StateKeeper interface, it should be all right (we only collect signatures for the batches that we expect). You can leave it as it is now, and we will upgrade it later. There is just a question of a race condition where the peer sends us a signature of a batch that we don't know about yet (i.e. we don't know the hash). But we will address this later as well.

node/actors/network/src/gossip/batch_signatures.rs
pompon0: Let's name it PersistentBatchStore and move it to storage crate. I suppose that it should be embedded somehow into the BlockStore, but I'll think about it more.

pompon0 (commented):
LG, I'll approve once comments are addressed and the CI passes - I want do double check the compatibility aspect.

node/actors/network/src/rpc/push_signature.rs
IAvecilla: Oh no, that actually makes a lot of sense. I must have overlooked that when I was initially addressing all the comments.

brunoffranca (approved):

node/actors/network/src/gossip/mod.rs
```diff
@@ -169,15 +169,27 @@ impl Network {
...
-                &self
-                    .batch_qc
-                    .get(&new_qc.message.number)
-                    .context("last qc")?
-                    .signers,
-            );
+            let weight = self
+                .genesis()
+                .attesters
+                .as_ref()
```
pompon0, line 175: extract the attesters once at the beginning of the function and return immediately if not available

node/actors/network/src/gossip/runner.rs
```diff
@@ -110,10 +113,8 @@ impl Network {
             ctx,
             self.cfg.rpc.push_block_store_state_rate,
         );
-        let push_signature_client = rpc::Client::<rpc::push_batch_votes::Rpc>::new(
-            ctx,
-            self.cfg.rpc.push_batch_signature_rate,
-        );
+        let push_signature_client =
+            rpc::Client::<rpc::push_batch_votes::Rpc>::new(ctx, self.cfg.rpc.push_batch_votes_rate);
```
pompon0, line 117: don't setup client/server if attesters is not present

node/actors/network/src/mux/transient_stream.rs
```diff
@@ -20,6 +20,7 @@ pub(crate) struct ReadStream(pub(super) sync::ExclusiveLock<ReadReusableStream>)
 pub(crate) struct WriteStream(pub(super) sync::ExclusiveLock<WriteReusableStream>);
 
 /// Error returned by `ReadStream::read` in case the stream has been closed by peer.
+#[allow(dead_code)]
```
pompon0, line 23: just remove it

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -10,7 +10,7 @@ pub struct BatchNumber(pub u64);
 impl BatchNumber {
     /// Increment the batch number.
     pub fn next(&self) -> BatchNumber {
-        BatchNumber(self.0.checked_add(1).unwrap_or(0))
+        BatchNumber(self.0 + 1)
```
pompon0, line 13: `checked_add(1).unwrap()`. Overflows are only checked in dev builds. In release build it will just wrap.

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -73,12 +73,12 @@ pub enum BatchQCAddError {
 
 impl BatchQC {
     /// Create a new empty instance for a given `Batch` message.
-    pub fn new(message: Batch, genesis: &Genesis) -> Self {
-        Self {
+    pub fn new(message: Batch, genesis: &Genesis) -> anyhow::Result<Self> {
+        Ok(Self {
             message,
-            signers: Signers::new(genesis.attesters.len()),
+            signers: Signers::new(genesis.attesters.as_ref().context("attesters")?.len()),
```
pompon0, line 79: no attester committe in genesis

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -88,6 +88,8 @@ impl BatchQC {
         ensure!(self.message == msg.msg, Error::InconsistentMessages);
         let i = genesis
             .attesters
+            .as_ref()
+            .expect("attesters set is empty in genesis") // This case should never happen
```
pompon0, line 92: don't panic, return an error

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -101,22 +103,25 @@ impl BatchQC {
     /// Verifies the signature of the BatchQC.
     pub fn verify(&self, genesis: &Genesis) -> Result<(), BatchQCVerifyError> {
         use BatchQCVerifyError as Error;
-        if self.signers.len() != genesis.attesters.len() {
+        let attesters = genesis
+            .attesters
+            .as_ref()
+            .expect("attesters set is empty in genesis"); // This case should never happen
+        if self.signers.len() != attesters.len() {
```
pompon0, line 110: ditto

node/libs/roles/src/attester/tests.rs
```diff
@@ -141,23 +141,37 @@ fn test_batch_qc() {
...
     let mut genesis3 = (*setup1.genesis).clone();
-    genesis3.attesters = Committee::new(setup1.genesis.attesters.iter().take(3).cloned()).unwrap();
+    genesis3.attesters = Committee::new(
+        setup1
+            .genesis
+            .attesters
+            .as_ref()
+            .unwrap()
+            .iter()
+            .take(3)
```
pompon0, line 151: setup1.genesis.attesters.as_ref().unwrap()[3].cloned()

node/libs/roles/src/attester/tests.rs
```diff
@@ -141,23 +141,37 @@ fn test_batch_qc() {
...
             .genesis
             .attesters
+            .as_ref()
+            .unwrap()
             .iter()
             .take(i)
             .map(|w| w.weight)
             .sum();
-        if expected_weight >= setup1.genesis.attesters.threshold() {
+        if expected_weight >= setup1.genesis.attesters.as_ref().unwrap().threshold() {
```
pompon0, line 174: again, unwrap just once at the beginning of the function

pompon0 (approved):

node/libs/roles/src/attester/tests.rs
IAvecilla: I believe it's not quite the same. The current implementation selects the first three attesters from the genesis, whereas with the new line, we would only pick the one in the third position. Although I don't see a problem in just picking one, since the genesis is only used for an error case for the verification. Perhaps that was the original idea behind the comment. I'll change it, and we can revert if necessary.

pompon0: oh, you are right. I misread the code. It was ok in the original form.

node/actors/network/src/gossip/runner.rs
```diff
@@ -113,9 +113,13 @@ impl Network {
...
-        let push_signature_client =
-            rpc::Client::<rpc::push_batch_votes::Rpc>::new(ctx, self.cfg.rpc.push_batch_votes_rate);
-        let push_signature_server = PushBatchVotesServer(self);
+        let mut push_signature_client: Option<rpc::Client<rpc::push_batch_votes::Rpc>> = None;
+        let mut push_signature_server: Option<PushBatchVotesServer> = None;
+        if self.genesis().attesters.is_some() {
+            push_signature_client =
+                rpc::Client::new(ctx, self.cfg.rpc.push_batch_votes_rate).into();
+            push_signature_server = PushBatchVotesServer(self).into();
+        };
```
pompon0, line 122: you should be able to create Client/Server adhoc in the if clause below, rather than go through this "Optional" dance.

pompon0: see the if statement with the ping client

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -106,7 +115,8 @@ impl BatchQC {
         let attesters = genesis
             .attesters
             .as_ref()
-            .expect("attesters set is empty in genesis"); // This case should never happen
+            .context("no attester committee in genesis")
```
pompon0, line 118: this line doesn't make sense, especially since you discard the error. use Option::ok_or/ok_or_else

pompon0:
remove -Dwarnings like here https://github.com/matter-labs/era-consensus/pull/114/files to fix the presubmit

node/actors/network/src/gossip/runner.rs
IAvecilla: Yeah, the approach used for the ping client is better. The thing is, I need the client below in [line 190](https://github.com/matter-labs/era-consensus/blob/5f352665542f8db9544578c5b1650317496d696c/node/actors/network/src/gossip/runner.rs#L190) to push the updates to the peer. I could avoid the server creation, but I need a reference to the client at the beginning to effectively use it there.

pompon0: you can put both of these into a single if, no?

node/actors/network/src/gossip/mod.rs
```diff
@@ -103,4 +117,72 @@ impl Network {
...
+                        self.genesis(),
+                    )
+                })
+                .context("new qc")?;
+
+            // Check votes for the correct QC.
+            for (_, sig) in votes.0 {
+                if self
+                    .batch_qc
+                    .clone()
```
aakoshh, line 160: It's not obvious to me why we clone the map in each step. Don't we want to accumulate all signatories before the `weight` is checked next? 

Also, shouldn't these signatures added stay in `batch_qc` after `update_batch_qc` returns? Why doesn't it take `&mut self`? 

If this is just done for the sake of checking individual voters, why does the total weight get checked here as well?

node/actors/network/src/gossip/mod.rs
```diff
@@ -103,4 +117,72 @@ impl Network {
...
+                    .get(&new_qc.message.number)
+                    .context("last qc")?
+                    .signers,
+            );
+
+            if weight < attesters.threshold() {
+                return Ok(());
+            };
+
+            // If we have enough weight, we can update the last viewed QC and propagate it.
```
aakoshh, line 185: I don't understand why we return if the weight is _less_ than the threshold. Shouldn't it wait for more votes?

node/actors/network/src/lib.rs
```diff
@@ -152,6 +162,8 @@ impl Runner {
                 }
             }
 
+            // TODO: check if we are active attester to get new L1 Batches, sign them and broadcast the signature
```
aakoshh, line 165: What happens with L1 batches that aren't new but no submitter has managed either to gather enough signatures over the gossip channel or to submit it to L1. Would attesters eventually re-broadcast their signatures?

aakoshh (commented):
I am sending some comments to ask some conceptual questions and to help me understand the code. Sorry to add more to the long comment history, I hope you don't mind.

As a general note:
* I see some TODOs left in the code. It would be helpful to state whether this PR completely implements what it says in the description, or there are things deferred to future PRs (perhaps with their ticket number). 
* Perhaps this PR could have been delivered in two parts: the attestations model first, and then hooking them into the flow.

node/actors/network/src/gossip/mod.rs
aakoshh: I also don't really understand the why the network actor [finishes the update task](https://github.com/matter-labs/era-consensus/pull/76/files#diff-7c22d1cc31d4af9bc4765751853b7c0161b50ebf892cf808a7553e1e951817e6R125-R129) if we return from here. Is that forever?

node/actors/network/src/gossip/runner.rs
IAvecilla: Right 👍

IAvecilla:
@aakoshh Most of your comments are valid and correct. However, the current logic won't work as intended because we don't have the syncing batches logic in place yet (which is in progress). This means that much of this code will change in the future once we can test it properly. This was just an initial implementation to see how the signature propagation would look.

I think @pompon0 mentioned something similar while I was writing this. :)

node/actors/network/src/gossip/mod.rs
aakoshh: Right, thanks. To be clear, I didn't start looking at this PR to understand the codebase, rather, I asked this question because I really didn't understand what the code in the PR was trying to achieve, and the description didn't contain any caveats about it being incomplete or nonsensical.

node/actors/network/src/gossip/mod.rs
aakoshh: Ah, I see, so this just arbitrarily quits. I didn't get that from the PR description, so I thought it's worth asking why I so totally don't get it.

node/actors/network/src/lib.rs
aakoshh: Thanks; IIUC the broadcasts and gossips continue until the cert appears on L1.

This is what I wanted to confirm. I asked @brunoffranca after watching one of his presentations; suggested looking at this PR.

PR #77, opened 2024-03-14
Weighted validators (BFT-408, BFT-424)

ElFantasma:
## What ❔

Add weights to the validator set and change the consensus threshold rules to require >80% of weight.

## Why ❔

We need for the L1 contracts to only accept blocks that were built by our node (which is the single leader). We can give more weight to our node, so that a QC isn't valid unless it signs. In turn this will be useful in the future


node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -127,11 +127,22 @@ impl StateMachine {
...
             .commit_message_cache
             .entry(message.view.number)
             .or_default();
+
+        // We check validators weight from current messages, stored by proposal
+        let mut by_proposal_before: HashMap<_, Vec<_>> = HashMap::new();
+        let entry_before = cache_entry.clone();
+        for msg in entry_before.values() {
+            by_proposal_before
+                .entry(msg.msg.proposal)
```
pompon0, line 140: you can just compute the weight of the proposal of the new message, rather than for all proposals.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -140,12 +151,21 @@ impl StateMachine {
             by_proposal.entry(msg.msg.proposal).or_default().push(msg);
         }
         let threshold = self.config.genesis().validators.threshold();
-        let Some((_, replica_messages)) =
-            by_proposal.into_iter().find(|(_, v)| v.len() >= threshold)
+        let Some((proposal, _replica_messages)) = by_proposal
```
pompon0, line 154: nit: there is a semantic difference between nameing a message "_" (value is dropped immediately) and "_sth" (value is dropped at the end of the scope).

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -137,28 +137,38 @@ impl StateMachine {
             .or_insert_with(|| validator::PrepareQC::new(message.view.clone()))
             .add(&signed_message, self.config.genesis());
 
-        // We store the message in our cache.
-        self.prepare_message_cache
+        // Work on current view messages
+        let entry = self
```
pompon0, line 141: please consider fixing https://linear.app/matterlabs/issue/BFT-424/fix-aggregation-of-replicacommit-messages in the same pr.

node/libs/roles/src/proto/validator.proto
```diff
@@ -166,3 +166,9 @@ message Signature {
 message AggregateSignature {
   optional bytes bn254 = 1; // required
 }
+
+message WeightedValidator {
+  optional PublicKey key = 1; // required
+  optional uint32 weight = 2; // required
```
pompon0, line 172: why not uint64?

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -1,10 +1,12 @@
 //! Messages related to the consensus protocol.
 use super::{BlockNumber, LeaderCommit, LeaderPrepare, Msg, ReplicaCommit, ReplicaPrepare};
 use crate::validator;
+use anyhow::Context;
```
pompon0, line 4: nit: `use anyhow::Context as _;`

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,25 +76,48 @@ impl Default for Fork {
 /// A struct that represents a set of validators. It is used to store the current validator set.
 /// We represent each validator by its validator public key.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct ValidatorSet {
+pub struct ValidatorCommittee {
     vec: Vec<validator::PublicKey>,
-    map: BTreeMap<validator::PublicKey, usize>,
+    indexes: BTreeMap<validator::PublicKey, usize>,
+    weights: Vec<usize>,
```
pompon0, line 82: why weights are not stored in the map?

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,25 +76,48 @@ impl Default for Fork {
...
 }
 
-impl ValidatorSet {
-    /// Creates a new ValidatorSet from a list of validator public keys.
-    pub fn new(validators: impl IntoIterator<Item = validator::PublicKey>) -> anyhow::Result<Self> {
+impl ValidatorCommittee {
+    /// Maximum validator weight
+    /// 100.00%
```
pompon0, line 87: why are we hardcoding the precision here? Why not allow arbitrary weights?

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -128,15 +161,37 @@ impl ValidatorSet {
...
+        self.weights
+            .iter()
+            .enumerate()
+            .filter(|(i, _)| signers.0[*i])
+            .map(|(_, weight)| weight)
+            .sum()
+    }
+
+    /// Compute the sum of signers weights.
+    pub fn weight_from_msgs<T: Variant<Msg>>(&self, signed: &[&validator::Signed<T>]) -> usize {
```
pompon0, line 185: this method doesn't take into account the possibility that the same signer occurs more than once.

node/libs/roles/src/validator/messages/leader_commit.rs
```diff
@@ -100,12 +100,12 @@ impl CommitQC {
             return Err(Error::BadSignersSet);
         }
 
-        // Verify that we have enough signers.
-        let num_signers = self.signers.count();
+        // Verify the signer's weights is enough.
```
pompon0, line 103: signers' weight is enough

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,25 +76,48 @@ impl Default for Fork {
...
 }
 
-impl ValidatorSet {
-    /// Creates a new ValidatorSet from a list of validator public keys.
-    pub fn new(validators: impl IntoIterator<Item = validator::PublicKey>) -> anyhow::Result<Self> {
+impl ValidatorCommittee {
+    /// Maximum validator weight
+    /// 100.00%
+    pub const MAX_WEIGHT: usize = 10000;
+
```
pompon0, line 89: why not uint

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,25 +76,48 @@ impl Default for Fork {
...
+    /// Required weight to verify signatures.
+    /// Currently 80.00%
+    const THRESHOLD: usize = 8000;
```
pompon0, line 92: ditto

node/libs/roles/src/validator/messages/consensus.rs
pompon0: also all the arithmetic on this type should be checked for overflows

node/libs/roles/src/validator/conv.rs
```diff
@@ -437,3 +442,21 @@ impl ProtoFmt for AggregateSignature {
...
+        Ok(Self {
+            key: read_required(&r.key).context("key")?,
+            weight: *required(&r.weight).context("weight")?,
+        })
+    }
+
+    fn build(&self) -> Self::Proto {
+        Self::Proto {
+            key: Some(self.key.build()),
+            weight: Some(self.weight),
```
pompon0, line 459: you should check that the value is from the allowed range here

node/libs/roles/src/validator/conv.rs
```diff
@@ -437,3 +442,21 @@ impl ProtoFmt for AggregateSignature {
...
     }
 }
+
+impl ProtoFmt for WeightedValidator {
+    type Proto = proto::WeightedValidator;
+
+    fn read(r: &Self::Proto) -> anyhow::Result<Self> {
+        Ok(Self {
+            key: read_required(&r.key).context("key")?,
+            weight: *required(&r.weight).context("weight")?,
```
pompon0, line 452: you should check that the value is from the allowed range here

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,25 +76,48 @@ impl Default for Fork {
...
         let mut set = BTreeSet::new();
+        let mut weights = BTreeMap::new();
         for validator in validators {
-            anyhow::ensure!(set.insert(validator), "Duplicate validator in ValidatorSet");
+            anyhow::ensure!(
+                set.insert(validator.key.clone()),
+                "Duplicate validator in ValidatorCommittee"
+            );
+            weights.insert(
+                validator.key,
```
pompon0, line 104: you should check that the weights are from the allowed range and that the sum is 1. Also add a test for those checks

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: Ah, right!

ElFantasma: And actually we can do the same in later weight calculation

node/libs/roles/src/validator/messages/consensus.rs
brunoffranca: +1. There's no reason to have a max weight. Just calculate the threshold from the sum of the weights.

node/libs/roles/src/validator/messages/consensus.rs
brunoffranca: Disagree. There's no need to make the weights add up to 1.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -127,25 +126,42 @@ impl StateMachine {
...
+        let weight_before = self.config.genesis().validators.weight_from_msgs(
+            cache_entry
+                .clone()
+                .iter()
+                .map(|(_, m)| m)
+                .filter(|m| m.msg.proposal == message.proposal)
+                .collect::<Vec<_>>()
+                .as_slice(),
+        );
+
```
brunoffranca, line 144: nit: If this calculation is only used for the later debug_assert, then I don't think it's worth it. Just remove the assert.

brunoffranca (changes requested):

node/libs/roles/src/validator/messages/consensus.rs
ElFantasma: 👍

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -128,15 +161,37 @@ impl ValidatorSet {
...
     pub fn threshold(&self) -> usize {
-        threshold(self.len())
+        Self::THRESHOLD
     }
 
-    /// Maximal number of faulty replicas allowed in this validator set.
+    /// Maximal number of faulty replicas allowed in this validator committee.
     pub fn faulty_replicas(&self) -> usize {
         faulty_replicas(self.len())
     }
```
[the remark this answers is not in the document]
brunoffranca: Sorry, not methods, functions.

node/libs/roles/src/validator/messages/consensus.rs
ElFantasma: switched all weights to u64, and also removed this constant as it is no longer necessary

ElFantasma:
Last commits fix BFT-424. This is ready for a new review

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -118,37 +118,32 @@ impl StateMachine {
...
-            .add(&signed_message, self.config.genesis());
+            .or_default()
+            .entry(message.clone())
+            .or_insert_with(|| CommitQC::new(message.clone(), self.config.genesis()));
+        commit_qc.add(&signed_message, self.config.genesis());
 
         // We store the message in our cache.
-        let cache_entry = self
-            .commit_message_cache
+        self.commit_message_cache
```
pompon0, line 131: do we still need it? It duplicates data with commit_qcs

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -118,37 +118,32 @@ impl StateMachine {
...
         // We add the message to the incrementally-constructed QC.
-        self.commit_qcs
+        let commit_qc = self
+            .commit_qcs
             .entry(message.view.number)
-            .or_insert_with(|| CommitQC::new(message.clone(), self.config.genesis()))
-            .add(&signed_message, self.config.genesis());
+            .or_default()
+            .entry(message.clone())
+            .or_insert_with(|| CommitQC::new(message.clone(), self.config.genesis()));
```
pompon0, line 127: we need to protect ourselves from a malicious validator spamming us with many replicacommits for the same view

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -132,34 +132,34 @@ impl StateMachine {
...
+            .insert(author.clone(), signed_message.clone());
+
+        // Now we check if we have enough weight to continue.
+        let weight = self.config.genesis().validators.weight(
+            prepare_qc
+                .map
+                .get(&signed_message.msg)
+                .unwrap_or(&Signers::new(self.config.genesis().validators.len()))
+                .clone(),
+        );
```
pompon0, line 154: here you should sum weights over all messages collected for the given view, not just the one just received

pompon0: and please add a test for it

ElFantasma: mmhh, I believe that only the weights for the message received are able to exceed the threshold 🤔

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -224,15 +224,24 @@ impl UTHarness {
         ctx: &ctx::Ctx,
         msg: ReplicaPrepare,
     ) -> Signed<LeaderPrepare> {
+        let expected_validator_weight =
```
pompon0, line 227: use the weights from genesis().validators, don't recompute it

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -252,15 +261,24 @@ impl UTHarness {
         ctx: &ctx::Ctx,
         msg: ReplicaCommit,
     ) -> Signed<LeaderCommit> {
+        let expected_validator_weight =
```
pompon0, line 264: ditto

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -83,7 +83,7 @@ impl UTHarness {
     /// Creates a new `UTHarness` with minimally-significant validator set size.
     pub(crate) async fn new_many(ctx: &ctx::Ctx) -> (UTHarness, BlockStoreRunner) {
         let num_validators = 6;
-        assert!(validator::faulty_replicas(num_validators) > 0);
+        assert!(validator::max_faulty_weight(num_validators as u64) > 0);
```
pompon0, line 86: just assert max_fault_weight on the `UTHarness::new().replica.validators` instead

node/libs/roles/src/validator/tests.rs
```diff
@@ -254,12 +261,13 @@ fn test_prepare_qc() {
...
             );
         }
-        if n >= setup1.genesis.validators.threshold() {
+        let expected_weight = n as u64 * setup1.genesis.validators.total_weight() / 6;
+        if expected_weight >= setup1.genesis.validators.threshold() {
             qc.verify(&setup1.genesis).unwrap();
         } else {
             assert_matches!(
                 qc.verify(&setup1.genesis),
-                Err(Error::NotEnoughSigners { .. })
```
pompon0, line 262: nit: IMO NotEnoughSigners is more comprehensive name than WeightNotReached. All that changes is the definition of "enough".

node/libs/roles/src/validator/tests.rs
```diff
@@ -268,3 +276,60 @@ fn test_prepare_qc() {
...
+
+#[test]
+fn test_validator_weights_sanity() {
+    let ctx = ctx::test_root(&ctx::RealClock);
+    let rng = &mut ctx.rng();
+
+    let setup = Setup::new(rng, 6);
+    // Validators weights
+    let weight = u64::MAX / 5;
+    let weights = [weight, weight, weight, weight, weight, weight];
```
pompon0, line 323: nit: you can equivalently write `[weight;6]`: https://gist.github.com/rust-play/0cc713303e7094544cfc881cb5b65ff6

node/actors/bft/src/leader/replica_prepare.rs
pompon0: prepareqc consists of signatures on potentially different messages, which just share view number. It is the total weight over all such messages that is supposed to exceed threshold.

node/libs/roles/src/proto/validator.proto
```diff
@@ -11,7 +11,7 @@ message Fork {
 
 message Genesis {
   optional Fork fork = 1; // required
-  repeated PublicKey validators = 2;
```
pompon0, line 14: We already have a deployment (stage), so please try to keep backward compatibility. It also means that we should support the old way of specifying validators (and make sure that hashes are stable) for the time being.

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: It works just as before 🤷‍♂️. It aggregates by view and validator while `commit_qcs` aggregates by view and (now) by message.

I guess we can obtain the validator from the CommitQC signers, but that's not optimal and also could have been done with previous `commit_qc` version.

node/actors/bft/src/testonly/ut_harness.rs
ElFantasma: Mhh, we can do that, but then a bug in weights distribution won't be catch in the asserts inside the for loop. Is that ok?

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: If I understand the code correctly, that should be prevented in the ["Duplicated message" check](https://github.com/matter-labs/era-consensus/blob/bf36b3c2844f742a83a51e553f49c86b11faa78a/node/actors/bft/src/leader/replica_commit.rs#L100), right?

But maybe we can refactor a bit the code to let `commit_qcs` aggregate by view and validator (just like `commit_message_cache`) instead of view and message, and then yes, remove the redundant data structure while performing the same checks. Does it make sense?

node/actors/bft/src/leader/replica_prepare.rs
brunoffranca: +1

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: Ah no, that won't work. I'm aggregating by message purposely to validate the _same_ message gets enough signers' weight

node/actors/bft/src/testonly/ut_harness.rs
ElFantasma: But that will fail on a number of existing tests that are creating validator committees with [one](https://github.com/matter-labs/era-consensus/blob/bf36b3c2844f742a83a51e553f49c86b11faa78a/node/actors/bft/src/leader/tests.rs#L381) or [two](https://github.com/matter-labs/era-consensus/blob/bf36b3c2844f742a83a51e553f49c86b11faa78a/node/actors/bft/src/leader/tests.rs#L478) validators only

node/actors/bft/src/testonly/ut_harness.rs
pompon0: that's ok

node/actors/bft/src/testonly/ut_harness.rs
pompon0: you should put the assert in the new_many constructor only. Just like it is now.

node/actors/bft/src/leader/replica_commit.rs
pompon0: ok

node/actors/bft/src/leader/replica_prepare.rs
ElFantasma: Fixed, and added a test case for it

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -145,13 +145,12 @@ impl StateMachine {
...
-            prepare_qc
-                .map
-                .get(&signed_message.msg)
-                .unwrap_or(&Signers::new(self.config.genesis().validators.len()))
-                .clone(),
-        );
+        let weight: u64 = prepare_qc
+            .clone()
+            .map
+            .into_values()
```
pompon0, line 151: `prepare_qc.map.values()`, no cloning should be necessary

node/actors/bft/src/leader/tests.rs
```diff
@@ -333,6 +333,68 @@ async fn replica_prepare_high_qc_of_future_view() {
...
+        // half of the validators sign replica_prepare
+        for i in 0..validators / 2 {
+            util.process_replica_prepare(ctx, util.keys[i].sign_msg(replica_prepare.clone()))
+                .await
+                .unwrap();
+        }
+
+        let mut replica_commit_result = None;
+        // The rest (minus one) of the validators sign other_replica_prepare
+        for i in validators / 2..validators - 1 {
```
pompon0, line 379: why "-1"?

node/actors/bft/src/testonly/ut_harness.rs
```diff
@@ -226,14 +226,13 @@ impl UTHarness {
...
             self.genesis().validators.total_weight() / self.keys.len() as u64;
-        let want_threshold = self.genesis().validators.threshold();
         let mut leader_prepare = None;
         let msgs: Vec<_> = self.keys.iter().map(|k| k.sign_msg(msg.clone())).collect();
         let mut first_match = true;
         for (i, msg) in msgs.into_iter().enumerate() {
             let res = self.process_replica_prepare(ctx, msg).await;
             match (
-                (i + 1) as u64 * expected_validator_weight < want_threshold,
+                (i + 1) as u64 * expected_validator_weight < self.genesis().validators.threshold(),
```
pompon0, line 235: aggregate the validators weight on this side as well.

pompon0: i.e. get rid of expected_validator_weight altogether

node/actors/bft/src/leader/replica_commit.rs
pompon0: I mean commit_message_cache unnecessarily contains the messages. It would be enough if it just contained the sets of validators - or even better, we could have a validator -> view map which will just indicate the highest view of messages that we have received from the given validator. And yes, I know it was suboptimal before, but we can fix it.

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -118,37 +118,32 @@ impl StateMachine {
...
-        let threshold = self.config.genesis().validators.threshold();
-        let Some((_, replica_messages)) =
-            by_proposal.into_iter().find(|(_, v)| v.len() >= threshold)
-        else {
+        // Now we check if we have enough weight to continue.
+        let weight = self
+            .config
+            .genesis()
+            .validators
+            .weight(commit_qc.signers.clone());
```
pompon0, line 141: nit: `weight()` could just take a reference as an argument

node/actors/bft/src/tests.rs
```diff
@@ -9,7 +9,10 @@ async fn run_test(behavior: Behavior, network: Network) {
 
     const NODES: usize = 11;
     let mut nodes = vec![behavior; NODES];
-    for n in &mut nodes[0..validator::threshold(NODES)] {
+    // validator::threshold(NODES) will calculate required nodes to validate a message
+    // assuming uniform weight distribution
```
pompon0, line 13: it would be better to make test accept [(Behavior,weight)] rather than [Behavior] here. This way we won't need to assume stuff about the `Test::run()`.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,95 +75,132 @@ impl Default for Fork {
...
 }
 
-impl ValidatorSet {
-    /// Creates a new ValidatorSet from a list of validator public keys.
-    pub fn new(validators: impl IntoIterator<Item = validator::PublicKey>) -> anyhow::Result<Self> {
+impl ValidatorCommittee {
+    /// Creates a new ValidatorCommittee from a list of validator public keys.
+    pub fn new(validators: impl IntoIterator<Item = WeightedValidator>) -> anyhow::Result<Self> {
         let mut set = BTreeSet::new();
+        let mut weighted_validators = BTreeMap::new();
```
pompon0, line 88: you shouldn't need both set and weighted_validators here

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,95 +75,132 @@ impl Default for Fork {
...
+    pub fn max_faulty_weight(&self) -> u64 {
+        max_faulty_weight(self.total_weight())
     }
 
-    /// Maximal number of faulty replicas allowed in this validator set.
-    pub fn faulty_replicas(&self) -> usize {
-        faulty_replicas(self.len())
+    /// Compute the sum of signers weights.
+    pub fn weight(&self, signers: Signers) -> u64 {
+        self.vec
```
pompon0, line 168: add an assert checking that length of Signers matches length of self.vec. Document that it panics if lengths do not match.

node/libs/roles/src/validator/testonly.rs
```diff
@@ -22,8 +23,13 @@ impl Setup {
     /// New `Setup` with a given `fork`.
     pub fn new_with_fork(rng: &mut impl Rng, validators: usize, fork: Fork) -> Self {
         let keys: Vec<SecretKey> = (0..validators).map(|_| rng.gen()).collect();
+        let weight = 1;
         let genesis = Genesis {
-            validators: ValidatorSet::new(keys.iter().map(|k| k.public())).unwrap(),
+            validators: ValidatorCommittee::new(keys.iter().map(|k| WeightedValidator {
```
pompon0, line 28: nit: we can shorten "ValidatorCommittee" to just "Committee".

node/libs/roles/src/validator/tests.rs
```diff
@@ -268,3 +276,60 @@ fn test_prepare_qc() {
...
+        .iter()
+        .enumerate()
+        .map(|(i, w)| WeightedValidator {
+            key: setup.keys[i].public(),
+            weight: *w,
+        })
+        .collect();
+
+    // Creation should overflow
+    assert_matches!(ValidatorCommittee::new(validators), Err(Error { .. }));
```
pompon0, line 334: Err(_) should be enough

node/actors/bft/src/leader/tests.rs
ElFantasma: Because threshold will be reached after 5 validator signatures. I'll add the explanation to the comment

node/libs/roles/src/validator/testonly.rs
ElFantasma: I like that, but this was a suggestion from @brunoffranca. Do you agree, Bruno?

node/libs/roles/src/validator/tests.rs
ElFantasma: Ah yes, that was my attempt to refine the expected error better.

node/libs/roles/src/proto/validator.proto
ElFantasma: Ok, fixed this creating the 3rd field but tried to keep the `validators` name for it to use `genesis.validators` all around the code, but the checker complains with:
```
Error: Field "2" with name "validator_keys" on message "Genesis" changed option "json_name" from "validators" to "validatorKeys".
Error: Field "2" on message "Genesis" changed name from "validators" to "validator_keys".
Error: buf found 2 breaking changes.
```

So, how should I fix this? 
- Should I change the name to the 3rd field? (that will make the code to use something like `genesis.weighted_validators` everywhere)
- Should I use `[json_name="validators"];` and `[json_name="weighted_validators"];` for 2nd and 3rd fields to let it keep the json names and allow us to use `genesis.validators` in our code? (that will make json representation a bit confusing, not sure if we are using that. Also I haven't tried this, not sure if it will work)
- Should I keep the `validator` name in the `proto::Genesis` struct but not in the `validator::Genesis` one? (that way field names won't match but the code will still be `genesis.validators` everywhere

It's a kind of a silly question, but I'm not comfortable with none of the options here 🤔

node/actors/bft/src/leader/tests.rs
ElFantasma: Or better, I'll make it independent of the number of validators required to commit

node/libs/roles/src/validator/testonly.rs
pompon0: to clarify: we will use it as "validator::Commitee" all over the place anyway.

node/libs/roles/src/proto/validator.proto
ElFantasma: cool, thanks

node/actors/bft/src/tests.rs
ElFantasma: Problem here is that we have to calculate the amount of honest nodes, and for that we are using these assumptions:
- each validator.weight = 1 (which is the default for tests)
- total_weight = number of nodes
I'll change the comment to explain this.

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: Ok, solved this checking if the author signature is in the qc.

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -132,34 +132,32 @@ impl StateMachine {
...
-            .prepare_message_cache
-            .get(&message.view.number)
-            .unwrap()
-            .len();
-
-        if num_messages < self.config.genesis().validators.threshold() {
+            .insert(author.clone(), signed_message.clone());
+
+        // Now we check if we have enough weight to continue.
+        let weight: u64 = prepare_qc
```
pompon0, line 148: turn it into a method `PrepareQC::weight(&self,&validators::Commitee)`

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -96,14 +96,18 @@ impl StateMachine {
...
-        if let Some(existing_message) = self
-            .commit_message_cache
-            .get(&message.view.number)
-            .and_then(|x| x.get(author))
-        {
+        if commit_qc.signers.0[self.config.genesis().validators.index(author).unwrap()] {
             return Err(Error::DuplicateMessage {
-                existing_message: existing_message.msg.clone(),
+                existing_message: commit_qc.message.clone(),
             });
```
pompon0, line 111: it is not an issue that a replica signs the same ReplicaCommit twice - CommitQC::add filters out the duplicates (see the implementation). The problem is that the same replica might sign DIFFERENT ReplicaCommit for the same view multiple times and THAT's what we should check here.

node/actors/bft/src/testonly/ut_harness.rs
pompon0:
```
let (x,y) = UTHarness::new(Ctx, num_validators).await;
assert!(x.genesis().validators.max_faulty_weight()>0);
(x,y)
```

node/libs/roles/src/validator/messages/tests.rs
```diff
@@ -33,8 +33,13 @@ fn fork() -> Fork {
 /// Hardcoded genesis.
 fn genesis() -> Genesis {
     Genesis {
-        validators: ValidatorSet::new(keys().iter().map(|k| k.public())).unwrap(),
+        validators: Committee::new(keys().iter().map(|k| WeightedValidator {
```
pompon0, line 36: good, also add a change detector test for the new encoding

node/libs/roles/src/proto/validator.proto
```diff
@@ -11,7 +11,8 @@ message Fork {
...
+  repeated PublicKey validators = 2  [deprecated = true];
```
pompon0, line 14: document that validators and validators_v1 are mutually exclusive. Technically you should be allowed to put those fields into a oneof, but buf is too dumb to recognize this as a backward compatible change: https://buf.build/docs/breaking/rules#field_same_oneof

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -74,96 +75,132 @@ impl Default for Fork {
...
-impl ValidatorSet {
-    /// Creates a new ValidatorSet from a list of validator public keys.
-    pub fn new(validators: impl IntoIterator<Item = validator::PublicKey>) -> anyhow::Result<Self> {
-        let mut set = BTreeSet::new();
+impl Committee {
+    /// Creates a new Committee from a list of validator public keys.
+    pub fn new(validators: impl IntoIterator<Item = WeightedValidator>) -> anyhow::Result<Self> {
+        let mut weighted_validators = BTreeMap::new();
+        let mut total_weight: u64 = 0;
         for validator in validators {
```
pompon0, line 90: we should also add test checking what happens if some/all validators have weight 0.

node/actors/bft/src/tests.rs
ElFantasma: Ok, I misunderstood your first comment

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: Ok. Just to be sure, this was the same in the previous check with the `commit_message_cache`? 
I mean, it's not that new code behaves differently: Same `ReplicaCommit` signed twice by the same replica was raising `DuplicateError` in previous version as well, right?

node/libs/roles/src/validator/conv.rs
```diff
@@ -27,25 +27,58 @@ impl ProtoFmt for Fork {
...
+                            weight: 1,
+                        }))
+                        .collect::<Result<_,_>>()
+                        .context("validators")?,
+                    0,
+                )
+            // empty validator set
+            } else {
+                (vec![], 0)
+            };
```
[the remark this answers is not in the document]
ElFantasma: Ah yes, I was raising an error there, but changed it to prevent code repetition. 
Will update the comment 👍

node/libs/roles/src/validator/messages/consensus.rs
ElFantasma: Should we raise an error on a 0 weight validator?
It seems it shouldn't be allowed, but maybe there is one scenario where it can happen or it is useful (like testing validators without vote or something like that)

node/actors/bft/src/leader/replica_commit.rs
ElFantasma: Actually, existing test is expecting an `Error::DuplicateMessage` when the `ReplicaCommit`s [match](https://github.com/matter-labs/era-consensus/blob/2f581685b5bdeb854b1d697966585ee7c517ff3c/node/actors/bft/src/leader/tests.rs#L555).
Is that test wrong, so?

ElFantasma: Ok @pompon0, @brunoffranca, need clarification for this:

**Previous version**: checks if same validator signed twice for same view **any** message        (will return error if signs _different_ or _same_ message).
 **Current version**: checks if same validator signed twice for same view **the same** message   (will return error _only_ if it signs _same_ message).
**Desired version?**: checks if same validator signed twice for same view **different** messages (will return error _only_ if it signs _different_ messages).

Do we agree we want **Desired version?**

node/libs/roles/src/validator/messages/consensus.rs
ElFantasma: I added a check (and a test) to forbid 0 weight validators in committee. Let me know if we should do otherwise.

node/actors/bft/src/leader/replica_commit.rs
brunoffranca: Either desired or previous version is okay.

pompon0: yes, desired and previous are okay. My point was to just reduce the amount of information we store (previous implementation stored messages both in aggregated and non-aggregated form, while it would be enough to remember aggregated form + for each validator what was the highest view of messages sent by them).

node/libs/roles/src/validator/messages/consensus.rs
pompon0: sg

node/libs/roles/src/validator/conv.rs
```diff
@@ -57,7 +57,7 @@ impl ProtoFmt for Genesis {
                         .context("validators")?,
                     0,
                 )
-            // empty validator set
+            // empty validator set, will raise an exception in Committee:new()
```
pompon0, line 60: there is no "raising exceptions" in rust. Use "Committe::new() will return an error" or similar.

node/libs/roles/src/validator/messages/tests.rs
```diff
@@ -31,15 +31,15 @@ fn fork() -> Fork {
 }
 
 /// Hardcoded genesis.
-fn genesis() -> Genesis {
+fn genesis(encoding_version: usize) -> Genesis {
```
pompon0, line 34: at this point genesis version deserves a separate type, instead of usize.

node/libs/roles/src/validator/messages/tests.rs
```diff
@@ -31,15 +31,15 @@ fn fork() -> Fork {
...
     Genesis {
         validators: Committee::new(keys().iter().map(|k| WeightedValidator {
```
pompon0, line 36: nit: a separate constructor for v0 would be nicer

node/libs/roles/src/validator/tests.rs
```diff
@@ -308,28 +293,40 @@ fn test_validator_committee_weights() {
...
     // Creation should overflow
     assert_matches!(Committee::new(validators), Err(_));
 }
+
+#[test]
+fn test_committee_with_zero_weights() {
+    let ctx = ctx::test_root(&ctx::RealClock);
+    let rng = &mut ctx.rng();
+
+    let validators: Vec<WeightedValidator> = vec![1000, 0, 800, 6000, 0, 700]
```
pompon0, line 322: nit: just slice will do

pompon0 (approved):

brunoffranca (approved):

PR #81, opened 2024-03-25
Adds validity checks to BLS public keys

brunoffranca:
## What ❔

Added checks for the validity of our BLS public keys as recommended by this [spec](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html). Namely we check that the public key is not zero and that it is in the correct subgroup.

We do it redundantly in two places (when we decode a public key and when we verify a signature).

## Why ❔

Invalid public keys are a security risk.

### DIFF
--- a/node/libs/crypto/src/bn254/mod.rs
+++ b/node/libs/crypto/src/bn254/mod.rs
@@ -81,6 +81,24 @@ impl PartialEq for SecretKey {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PublicKey(G2);
 
+impl PublicKey {
+    /// Checks if the public key is not the identity element and is in the correct subgroup. Verifying signatures
+    /// against public keys that are not valid is insecure.
+    pub fn is_valid(&self) -> bool {
+        // Check that the point is not the identity element.
+        if self.0.is_zero() {
+            return false;
+        }
+
+        // We multiply the point by the order and check if the result is the identity element.
+        // If it is, then the point is on the correct subgroup.
+        let order = Fr::char();
+        let mut p = self.0;
+        p.mul_assign(order);
+        p.is_zero()
+    }
+}
+
 impl Default for PublicKey {
     fn default() -> Self {
         PublicKey(G2::zero())
@@ -97,16 +115,13 @@ impl ByteFmt for PublicKey {
     fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
         let arr: [u8; 64] = bytes.try_into()?;
         let p = G2Compressed::from_fixed_bytes(arr).into_affine()?;
+        let pk = PublicKey(p.into());
 
-        if p.is_zero() {
-            anyhow::bail!("Public key can't be zero")
+        if !pk.is_valid() {
+            anyhow::bail!("Public key is not valid")
         }
 
-        if p.scale_by_cofactor().is_zero() {
-            anyhow::bail!("Public key must be in the subgroup")
-        }
-
-        Ok(PublicKey(p.into_projective()))
+        Ok(pk)
     }
 
     fn encode(&self) -> Vec<u8> {
@@ -139,15 +154,9 @@ impl Signature {
     pub fn verify(&self, msg: &[u8], pk: &PublicKey) -> Result<(), Error> {
         let hash_point = hash::hash_to_g1(msg);
 
-        // Verify public key. Since we already check the validity of a
+        // Verify public key is valid. Since we already check the validity of a
         // public key when constructing it, this should never fail (in theory).
-        if pk.0.is_zero() {
-            unreachable!();
-        }
-
-        if pk.0.into_affine().scale_by_cofactor().is_zero() {
-            unreachable!();
-        }
+        assert!(pk.is_valid());
 
         // First pair: e(H(m): G1, pk: G2)
         let a = Bn256::pairing(hash_point, pk.0);
@@ -212,8 +221,13 @@ impl AggregateSignature {
         // is one fewer pairing to calculate.
         let mut pairs: HashMap<&[u8], PublicKey> = HashMap::new();
         for (msg, pk) in msgs_and_pks {
+            // Verify public key is valid. Since we already check the validity of a
+            // public key when constructing it, this should never fail (in theory).
+            assert!(pk.is_valid());
+
             pairs.entry(msg).or_default().0.add_assign(&pk.0);
         }
+
         // First pair: e(sig: G1, generator: G2)
         let a = Bn256::pairing(self.0, G2::one());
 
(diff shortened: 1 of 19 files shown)


pompon0:
Please add a test for the constructor.

dnkolegov (approved):
Great work @brunoffranca. Thanks.

brunoffranca:
@pompon0 I don't see a way of creating a point outside the correct subgroup. If I think of anything meanwhile, I'll add it.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -132,6 +139,16 @@ impl Signature {
     pub fn verify(&self, msg: &[u8], pk: &PublicKey) -> Result<(), Error> {
         let hash_point = hash::hash_to_g1(msg);
 
+        // Verify public key. Since we already check the validity of a
+        // public key when constructing it, this should never fail (in theory).
+        if pk.0.is_zero() {
```
pompon0, line 144: nit `assert!(!pk.0.is_zero());`, analogically the other check

node/libs/crypto/src/bn254/tests.rs
```diff
@@ -89,6 +89,18 @@ fn aggregate_signature_failure_smoke() {
...
 
+#[test]
+fn pk_validity_checks() {
+    // Check that we can't decode the null point.
+    let pk = PublicKey::default();
+    let bytes = pk.encode();
+    let pk_decoded = PublicKey::decode(&bytes);
+    let err = pk_decoded.unwrap_err();
+    assert_eq!(format!("{}", err.root_cause()), "Public key can't be zero");
+
```
pompon0, line 100: don't compare human readable error messages. Just use `assert!(pk_decoded.is_err())`. Alternatively implement a decode() method which returns a strongly typed error that you may match against and make ByteFmt::decode just call your custom decode.

node/libs/crypto/src/bn254/tests.rs
```diff
@@ -89,6 +89,18 @@ fn aggregate_signature_failure_smoke() {
...
+    // TODO: Create a G2 point in the wrong subgroup and test against it.
```
pompon0, line 101: I suppose you can sample at random public key material, until you find a point in the wrong subgroup and then pass it to decode() to check that it also returns an error.

pompon0: or sample and hardcode into the test.

pompon0 (approved):

node/libs/crypto/src/bn254/tests.rs
brunoffranca: Nope. Random G2 generation always [creates points in the correct subgroup](https://github.com/matter-labs/pairing/blob/master/src/bn256/ec.rs#L1149).
There's a [test](https://github.com/matter-labs/pairing/blob/master/src/bn256/ec.rs#L1522) that I can port. I'll probably do it, just not today.

pompon0: not random g2, I mean random slice of bytes

moshababo: @brunoffranca reading the code you referenced, it always creates a point from a random `x` because a loop is used. As mentioned in https://github.com/matter-labs/era-consensus/pull/80, it's about 50% chance.

node/libs/crypto/src/bn254/mod.rs
moshababo, line 144: Why not verify prior to `hash_to_g1`?

moshababo, line 144: Why not verify prior to `hash_to_g1()`?

brunoffranca:
So, there was a fatal flaw in the subgroup check. To check that a given point is in the correct subgroup, we don't multiply by the cofactor, we multiply by the order. I fixed that and found a way of producing testing it appropriately.
Another bug is that we were not doing the validity checks for aggregate signature verification. Also fixed it.
Finally, refactored it a bit to make it more readable.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -96,10 +114,14 @@ impl Hash for PublicKey {
...
         let arr: [u8; 64] = bytes.try_into()?;
-        let p = G2Compressed::from_fixed_bytes(arr)
-            .into_affine()?
-            .into_projective();
-        Ok(PublicKey(p))
+        let p = G2Compressed::from_fixed_bytes(arr).into_affine()?;
+        let pk = PublicKey(p.into());
+
+        if !pk.is_valid() {
+            anyhow::bail!("Public key is not valid")
```
pompon0, line 121: anyhow::ensure!

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -81,6 +81,24 @@ impl PartialEq for SecretKey {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PublicKey(G2);
 
+impl PublicKey {
+    /// Checks if the public key is not the identity element and is in the correct subgroup. Verifying signatures
+    /// against public keys that are not valid is insecure.
+    pub fn is_valid(&self) -> bool {
```
pompon0, line 87: it can be private, no?

pompon0: nit: more canonical would be to make it `fn verify(&self) -> Result<()>`, rather than return a bool.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -81,6 +81,24 @@ impl PartialEq for SecretKey {
...
+        let order = Fr::char();
+        let mut p = self.0;
+        p.mul_assign(order);
+        p.is_zero()
+    }
+}
+
 impl Default for PublicKey {
     fn default() -> Self {
         PublicKey(G2::zero())
```
pompon0, line 104: It is problematic that we allow to construct an invalid public key. We should either remove the default constructor, or keep the is_valid/verify public and revert my suggestion of using `assert!` instead of returning an error.

node/libs/crypto/src/bn254/mod.rs
pompon0: I think you should also add a check to SecretKey::decode(), because SecretKey::public() also constructs a PublicKey

node/libs/crypto/src/bn254/mod.rs
pompon0: Ok, I've tried to remove the default constructor and it compiles just fine with minimal changes to AggregateSignature::verify.

node/libs/crypto/src/bn254/mod.rs
brunoffranca: It can be private, but I think it's better to keep it public. It's a useful method.

pompon0: I mean, all PublicKey objects ever constructed are expected to be valid, so having verify() in the public API will be quite misleading...

PR #90, opened 2024-04-10
Fixing more crypto issues

brunoffranca:
## What ❔
Fixing more crypto issues found in the security review. Namely,


node/libs/crypto/src/bls12_381/mod.rs
```diff
@@ -34,7 +47,14 @@ impl SecretKey {
 
     /// Produces a signature using this [`SecretKey`]
     pub fn sign(&self, msg: &[u8]) -> Signature {
-        Signature(self.0.sign(msg, &[], &[]))
+        Signature(self.0.sign(msg, DST, &[]))
+    }
+
+    /// Produces a proof of possession using this [`SecretKey`]
```
dnkolegov, line 53:
```suggestion
    /// Produces a proof of possession for the public key corresponding to this [`SecretKey`]
```

node/libs/crypto/src/bls12_381/mod.rs
```diff
@@ -23,6 +35,7 @@ pub const INFINITY_PUBLIC_KEY: [u8; PUBLIC_KEY_BYTES_LEN] = [
 ];
 
 /// Type safety wrapper around a `blst` SecretKey
+#[derive(ZeroizeOnDrop)]
```
dnkolegov, line 38: It seems that blst [zeroizes](https://github.com/supranational/blst/blob/master/bindings/rust/src/lib.rs#L550) its `SecretKey`, but let's keep it here too to know that we zeroize the key

node/libs/crypto/src/ed25519/mod.rs
```diff
@@ -3,15 +3,17 @@
...
 use ed25519_dalek as ed;
+use zeroize::ZeroizeOnDrop;
 
 pub mod testonly;
 
 #[cfg(test)]
 mod tests;
 
 /// ed25519 secret key.
+#[derive(ZeroizeOnDrop)]
```
dnkolegov, line 16: It is [enabled](https://github.com/dalek-cryptography/ed25519-dalek?tab=readme-ov-file#feature-flags) by default, but let's keep it here

dnkolegov (approved):
Thank you @brunoffranca.
LGTM!

One note: 
`CP-1-BLS-2` was about the zeroization of `key_material` removed from the `generate` function earlier. I will update the report.

PR #93, opened 2024-04-11
New attester role for full nodes

IAvecilla:
## What ❔

This PR introduces a new role called `attester`.

## Why ❔

This role will handle various keys and messages associated with L1 batches. In the future, this will enable the use of a new signature scheme specifically for L1 batch signatures, distinguishing them from the block signatures of validators.


node/libs/roles/src/proto/validator.proto
```diff
@@ -12,6 +12,7 @@ message Fork {
 message Genesis {
   optional Fork fork = 1; // required
   repeated PublicKey validators = 2;
+  repeated BatchPublicKey attesters = 3;
```
IAvecilla, line 15: @pompon0 @brunoffranca This is what I'm trying to do in the `Genesis` message to add keys from the attesters but it's failing

node/libs/roles/src/attester/keys/batch_public_key.rs
```diff
@@ -0,0 +1,37 @@
+use std::fmt;
+use zksync_consensus_crypto::{bn254, ByteFmt, Text, TextFmt};
+
+use crate::validator::PublicKey;
+
+/// A public key for a validator used in L1 batch signing.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BatchPublicKey(pub(crate) bn254::PublicKey);
+
+impl ByteFmt for BatchPublicKey {
```
pompon0, line 10: please remove the "Batch" prefix from PublicKey, SecretKey, Signature. The fact that attester key can be used to sign L1Batches is the property of the attester message.

node/libs/roles/src/proto/validator.proto
pompon0: you need to add `import zksync/roles/attester.proto` and use relative path `attester.BatchPublicKey`.

node/libs/roles/src/proto/attester.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package zksync.roles.attester;
+
+import "zksync/std.proto";
+
+message L1BatchMsg {
```
pompon0, line 7: "Msg" suffix is redundant here

node/libs/roles/src/attester/keys/batch_public_key.rs
brunoffranca: +1

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// The public key of the signer.
+    pub key: BatchPublicKey,
+    /// The signature.
+    pub sig: BatchSignature,
+}
+
+/// Struct that represents a bit map of validators. We use it to compactly store
+/// which validators signed a given message.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Attestations(pub BitVec);
```
pompon0, line 53: rename to "Signers"

node/libs/roles/src/proto/attester.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package zksync.roles.attester;
+
+import "zksync/std.proto";
```
pompon0, line 5: remove if your are not using anything from std.proto

brunoffranca (commented):
Overall, the direction looks good to me. :+1:

node/libs/roles/src/proto/attester.proto
pompon0: it is still here :)

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,234 @@
...
+        faulty_replicas(self.len())
+    }
+}
+
+/// Calculate the maximum number of faulty replicas, for a given number of replicas.
+pub fn faulty_replicas(n: usize) -> usize {
+    // Calculate the allowed maximum number of faulty replicas. We want the following relationship to hold:
+    //      n = 5*f + 1
+    // for n total replicas and f faulty replicas. This results in the following formula for the maximum
+    // number of faulty replicas:
```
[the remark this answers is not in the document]
pompon0: also, attesters will have weights too, no?

node/libs/roles/src/proto/attester.proto
IAvecilla: Sorry, fixed!

node/libs/roles/src/attester/messages/msg.rs
```diff
@@ -0,0 +1,234 @@
...
+        Ok(this)
+    }
+}
+
+/// Strongly typed signed l1 batch message.
+/// WARNING: signature is not guaranteed to be valid.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SignedBatchMsg {
+    /// The message that was signed.
+    pub msg: L1Batch,
```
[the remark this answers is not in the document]
IAvecilla: Done!

IAvecilla:
@pompon0 @brunoffranca The CI is failing due to the incompatibility with the protobuf messages, but if this is merged into the `signature_broadcast` branch, can't these issues be resolved before merging it into the main branch?

brunoffranca:
Yes, that's ok. The CI is only enforced when merging into main anyway.

node/libs/roles/src/attester/messages/msg.rs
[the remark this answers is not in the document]
brunoffranca: Majority is technically enough, but given that our validator and attester sets are the same, having different thresholds would just decrease our security (against liveness attacks).
They have weights, yes. But Esteban is still working on that PR, so maybe we can ignore it for now?

node/libs/roles/src/attester/messages/l1_batch.rs
```diff
@@ -1,22 +1,25 @@
...
 
-use super::{Genesis, Signed};
+use super::{SignedBatchMsg, Signers};
 
 /// A message to send by validators to the gossip network.
 /// It contains the validators signature to sign the block batches to be sent to L1.
-#[derive(Debug, Clone, Eq, PartialEq, Default)]
-pub struct L1BatchMsg;
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default)]
+pub struct L1Batch;
```
brunoffranca, line 11: I think we can already make this contain a Vec<u8>. That's what a L1Batch will be most likely.

brunoffranca (approved):
Looks good. Just remember that you'll need to merge Esteban's PR later and change the attester to use weights.

node/libs/roles/src/attester/messages/l1_batch.rs
pompon0: I thought that we want to include just a hash/other digest of the L1Batch of some form, no? (it doesn't change the fact that we will keep just a slice of bytes anyway, and let the user define the content). Also we will need the L1Batch number.

pompon0 (approved):
@IAvecilla not only protobuf ci is failing but also the unit tests are failing. Still it doesn't matter much as you are merging stuff between your own branches anyway.

node/libs/roles/src/attester/messages/l1_batch.rs
brunoffranca: True, we need the number. Would you prefer that we strongly type the hash? I guess if we have multiple fields then it makes sense. With just one field it looks weird.

brunoffranca: Given that we'll also have batch syncing later on, maybe we should already name this L1BatchHeader?

PR #96, opened 2024-04-15
Support multiple leader-selection modes (BFT-416)

moshababo:
## What ❔

Add support for multiple leader selection modes: 
1. **Round robin**: based on validator's index, non-weighted. 
2. **Sticky**: constant leader. 
3. **Weighted**: pseudo-random, weighted. 

## Why ❔

To ease protocol upgrades.

---

### Notes
* Selection mode is considered as part of the genesis/network config.
* Weighted selection calculation can be optimized, but I'm not sure it's needed.

### TODOs
* Wait for https://github.com/matter-labs/era-consensus/pull/77 to be merged.
* Consider decouple `view_leader()` code and perhaps all of `Committee` struct from `/messages/consensus.rs` since it's not really related to messaging.
* Add external config support.
* Add tests.
* Add docs.

### DIFF
--- a/node/libs/roles/src/validator/conv.rs
+++ b/node/libs/roles/src/validator/conv.rs
@@ -4,7 +4,7 @@ use super::{
     MsgHash, NetAddress, Payload, PayloadHash, Phase, PrepareQC, ProtocolVersion, PublicKey,
     ReplicaCommit, ReplicaPrepare, Signature, Signed, Signers, View, ViewNumber, WeightedValidator,
 };
-use crate::{node::SessionId, proto::validator as proto};
+use crate::{node::SessionId, proto::validator as proto, validator::LeaderSelectionMode};
 use anyhow::Context as _;
 use std::collections::BTreeMap;
 use zksync_consensus_crypto::ByteFmt;
@@ -63,7 +63,8 @@ impl ProtoFmt for Genesis {
             };
         Ok(Self {
             fork: read_required(&r.fork).context("fork")?,
-            validators: Committee::new(validators.into_iter()).context("validators")?,
+            validators: Committee::new(validators.into_iter(), LeaderSelectionMode::RoundRobin)
+                .context("validators")?,
             version,
         })
     }
--- a/node/libs/roles/src/validator/messages/consensus.rs
+++ b/node/libs/roles/src/validator/messages/consensus.rs
@@ -3,7 +3,7 @@ use super::{BlockNumber, LeaderCommit, LeaderPrepare, Msg, ReplicaCommit, Replic
 use crate::validator;
 use anyhow::Context;
 use bit_vec::BitVec;
-use std::{collections::BTreeMap, fmt};
+use std::{collections::BTreeMap, fmt, hash::Hash};
 use zksync_consensus_crypto::{keccak256::Keccak256, ByteFmt, Text, TextFmt};
 use zksync_consensus_utils::enum_util::{BadVariantError, Variant};
 
@@ -70,6 +70,22 @@ impl Default for Fork {
     }
 }
 
+#[derive(Default, Clone, Debug, Eq, PartialEq)]
+pub enum LeaderSelectionMode {
+    #[default]
+    RoundRobin,
+    Sticky(validator::PublicKey),
+    Weighted,
+}
+
+pub fn weighted_eligibility(input: u64, total_weight: u64) -> u64 {
+    let input_bytes = input.to_be_bytes();
+    let hash = Keccak256::new(&input_bytes);
+    let hash_64 = &hash.as_bytes()[0..8];
+    let hash_num = u64::from_be_bytes(hash_64.try_into().unwrap());
+    hash_num % total_weight
+}
+
 /// A struct that represents a set of validators. It is used to store the current validator set.
 /// We represent each validator by its validator public key.
 #[derive(Clone, Debug, PartialEq, Eq, Default)]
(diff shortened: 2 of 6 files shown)


pompon0 (commented):
Leader selection algorithm should be encoded in the proto. Also remember to maintain backward compatibility of the genesis encoding. Just a thought: wouldn't it be more convenient to generalize both strategies to support arbitrary subset of comittee as set of proposers?

moshababo:
> Just a thought: wouldn't it be more convenient to generalize both strategies to support arbitrary subset of comittee as set of proposers?

Can you elaborate more? it sounds to me like an additional feature, not a reduction.

moshababo:
> Also remember to maintain backward compatibility of the genesis encoding. 

I've added an additional field with a new number to `proto::Genesis`, so that should be fine. 
However, adding the field to the Rust type alters the hash for `genesis_v0` / `genesis_v1`, causing the tests to fail. Do we need to maintain support for the current hardcoded hash values?

pompon0:
> However, adding the field to the Rust type alters the hash for genesis_v0 / genesis_v1, causing the tests to fail. Do we need to maintain support for the current hardcoded hash values?

Yes, that's the whole point of hashes. You need to maintain the genesis_v0 hash, genesis_v1 is not stabilized yet (i.e. not deployed anywhere), so this one can be updated.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -70,6 +70,29 @@ impl Default for Fork {
...
+    Weighted,
+}
+
+/// Calculates the pseudo-random eligibility of a leader based on the input and total weight.
+pub(crate) fn leader_weighted_eligibility(input: u64, total_weight: u64) -> u64 {
+    let input_bytes = input.to_be_bytes();
+    let hash = Keccak256::new(&input_bytes);
+    let hash_64 = &hash.as_bytes()[0..8];
+    let hash_num = u64::from_be_bytes(hash_64.try_into().unwrap());
+    hash_num % total_weight
```
pompon0, line 93: this doesn't generate a uniform distribution for large weights

pompon0:
> Can you elaborate more? it sounds to me like an additional feature, not a reduction.

I mean if you have a round robin/weighted strategy over a subset of replicas, then RoundRobin/Weighted is a special case where the subset if the set of all replicas and Sticky i a special case where the the subest is a singleton.

Now that you have added both RoundRobin and weighted it gets a bit more complex. IMO RoundRobin should be the only supported variant for the genesis_v0, and weighted/sticky should be the only supported variants for genesis_v1.

node/libs/roles/src/validator/messages/consensus.rs
moshababo: For `total_weight` which is `u64::MAX`, it will return the value of the hash first 8 bytes. Is that a problem?
Other option is to convert both values to big uint, perform the modulo, and then convert back to a single u64, knowing it should be less then that.

brunoffranca:
I agree with Moshe, it seems like an extra feature. We don't need this much generalization in my opinion. Especially since this will be a configuration option in the end, so we need it to be understandable for users.
Let's keep with the original plan and have those 3 modes.

node/libs/roles/src/validator/messages/consensus.rs
pompon0: I mean that if you have 2 validators each with weight u64::MAX/3, then the first validator will get 2/3 of the blocks.

node/libs/roles/src/validator/conv.rs
```diff
@@ -65,7 +65,12 @@ impl ProtoFmt for Genesis {
             fork: read_required(&r.fork).context("fork")?,
             validators: Committee::new(validators.into_iter()).context("validators")?,
             version,
-            leader_selection: read_required(&r.leader_selection).context("leader_selection")?,
+            leader_selection: r
+                .leader_selection
+                .as_ref()
+                .map(ProtoFmt::read)
```
pompon0, line 71: read_optional

node/libs/roles/src/validator/conv.rs
```diff
@@ -57,14 +57,20 @@ impl ProtoFmt for Genesis {
...
+                // empty validator set, Committee:new() will later return an error.
             } else {
                 (vec![], GenesisVersion::CURRENT)
             };
         Ok(Self {
             fork: read_required(&r.fork).context("fork")?,
             validators: Committee::new(validators.into_iter()).context("validators")?,
             version,
+            leader_selection: r
+                .leader_selection
```
pompon0, line 69: please verify that Sticky.key actually belongs to the committee

pompon0: ...and when the bft actor is starting as well.

node/libs/roles/src/validator/messages/consensus.rs
pompon0: i.e. IMO we should make the distribution error significantly smaller for every possible inputs.

node/libs/roles/src/validator/conv.rs
moshababo: Fixed.

node/libs/roles/src/validator/conv.rs
moshababo: Fixed here: https://github.com/matter-labs/era-consensus/pull/96/commits/5bd14b38ff3c7d5289800af4d9bf906712d74d5c

node/libs/roles/src/validator/messages/consensus.rs
moshababo: Thanks for the clarification. The solution I suggested was implemented here: https://github.com/matter-labs/era-consensus/pull/96/commits/feabe397ef86e49db3ca249bc2d957080fd446e3
I assume that using the hash output's entire 256 bits to reduce non-uniformity is sufficient. LMK if I'm still missing something.

pompon0: LG, as far as I'm concerned u128 would also be good enough, but u256 is good as well.

pompon0 (approved):

pompon0:
FYI I've added some comments on the commit, but for some reason they are not visible in the main conversation.

PR #97, opened 2024-04-16
Check sk < r for bn254

dnkolegov:
## What ❔

This PR adds a check that generated or decoded secret key is in [1, r) range.

## Why ❔

To satisfy [BLS](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04) RFC draft in that sense.

### DIFF
--- a/node/libs/crypto/src/bn254/mod.rs
+++ b/node/libs/crypto/src/bn254/mod.rs
@@ -30,11 +30,12 @@ pub struct SecretKey(Fr);
 
 impl SecretKey {
     /// Generates a secret key from a cryptographically-secure entropy source.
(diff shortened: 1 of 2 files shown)


brunoffranca:
This seems unnecessary. The pairing crate should only create Fr values that are smaller than r, by construction. I'll double check this.
Anyway, I wouldn't spend any time fixing the BN curve since it's likely we will abandon it.

brunoffranca (approved):

brunoffranca:
The rand implementation is somewhere in the ff crate, I can't find it now. Let's just merge it, an extra check can't hurt.

PR #100, opened 2024-04-19
Tests and prevention for several potential attack types (BFT-419)

ElFantasma:
## What ❔
Prevents OOM attack where a malicious replica spamming `ReplicaPrepare` messages for future views could grow the number of PrepareQC stored in memory by cleaning PrepareQCs from views without other replicas references. This was done for CommitQC as well.

The ticket specifies 3 other types of attack, but number 2 and 3 seem no longer possible, as message signature and order are checked before adding the message to the prepare/commit quorum certificate aggregation maps.

For attack number 4 we now consider a message for a past view as invalid and we are discarding it, so messages out of order are not recoverable. 

## Why ❔
This PR is intended to address several potential attack types described in [BFT-419 - couple of attacks with 1 malicious replica](https://linear.app/matterlabs/issue/BFT-419/couple-of-attacks-with-1-malicious-replica)


node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -128,8 +129,24 @@ impl StateMachine {
...
 
-        // Now we check if we have enough weight to continue.
+        // Calculate the CommitQC signers weight.
         let weight = self.config.genesis().validators.weight(&commit_qc.signers);
+
+        // Update commit message current view number for author
+        self.replica_commit_views
+            .insert(author.clone(), message.view.number);
+
+        // Clean up commit_qcs for the case that no replica is at the view
```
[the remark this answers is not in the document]
pompon0: nit: you can optimize it by iterating only over the signers of the given commit_qc (O(n^2) -> O(n))

node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -38,9 +38,9 @@ pub(crate) enum Error {
     /// Invalid message.
     #[error("invalid message: {0:#}")]
     InvalidMessage(#[source] anyhow::Error),
-    /// Duplicate message from a replica.
-    #[error("Replica signed more than one message for same view (message: {message:?}")]
-    DuplicateSignature {
+    /// A message was already sent for a replica at the same view.
```
pompon0, line 41: IMO we can treat this as Old message.

node/actors/bft/src/leader/tests.rs
```diff
@@ -395,6 +395,38 @@ async fn replica_prepare_different_messages() {
     .unwrap();
 }
 
+/// Check that leader won't accumulate undefined amount of messages if
```
pompon0, line 398: <3

pompon0:
2,3,4 are still possible because pruning of messages happens in prunable_mpsc, i.e. even before it reaches bft actor.

brunoffranca (approved):
Good work. Approved!

node/actors/bft/src/leader/state_machine.rs
```diff
@@ -207,15 +216,39 @@ impl StateMachine {
...
+    fn inbound_filter_function(new_req: &ConsensusReq) -> SelectionFunctionResult {
+        // Verify message signature
+        match new_req.msg.verify() {
+            Ok(_) => SelectionFunctionResult::Keep,
+            _ => SelectionFunctionResult::DiscardNew,
+        }
+    }
+
+    fn inbound_selection_function(
+        pending_req: &ConsensusReq,
```
pompon0, line 228: 'old_req'?

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
```diff
@@ -14,15 +14,30 @@ use std::{collections::VecDeque, fmt, sync::Arc};
...
+}
+
 /// Creates a channel and returns the [`Sender`] and [`Receiver`] handles.
 /// All values sent on [`Sender`] will become available on [`Receiver`] in the same order as it was sent,
 /// unless will be pruned before received.
 ///
 /// * [`T`]: The type of data that will be sent through the channel.
-/// * [`pruning_predicate`]: A function that determines whether an unreceived, pending value in the buffer (represented by the first `T`) should be pruned
-/// based on a newly sent value (represented by the second `T`).
+/// * [`filter_function`]: A function that checks the newly sent value and avoids adding to the queue if it returns false
```
pompon0, line 33: comment outdated

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
```diff
@@ -51,17 +67,65 @@ struct Shared<T> {
...
+                        break;
+                    }
+                    SelectionFunctionResult::DiscardNew => return,
+                }
+            }
+            // Stage 2: Swap retained value into current idx.
+            while cur < len {
+                match (self.selection_function)(&mut buf[cur], &value) {
+                    SelectionFunctionResult::Keep => {
+                        buf.swap(idx, cur);
```
pompon0, line 106: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.swap idx and cur can be equal. Duplicating the loop is not needed.

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
pompon0: nit: actually DiscardOld is not meaningful for filter_function, so perhaps bool was a better choice there?

pompon0 (approved):

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
```diff
@@ -51,17 +67,65 @@ struct Shared<T> {
...
+        // Check if new value is valid to keep
+        if (self.filter_function)(&value) != SelectionFunctionResult::Keep {
+            return;
+        }
         self.shared.send.send_modify(|buf| {
-            buf.retain(|pending_value| !(self.pruning_predicate)(pending_value, &value));
-            buf.push_back(value);
+            let len = buf.len();
+            let mut idx = 0;
+            let mut cur = 0;
```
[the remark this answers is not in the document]
ElFantasma: Same here, did the short circuit version for efficiency. But this one is cleaner, I'll use this one if you are ok with it. 👍

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
ElFantasma: Yeah, that was my first version, but I found the code clearer defining both functions returning a `SelectionFunctionResult`  (like [here](https://github.com/matter-labs/era-consensus/blob/5d200d81073708f06a4a5911ab2c48c12739c103/node/actors/bft/src/replica/state_machine.rs#L173-L209))

ElFantasma: I debated myself on what version is better. I'm open for suggestions.

pompon0: well, you can have 2 enums.

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
pompon0: I'm not sure I follow. If we want this to be efficient, it would be better to have the queue group messages by sender, so that you can lookup them directly when adding a new message, instead of iterating over the whole thing. If we do O(n) insertion, then simplest implementation is a way to go.

ElFantasma: Yes, it's just a nitpick (to avoid 'same index' swaps). If we group by sender we wouldn't have a generic data structure. I just adapted the [VeqDeque](https://github.com/rust-lang/rust/blob/master/library/alloc/src/collections/vec_deque/mod.rs#L2057-L2089) implementation to have at least same performance as before. 
But already changed the implementation to your suggestion.

node/libs/concurrency/src/sync/prunable_mpsc/mod.rs
ElFantasma: Going for the boolean version. It's cleaner to define a predicate.

PR #101, opened 2024-04-23
Rewritten sync_blocks actor to maximize utilization.

pompon0:
Previous implementation severely limited the total throughput.


brunoffranca:
So we no longer have a sync actor, everything is in the network crate. Is this going to work well when we need to sync other data, like L1 batches, transactions, attester signatures, etc?

pompon0:
yes, nothing changes in that respect, the same amount of work is needed to support fetching other primitives.

brunoffranca (approved):

brunoffranca:
I still need to update the docs. Don't merge yet.

PR #103, opened 2024-04-24
Change validator key to BLS12-381

brunoffranca:
## What ❔
Changes the validator keys to use the BLS12-381 curve.

## Why ❔
Now that we decided to separate validator and attester keys, we can use a more efficient curve for the validator keys.

### DIFF
--- a/node/libs/crypto/src/bls12_381/mod.rs
+++ b/node/libs/crypto/src/bls12_381/mod.rs
@@ -71,7 +71,7 @@ impl ByteFmt for SecretKey {
     fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
         bls::SecretKey::from_bytes(bytes)
             .map(Self)
-            .map_err(|e| anyhow!("Failed to decode secret key: {e:?}"))
+            .map_err(|e| anyhow::format_err!("Failed to decode secret key: {e:?}"))
     }
 
     fn encode(&self) -> Vec<u8> {
@@ -106,7 +106,7 @@ impl ByteFmt for PublicKey {
     fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
         bls::PublicKey::key_validate(bytes)
             .map(Self)
-            .map_err(|e| anyhow!("Failed to decode public key: {e:?}"))
+            .map_err(|e| anyhow::format_err!("Failed to decode public key: {e:?}"))
     }
 
     fn encode(&self) -> Vec<u8> {
@@ -137,7 +137,9 @@ impl Signature {
 
         match result {
             BLST_ERROR::BLST_SUCCESS => Ok(()),
-            err => Err(anyhow!("Signature verification failure: {err:?}")),
+            err => Err(anyhow::format_err!(
+                "Signature verification failure: {err:?}"
+            )),
         }
     }
 }
@@ -147,7 +149,7 @@ impl ByteFmt for Signature {
     fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
         bls::Signature::sig_validate(bytes, false)
             .map(Self)
-            .map_err(|err| anyhow!("Error decoding signature: {err:?}"))
+            .map_err(|err| anyhow::format_err!("Error decoding signature: {err:?}"))
     }
 
     fn encode(&self) -> Vec<u8> {
@@ -208,7 +210,9 @@ impl AggregateSignature {
         for (msg, pk) in msgs_and_pks {
             if let Some(existing_pk) = tree_map.get_mut(msg) {
                 if let Err(err) = existing_pk.add_public_key(&pk.0, false) {
-                    return Err(anyhow!("Error aggregating public keys: {err:?}"));
+                    return Err(anyhow::format_err!(
+                        "Error aggregating public keys: {err:?}"
+                    ));
                 }
             } else {
                 tree_map.insert(msg, bls::AggregatePublicKey::from_public_key(&pk.0));
@@ -231,7 +235,9 @@ impl AggregateSignature {
 
         match result {
             BLST_ERROR::BLST_SUCCESS => Ok(()),
-            err => Err(anyhow!("Aggregate signature verification failure: {err:?}")),
+            err => Err(anyhow::format_err!(
+                "Aggregate signature verification failure: {err:?}"
+            )),
         }
     }
 }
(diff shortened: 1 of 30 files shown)


dnkolegov:
@brunoffranca, does that mean that the bn254 curve will not be used at all?

brunoffranca:
It will be used to sign L1 batches. L2 blocks and all consensus messages will use BLS12-381.
After EIP-2537 gets merged, everything will use BLS12-381.

node/libs/crypto/src/bls12_381/mod.rs
```diff
@@ -119,19 +132,20 @@ pub struct Signature(bls::Signature);
...
 impl Signature {
     /// Verifies a signature against the provided public key
-    pub fn verify(&self, msg: &[u8], pk: &PublicKey) -> Result<(), Error> {
+    pub fn verify(&self, msg: &[u8], pk: &PublicKey) -> anyhow::Result<()> {
         let result = self.0.verify(true, msg, DST, &[], &pk.0, true);
 
         match result {
             BLST_ERROR::BLST_SUCCESS => Ok(()),
-            err => Err(Error::SignatureVerification(err)),
+            err => Err(anyhow!("Signature verification failure: {err:?}")),
```
pompon0, line 140: `anyhow::format_err!`

node/libs/crypto/src/bls12_381/mod.rs
```diff
@@ -153,22 +167,31 @@ impl Ord for Signature {
...
+#[derive(Clone, Debug)]
+pub struct AggregateSignature(bls::AggregateSignature);
 
-impl AggregateSignature {
-    /// Generates an aggregate signature from a list of signatures
-    pub fn aggregate<'a>(sigs: impl IntoIterator<Item = &'a Signature>) -> Result<Self, Error> {
-        let sigs: Vec<&bls::Signature> = sigs.into_iter().map(|s| &s.0).collect();
+impl Default for AggregateSignature {
+    fn default() -> Self {
+        // This can't fail in production since we are decoding a known value.
```
pompon0, line 176: I personally like the fact that we can have an empty signature, but isn't it against some RFC or sth?

pompon0 (approved):

node/libs/crypto/src/bls12_381/mod.rs
brunoffranca: I don't think so. Technically the identity point is a valid signature, but I'll double check it.

dnkolegov: What does "empty signature" mean from cryptography perspective?
It cannot be empty. Since aggregation operation must have at least one signature as input.

brunoffranca: It's the identity point. Which in here corresponds to the Default aggregate signature.

node/libs/crypto/src/bls12_381/mod.rs
brunoffranca: It seems to be just a reexport of `anyhow!` ([https://docs.rs/anyhow/latest/anyhow/index.html#reexports](https://docs.rs/anyhow/latest/anyhow/index.html#reexports)). Do you still think it makes sense to change it?

brunoffranca:
@pompon0 I added two more commits removing `MsgHash`. It doesn't make sense to be hashing twice. See if you approve.

pompon0:
@brunoffranca the point of having MsgHash was to avoid passing the whole message from place to place, and to avoid cloning as the messages can get quite large. I agree that so far we have underutilized those optimizations, but I can already see that you had to add a couple of extra clones in this PR already.

node/libs/roles/src/node/messages.rs
```diff
@@ -14,10 +14,13 @@ pub enum Msg {
...
 
-impl Msg {
-    /// Get the hash of this message.
-    pub fn hash(&self) -> MsgHash {
-        MsgHash(keccak256::Keccak256::new(&zksync_protobuf::canonical(self)))
+impl ByteFmt for Msg {
+    fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
+        zksync_protobuf::decode(bytes)
+    }
+
```
pompon0, line 21: don't expose the bytefmt of the message. Keep it an implementation detail of signing.

pompon0: also don't define the decode

node/libs/crypto/src/bls12_381/mod.rs
```diff
@@ -177,15 +200,15 @@ impl AggregateSignature {
     pub fn verify<'a>(
         &self,
         msgs_and_pks: impl Iterator<Item = (&'a [u8], &'a PublicKey)>,
```
pompon0, line 202: now the messages that are passed can be large and therefore indexing data structures like BTreeMap with messages will be inefficient.

node/libs/crypto/src/bn254/mod.rs
```diff
@@ -227,6 +227,15 @@ impl AggregateSignature {
...
+    pub fn aggregate<'a>(sigs: impl IntoIterator<Item = &'a Signature>) -> Self {
+        let mut agg = Self::default();
+        for sig in sigs {
+            agg.add(sig);
+        }
+        agg
+    }
+
     /// This function is intentionally non-generic and disallow inlining to ensure that compilation optimizations can be effectively applied.
     /// This optimization is needed for ensuring that tests can run within a reasonable time frame.
```
pompon0, line 240: same as in bls: this implementation of verify() is inefficient with long messages.

node/libs/crypto/src/bls12_381/mod.rs
pompon0: or maybe it is not that bad, nvm

pompon0:
> Do you believe that this extra cloning does make a performance difference? And that we will actually start utilizing this MsgHash construct more in the future? If yes, I can remove these two commits.

It is hard for me to tell. We can reintroduce it later, once there is a need to do some serious profiling. Just please avoid introducing `Clone` requirements in signatures, and please remove MsgHash from node role as well for consistency (for node role it definitely doesn't matter).

pompon0 (changes requested):
added some comments

pompon0:
also FYI, if we reintroduce it later it will be a breaking change to the protocol.

brunoffranca:
Ok, like I always say, tie break goes to the reviewer. So I'll remove these two commits.

PR #112, opened 2024-04-29
Return Result<(), Error> on QC add 

ElFantasma:
## What ❔

Refactored the code to return errors on `PrepareQC::add` and `CommitQC::add`. We were just ignoring messages from different views, or signed by replicas not in the committee, or signed twice. Now these scenarios return an error.

## Why ❔

From a code review it was suggested to return errors in `PrepareQC::add` and `CommitQC::add` instead of ignoring the message. FWIW, these cases are already been checked by our current caller, but for correctness we believe these functions should provide proper errors in case addition to the QC is not performed.

### DIFF
--- a/node/actors/bft/src/leader/replica_commit.rs
+++ b/node/actors/bft/src/leader/replica_commit.rs
@@ -105,6 +105,7 @@ impl StateMachine {
         // ----------- All checks finished. Now we process the message. --------------
 
         // Add the message to the QC.
+        // Should always succeed as all checks have been already performed
         commit_qc
             .add(&signed_message, self.config.genesis())
             .expect("Could not add message to CommitQC");
--- a/node/actors/bft/src/leader/replica_commit.rs
+++ b/node/actors/bft/src/leader/replica_commit.rs
@@ -105,7 +105,9 @@ impl StateMachine {
         // ----------- All checks finished. Now we process the message. --------------
 
         // Add the message to the QC.
-        commit_qc.add(&signed_message, self.config.genesis());
+        commit_qc
+            .add(&signed_message, self.config.genesis())
+            .expect("Could not add message to CommitQC");
 
         // Now we check if we have enough weight to continue.
         let weight = self.config.genesis().committee.weight(&commit_qc.signers);
--- a/node/actors/bft/src/leader/replica_prepare.rs
+++ b/node/actors/bft/src/leader/replica_prepare.rs
@@ -115,7 +115,9 @@ impl StateMachine {
             .prepare_qcs
             .entry(message.view.number)
             .or_insert_with(|| validator::PrepareQC::new(message.view.clone()));
-        prepare_qc.add(&signed_message, self.config.genesis());
+        prepare_qc
+            .add(&signed_message, self.config.genesis())
+            .expect("Could not add message to PrepareQC");
 
         // We store the message in our cache.
         self.prepare_message_cache
--- a/node/actors/bft/src/leader/replica_prepare.rs
+++ b/node/actors/bft/src/leader/replica_prepare.rs
@@ -115,6 +115,8 @@ impl StateMachine {
             .prepare_qcs
             .entry(message.view.number)
             .or_insert_with(|| validator::PrepareQC::new(message.view.clone()));
+
+        // Should always succeed as all checks have been already performed
         prepare_qc
             .add(&signed_message, self.config.genesis())
             .expect("Could not add message to PrepareQC");
--- a/node/libs/roles/src/validator/messages/leader_commit.rs
+++ b/node/libs/roles/src/validator/messages/leader_commit.rs
@@ -58,7 +58,7 @@ pub enum CommitQCVerifyError {
 #[derive(thiserror::Error, Debug)]
 pub enum CommitQCAddError {
     /// Inconsistent views.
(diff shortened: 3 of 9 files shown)


node/actors/bft/src/leader/replica_commit.rs
```diff
@@ -105,7 +105,9 @@ impl StateMachine {
         // ----------- All checks finished. Now we process the message. --------------
 
         // Add the message to the QC.
-        commit_qc.add(&signed_message, self.config.genesis());
```
pompon0, line 108: add a comment that this should always succeed, because the check was already performed.

node/actors/bft/src/leader/replica_prepare.rs
```diff
@@ -115,7 +115,9 @@ impl StateMachine {
             .prepare_qcs
             .entry(message.view.number)
             .or_insert_with(|| validator::PrepareQC::new(message.view.clone()));
-        prepare_qc.add(&signed_message, self.config.genesis());
+        prepare_qc
```
pompon0, line 118: ditto

node/libs/roles/src/validator/messages/leader_commit.rs
```diff
@@ -54,6 +54,26 @@ pub enum CommitQCVerifyError {
...
+#[derive(thiserror::Error, Debug)]
+pub enum CommitQCAddError {
+    /// Inconsistent views.
+    #[error("Trying to add a message from a different view")]
+    InconsistentViews,
+    /// Signer not present in the committee.
+    #[error("Signer not in committee: {signer:?}")]
+    SignerNotInCommittee {
+        /// Signer of the message.
+        signer: Box<validator::PublicKey>,
```
pompon0, line 67: nit: key is not that large. Any specific reason to box it?

node/libs/roles/src/validator/messages/leader_commit.rs
```diff
@@ -54,6 +54,26 @@ pub enum CommitQCVerifyError {
     BadSignature(#[source] anyhow::Error),
 }
 
+/// Error returned by `CommitQC::add()`.
+#[derive(thiserror::Error, Debug)]
+pub enum CommitQCAddError {
```
pompon0, line 59: kudos for adding strong error types here. How about adding a small test for those methods while you are at it?

node/libs/roles/src/validator/messages/leader_commit.rs
pompon0: like, it is fine as is, I'm just curious.

node/libs/roles/src/validator/messages/leader_prepare.rs
```diff
@@ -46,6 +46,26 @@ pub enum PrepareQCVerifyError {
...
+    #[error("Signer not in committee: {signer:?}")]
+    SignerNotInCommittee {
+        /// Signer of the message.
+        signer: Box<validator::PublicKey>,
+    },
+    /// Message already present in PrepareQC.
+    #[error("Message already signed for PrepareQC: {existing_message:?}")]
+    Exists {
+        /// Existing message from the same replica.
+        existing_message: Box<ReplicaPrepare>,
```
pompon0, line 65: this is kind of redundant, since it is always a copy of the input msg :P

node/libs/roles/src/validator/messages/leader_prepare.rs
```diff
@@ -80,24 +100,33 @@ impl PrepareQC {
 
     /// Add a validator's signed message.
     /// Message is assumed to be already verified.
-    // TODO: check if there is already a message from that validator.
```
pompon0, line 83: oh, I now remember that this TODO is about: we just check here that we don't have a signature from this validator on msg, but since PrepareQC collects signatures for various messages, we should check that we don't have a signature from this validator on ANY msg in the certificate.

node/libs/roles/src/validator/messages/leader_commit.rs
ElFantasma: Clippy was complaining about it. Not sure why as its not complaining in other similar parts of the code

node/libs/roles/src/validator/messages/leader_prepare.rs
brunoffranca: +1

node/libs/roles/src/validator/messages/leader_commit.rs
```diff
@@ -76,18 +96,28 @@ impl CommitQC {
...
-    pub fn add(&mut self, msg: &Signed<ReplicaCommit>, genesis: &Genesis) {
+    pub fn add(
+        &mut self,
+        msg: &Signed<ReplicaCommit>,
+        genesis: &Genesis,
+    ) -> Result<(), CommitQCAddError> {
+        use CommitQCAddError as Error;
         if self.message != msg.msg {
-            return;
+            return Err(Error::InconsistentViews);
```
brunoffranca, line 106: The error is not quite right. If the message differs it can be because the view is different or because the proposal is different. Better to rename it.

ElFantasma: Ah! you are right. Fixing it

pompon0 (approved):

brunoffranca (approved):

PR #113, opened 2024-05-01
Added comment explaining discrepancy in the field names.

pompon0:
When changing the crypto keys back from bn254 to bls12_381, we forgot to rename the proto fields as well. It is not feasible to do it now, because of backward compatibility, so I've just added comments to explain the discrepancy.

### DIFF
--- a/node/libs/roles/src/proto/validator.proto
+++ b/node/libs/roles/src/proto/validator.proto
@@ -175,14 +175,17 @@ message Signed {
 }
 
 message PublicKey {
+  // The name is wrong, it should be bls12_381.
   optional bytes bn254 = 1; // required
 }
 
 message Signature {
+  // The name is wrong, it should be bls12_381.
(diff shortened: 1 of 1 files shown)


brunoffranca:
Why can't we change it in a backwards compatible way?

brunoffranca:
Also, is the field name visible only in code? I guess not, it's probably visible in logs too.

pompon0:
we store commitqc in postgres in jsonb format

brunoffranca:
Ok. But why can't we change the name?

pompon0:
because the commitqc contains a signature which contains this field. Jsonb uses the text names in encoding.

brunoffranca (approved):
I still think we should find a way of changing the name that is backwards compatible. After all that's why we use protobuf. But since this is mostly a nuisance, I'll approve it anyway.

PR #114, opened 2024-05-16
Enabled support for inactive validators.

pompon0:
We need to be able to configure a node to be a validator BEFORE it is added to the validator committee so that there is no downtime.

### DIFF
--- a/node/actors/executor/src/lib.rs
+++ b/node/actors/executor/src/lib.rs
@@ -103,25 +103,8 @@ impl Executor {
         }
     }
 
-    /// Verifies correctness of the Executor.
-    fn verify(&self) -> anyhow::Result<()> {
-        if let Some(validator) = self.validator.as_ref() {
-            if !self
-                .block_store
-                .genesis()
-                .committee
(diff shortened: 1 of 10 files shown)


node/actors/executor/src/lib.rs
```diff
@@ -133,25 +116,36 @@ impl Executor {
...
-            }
             let (net, runner) =
                 network::Network::new(network_config, self.block_store.clone(), network_actor_pipe);
             net.register_metrics();
-            runner.run(ctx).await.context("Network stopped")
+            s.spawn(async { runner.run(ctx).await.context("Network stopped") });
+
+            // Run the bft actor iff this node is an active validator.
+            let Some(validator) = self.validator else {
+                return Ok(());
```
moshababo, line 126: Maybe add a log entry here as well?

moshababo (approved):

node/actors/executor/src/lib.rs
pompon0: done

PR #115, opened 2024-05-21
BFT-463: Add LeaderSelectionMode::Rota

aakoshh:
## What ❔

Adds a `LeaderSelectionMode::Rota` to allow specifying leaders in subsequent views using a list of public keys. The 1st validator on the list is going to be the leader in view 0, the 2nd in view 1, the 3rd in view 2, wrapping around to start from the beginning as the views increase. There can be repeated keys on the list, and they all have to be part of the committee. 

## Why ❔

This is part of BFT-439 to modify the BFT tests to use the approach outlined in the [Twins paper](https://arxiv.org/abs/2004.10617). Part of that is the ability to control who is the leader in each of the views. 

With this change we should be able to generate a fixed leader schedule for the first few rounds of a protocol. The implementation wraps around so it's a total function, but in practice the tests will provide leaders for all the rounds executed in the test.

### Why not modify `Sticky`?

I thought about whether I should add a new variant to `LeaderSelectionMode` or modify `LeaderSelectionMode::Sticky` to allow multiple values. On the Protobuf bytecode level they should be compatible, and I thought `Sticky` is probably used mostly in tests. 

Based on the answers in the comments below, this is not the case, and it is in fact used in production to represent the centralised sequencer. The genesis it is part of is stored in Postgres as JSONB, so changes in the key name would not be backwards compatible. The `protobuf_conformance` workflow correctly highlighted this regression 🎉

### DIFF
--- a/node/libs/roles/src/validator/messages/consensus.rs
+++ b/node/libs/roles/src/validator/messages/consensus.rs
@@ -59,8 +59,8 @@ pub enum LeaderSelectionMode {
     /// Select in a round-robin fashion, based on validators' index within the set.
     RoundRobin,
 
-    /// Select based on a sticky assignment to a specific validator.
-    Sticky(validator::PublicKey),
+    /// Select based on a sticky assignment to a non-empty list of specific validators.
+    Sticky(Vec<validator::PublicKey>),
 
     /// Select pseudo-randomly, based on validators' weights.
     Weighted,
@@ -174,8 +174,9 @@ impl Committee {
                 }
                 unreachable!()
             }
-            LeaderSelectionMode::Sticky(pk) => {
-                let index = self.index(pk).unwrap();
+            LeaderSelectionMode::Sticky(pks) => {
+                let index = view_number.0 as usize % pks.len();
+                let index = self.index(&pks[index]).unwrap();
                 self.get(index).unwrap().key.clone()
             }
         }
@@ -309,9 +310,13 @@ impl fmt::Debug for Genesis {
 impl Genesis {
     /// Verifies correctness.
     pub fn verify(&self) -> anyhow::Result<()> {
-        if let LeaderSelectionMode::Sticky(pk) = &self.leader_selection {
-            if self.validators.index(pk).is_none() {
-                anyhow::bail!("leader_selection sticky mode public key is not in committee");
+        if let LeaderSelectionMode::Sticky(pks) = &self.leader_selection {
+            for pk in pks {
+                if self.validators.index(pk).is_none() {
+                    anyhow::bail!(
+                        "leader_selection sticky mode public key is not in committee: {pk:?}"
+                    );
+                }
             }
         }
 
--- a/node/libs/roles/src/validator/messages/consensus.rs
+++ b/node/libs/roles/src/validator/messages/consensus.rs
@@ -59,11 +59,14 @@ pub enum LeaderSelectionMode {
     /// Select in a round-robin fashion, based on validators' index within the set.
     RoundRobin,
 
-    /// Select based on a sticky assignment to a non-empty list of specific validators.
-    Sticky(Vec<validator::PublicKey>),
+    /// Select based on a sticky assignment to a specific validator.
+    Sticky(validator::PublicKey),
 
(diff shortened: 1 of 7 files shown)


pompon0:
please remember to make the pr backward compatible

aakoshh:
> please remember to make the pr backward compatible

I tried; can you highlight what you consider a breaking change?

aakoshh:
Are there some JSON based parsers, for config or genesis?

aakoshh:
I can see the config being read from JSON in [node/tools/main](https://github.com/matter-labs/era-consensus/blob/3f5b4f63a44c78e2386eb53abce12f54f9bdaa0e/node/tools/src/main.rs#L35). All other JSON-Protobuf conversions seem to be in various tests. 

So my question is, do you prefer to have a new variant next to `Sticky` because it's used in production configs, or is it a breaking change in JSON format that doesn't actually affect anything other than in-memory tests? 

Happy to add a new variant, just thought I'd open the PR and ask here.

pompon0:
genesis is stored in postgresdb in jsonb format

pompon0:
also, Sticky is currently the only variant used in zksync-era and it will be in use for foreseeable future (main node being the only proposer).

aakoshh:
> also, Sticky is currently the only variant used in zksync-era and it will be in use for foreseeable future (main node being the only proposer)

Thanks. I thought you would do that with a single sized committee and round-robin. 

Okay, what should I call this thing with multiple users... 🤔

pompon0:
plan is to make ENs non-proposer validators

aakoshh:
Hate to ask, but what's an "EN"?

brunoffranca:
External node. That's old terminology though, we can start calling them full nodes.

node/libs/roles/src/validator/messages/consensus.rs
```diff
@@ -64,6 +64,9 @@ pub enum LeaderSelectionMode {
 
     /// Select pseudo-randomly, based on validators' weights.
     Weighted,
+
+    /// Select based on a sticky assignment to a non-empty list of specific validators.
```
brunoffranca, line 68: Comment is not clear enough on how this mode works.

aakoshh:
Thanks @brunoffranca 
That's what I thought, except in your presentation the full nodes were connected through the gossip-network, not the consensus-network, so "non-proposing _validator_" was slightly confusing into which network they would belong to, and if the consensus (as validators), then why external (or full) at the same time?

brunoffranca (approved):
Looks good. I would maybe rename the mode though. Round robin and Rota are synonyms. Since we are rotating on a subset of validators, maybe call it Subset?

node/libs/roles/src/validator/messages/consensus.rs
aakoshh: Hm, I thought I sent a new version already 👀

aakoshh: Sorry, updated.

aakoshh:
> Round robin and Rota are synonyms. 

I don't think they are synonyms... supervisors create the rota for workers to do, they aren't necessarily doing any round robin 🤔 

From Google:
noun
1.
BRITISH
a list showing when each of a number of people has to do a particular job.

> Since we are rotating on a subset of validators, maybe call it Subset?

The problem is it's not a subset, it's a list that can be arbitrarily long and contain arbitrarily repeating elements, e.g. [Alice, Bob, Alice, Alice, Alice, Charlie].

brunoffranca:
So, I think Grzegorz comment was a bit confusing. Let me clear this up.
We have 2 types of nodes: full nodes and validators. Full nodes (previously called external nodes) only participate in the gossip network. Validators participate in the gossip and validator networks.
The plan now is for us to have several validators but only one will be the leader (using the Sticky selection mode).

aakoshh:
From [vocabulary.com](https://www.vocabulary.com/dictionary/rota): 
> You can also call a rota a roster. 

@brunoffranca would you prefer `Roster`?

brunoffranca:
I can see you like the name Rota. Let's keep it.

pompon0 (approved):

PR #116, opened 2024-05-22
BFT-464: Twins generator

aakoshh:
## What ❔

Implements utilities to support generating [Twins scenarios](https://arxiv.org/pdf/2004.10617)
- [x] Generate all possible partitioning of nodes
- [x] Prune partitions so that either always, or at least eventually, they always have a partition capable of producing a QC
- [x] Function to size items based on desired number of replicas and twins
- [x] Pair partitions with possible leaders
- [x] Permutate leader-partition combinations for a desired number of rounds
- [x] Generate a desired number of random scenarios

# Liveness
I haven't yet added the the logic for creating rounds where there is no chance for a quorum in any of the partitions. It would be easy to do, but I'd like to ask your opinion on how many to mix in, or whether we should always end with a potential quroum, etc. Perhaps just a probability of a no-quorum round?

## Why ❔

So that we can generate the combinations and permutations that we're going to execute with the mock network.


node/actors/bft/src/testonly/twins/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl<'a, T> Partitioner<'a, T> {
+    /// Generate all possible partitioning.
+    fn generate(items: Partition<'a, T>, num_partitions: usize) -> Vec<Partitioning<'a, T>> {
+        if num_partitions == 0 || items.len() < num_partitions {
+            // Impossible to partition.
+            Vec::new()
+        } else if num_partitions == items.len() {
+            // Each items stands alone.
```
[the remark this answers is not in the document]
aakoshh: I started with that, but like this better for some reason.

node/actors/bft/src/testonly/twins/tests.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use rand::Rng;
+use zksync_concurrency::ctx;
+
+use super::{partitions, Partitioning};
+
+#[test]
+fn test_partitions() {
+    let got = partitions(&[&1, &2, &3], 2);
+    let want: HashSet<Partitioning<_>> = HashSet::from_iter(vec![
```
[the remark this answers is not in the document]
aakoshh: I kept the inner part as `Vec<Vec<_>>` because:
* I thought it would be an easier syntax to define all the expected cases in the test with `vec![...]`
* the algorithm actually preserves the order to some extent, e.g. the leftmost item is always in the first partition

But I didn't have any particular expectation about the order of in which they appear, so I thought an external `HashSet` can be used to check equivalence from that point.

Maybe I'll add some different labels to remove any doubt that it's from the ordering of the item type.

node/actors/bft/src/testonly/twins/mod.rs
```diff
@@ -10,125 +10,45 @@
...
-                if ins_part == first_empty {
-                    first_empty + 1
-                } else {
-                    first_empty
-                },
-            );
-            // Backtrack
-            self.acc[ins_part].pop();
-        }
+    /// This is a reflexive, they are twins of each other.
```
pompon0, line 49: afaict this is a symmetric relation, but not reflexive.

node/actors/bft/src/testonly/twins/partition.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl<'a, T> Partitioner<'a, T> {
+    /// Generate all possible partitioning.
+    fn generate(items: Partition<'a, T>, num_partitions: usize) -> Vec<Partitioning<'a, T>> {
+        if num_partitions == 0 || items.len() < num_partitions {
+            // Impossible to partition.
+            return Vec::new();
+        } else if num_partitions == items.len() {
+            // Each items stands alone.
```
pompon0, line 46: I meant `return` instead of `else`. It doesn't make sense to have both.

node/actors/bft/src/testonly/twins/mod.rs
aakoshh: Ah yes, very true, thanks 🤦

node/actors/bft/src/testonly/twins/partition.rs
aakoshh: Sorry, I don't understand. The `else` has been removed.

node/actors/bft/src/testonly/twins/tests.rs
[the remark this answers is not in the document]
pompon0: fyi HashSet can be constructed just like `[a,b,c].into()` assuming that rust can deduce the target type.

node/actors/bft/src/testonly/twins/partition.rs
```diff
@@ -0,0 +1,115 @@
...
+        // Now generate all possible allocations.
+        for ins_part in min_part..=max_part {
+            self.acc[ins_part].push(self.items[idx]);
+            self.go(
+                idx + 1,
+                if ins_part == first_empty {
+                    first_empty + 1
+                } else {
+                    first_empty
+                },
```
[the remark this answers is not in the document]
aakoshh: Has to be `first_empty + ((ins_part == first_empty) as usize)` 
Okay, looks better indeed.

node/actors/bft/src/testonly/twins/mod.rs
```diff
@@ -10,125 +10,45 @@
...
-    /// * `B` can go into minimally into `P1` because there are 2 more items after it, which
-    /// is enough to fill all remaining partitions; or it can go into `P2`, because `P1` is filled.
-    /// * `C` depends on what we did with `B`: if `B` is in `P1` then `C` has to minimally go into
-    /// `P2` to make sure no partition will be left empty at the end; if `B` is in `P2` then `C`
-    /// can go either in `P1`, `P2` or `P3`
+/// Abstraction for nodes which can have a twin.
+pub trait Twin: HasKey
+where
+    Self: PartialEq,
+{
```
pompon0, line 43: nit: I trust that you have some need to make this abstract, but afaic having concrete types was also ok. Using traits has its own limitations so use your judgment.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,182 @@
...
+use rand::{seq::SliceRandom, Rng};
+
+use super::{partitions, HasKey, Partitioning, Twin};
+
+/// A cluster holds all the nodes in the simulation, some of which are twins.
+pub struct Cluster<T> {
+    /// All nodes, twins and originals.
+    nodes: Vec<T>,
+    /// The ideal size of the committee, ie. the `n` in the `n = 5 * f + 1`.
+    num_replicas: usize,
```
pompon0, line 12: do you require `num_replicas%5==1` though? If so, it would be nice to assert that in the constructor.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /// The number of faulty nodes the committee can tolerate for the `n = 5 * f + 1` relation to hold.
+    pub fn num_faulty(&self) -> usize {
+        (self.num_replicas - 1) / 5
+    }
+
+    /// The number of signatures required for a quorum, ie. `4 * f + 1`
+    pub fn quorum_size(&self) -> usize {
+        self.num_faulty() * 4 + 1
```
pompon0, line 69: in general quorum is n-f, not 4f+1. IMO it would be good to stick to that notation, even if your code only supports n=5f+1.

node/actors/bft/src/testonly/twins/mod.rs
aakoshh: It's not that I absolutely needed traits to be abstract, I just like to develop in a way where I can ignore stuff like what exactly is going to be used for this or that type (like a `bn245` or whatever key is used) when it's completely irrelevant to this case. If we work with abstractions, then there can't be any tight coupling either, and the minimum expectations can be expressed. Code is also easier to extract, should we want to. 

The `Twin` itself is just an idea for how to deal with these things. It turns out it's not really needed, there could be a constructor which puts the job of creating twins on the caller, but I thought it can't hurt to have it, to make it easier to understand what's going on.

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Great question, I don't think we actually have such a requirement. If you called this with, say, 4 replicas, then it should calculate the `num_faulty` as 0. If you still ask for twins to be created, then you should see the consensus fail. So I would like you to be able to create anything you want, not just the happy path.

node/actors/bft/src/testonly/twins/partition.rs
```diff
@@ -0,0 +1,115 @@
+use std::cmp::min;
+
+/// A group of nodes that can talk to each other.
+pub type Partition<'a, T> = Vec<&'a T>;
```
pompon0, line 4: is Partition/Partitioning naming from the paper? Isn't https://en.wikipedia.org/wiki/Partition_of_a_set a more standard naming? I.e. Partion -> Set, Partitioning -> Partition?

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: In our literature I have always seen 4*f+1, so I thought for anyone reading this, it's more helpful as a distinction. 

For example if we take a look at https://www.notion.so/matterlabs/Two-step-Fast-HotStuff-487def0222f14bb780648357380691e8 I count 14 instances of "4f+1" and exactly 1 "n-f"

But I agree, `n-f` is more general and we can change the equation to be more general, no problem.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        // Potential leaders
+        let keys = cluster
+            .replicas()
+            .iter()
+            .map(|r| r.key())
+            .collect::<Vec<_>>();
+
+        // Create all possible partitionings; the paper considers 2 or 3 partitions to be enough.
+        let partitions = (1..=max_partitions).flat_map(|np| partitions(cluster.nodes(), np));
```
pompon0, line 128: can you make the naming a bit more consistent? Right now `partitions` variable is actually partitionings which consist of partitions.

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: And if `f = 0` it will actually be `n` which is better than 1 👍

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Yeah, I know. It's just "partitionings" is a daft name 🤔

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+/// Count the number of node keys that do not have twins in the group.
+pub fn unique_key_count<T>(nodes: &[T]) -> usize
+where
+    T: HasKey,
+{
+    let mut seen = BTreeSet::new();
+    let mut cnt = 0;
```
pompon0, line 171: nit: `nodes.iter().collect::<BTreeSet<_>>().len()`

node/actors/bft/src/testonly/twins/scenario.rs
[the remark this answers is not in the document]
aakoshh: I think I'm okay with it as it is. It only generates exactly n partitions, doesn't stop halfway. Here we include it from 1 to np, but that doesn't have to be the case. Easier to think about that function on its own as it is IMO.

pompon0 (approved):

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Hm, TBH I don't know why I didn't do that 👍

node/actors/bft/src/testonly/twins/partition.rs
pompon0: ? `} else if num_partitions == items.len() {`

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Although it's the keys, not the whole thing that needs to be counted.

node/actors/bft/src/testonly/twins/partition.rs
pompon0: if you want to keep this naming, then please rename partitions to partitionings consistently

node/actors/bft/src/testonly/twins/partition.rs
aakoshh: Yes? `else if` with `return`, not `else` 🤔 

What used to be the `else` starts at `// Create empty`

node/actors/bft/src/testonly/twins/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+mod partition;
+mod scenario;
+
+pub use partition::*;
+pub use scenario::*;
+
+/// Abstraction for things that can lead a round by the virtue of having a validator key.
+pub trait HasKey {
+    /// Validator public key.
```
pompon0, line 22: nit: unless you really need a linear ordering, consider Hash instead of Ord.

node/actors/bft/src/testonly/twins/partition.rs
pompon0: oh, I meant removing both `else` and replacing `else if` with `if` as the non-happy path exits anyway.

node/actors/bft/src/testonly/twins/partition.rs
aakoshh: Renamed to `Split` and `splits`

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Renamed to `splits`

node/actors/bft/src/testonly/twins/mod.rs
aakoshh: I went with `Ord` because I thought you are more likely to have it for your public key types than `Hash`, since for deterministic behaviour we have to use `BTreeMap`. Also because with `Ord` we _can_ have a deterministic order for the keys in the scenarios, should we ever need it.

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: Hm, actually if we consider n=5f+1 and take n=7, then f=1, and the 4f+1 the replicas would be waiting for is 5, yet n-f is 6.

aakoshh: For a normal consensus it would be `quorum_size = (size + maxFaulty) / 2 + 1`

aakoshh: Okay, so the spec says 4f+1, but it's just a theoretical value that we don't actually use. Like most "normal" consensus that use n*2/3+1, we use the equivalent of n*4/5+1, at least the new tests I added pass with this criteria.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /// The number of signatures required for a quorum, ie. `4 * f + 1` or more generally `n - f`, which in this case equals `n*4/5+1`
+    pub fn quorum_size(&self) -> usize {
+        self.num_replicas() - self.num_faulty()
+    }
+
+    /// The number of signatures required for a high-vote, ie. `2 * f + 1`
+    pub fn subqourum_size(&self) -> usize {
+        self.num_faulty() * 2 + 1
```
aakoshh, line 74: This should be `self.num_fault() * 2 / 5 + 1` for it not to degenerate into 1 if `num_replicas <= 5`.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,171 @@
...
+        assert!(!cluster.nodes().is_empty(), "empty cluster");
+
+        // Potential leaders
+        let keys = cluster.replicas().iter().map(|r| r.key()).collect();
+
+        // Create all possible partitionings; the paper considers 2 or 3 partitions to be enough.
+        let splits = (1..=max_partitions).flat_map(|np| splits(cluster.nodes(), np));
+
+        // Prune partitionings so that all of them have at least one where a quorum is possible.
+        // Alternatively we could keep all and make sure every scenario has eventually one with a quorum, for liveness.
```
aakoshh, line 131: One thing to consider here would be how the network is implemented in the tests. My idea was that I'd look at the `view` of the message, and then look at the partitioning in that round, and deliver to all nodes who are in the same partition as the sender. 

The ChonkyBFT spec repeats timeouts periodically, however if ever node would be in a partition with a size less than 4f+1, then they would never have a TimeoutQuorum, and would never move on to the next view. 

Whereas if at least one partition has 4f+1 views, then they can time out, and in the next view they might be broadcasting to their stranded brethren, who at that point can move on to the future view on which they now observe a quorum in the justification.

node/actors/bft/src/testonly/twins/mod.rs
pompon0: this is testonly so it doesn't matter.

node/actors/bft/src/testonly/twins/scenario.rs
pompon0: this is not a degenerate case. If you want to change it, you need to prove correctness of the algorithm for all possible values of n.

node/actors/bft/src/testonly/twins/scenario.rs
pompon0: The currently implemented network semantics is that all (consensus) messages are delivered eventually, unless overridden by a newer message (see MessagePool for exact semantics).

node/actors/bft/src/testonly/twins/scenario.rs
pompon0: Then please remove this assumption from your pr. In particular adjust this comment.

node/actors/bft/src/testonly/twins/scenario.rs
aakoshh: So we start with `n = 5f+1` and therefore `f = (n-1)/5`. @brunoffranca says the correct formula would be `sq = n-3*f`, so let's see: 

```
sq = n-3*f 
sq = n - 3 * (n-1)/5
sq = n - 3/5n + 3/5
sq = 2/5n + 3/5
```

Now let's start from the current `sq = 2f+1`:
```
sq = 2f+1
sq = 2*(n-1)/5 + 1
sq = 2/5*n -2/5 + 1
sq = 2/5n + 3/5
```

I believe these are equivalent in mathematical terms, however due to the integer arithmetic, for n<6 we have f=0, but that's just an artifact of how computers work. 

Similarly how the spec says the commit quorum size is 4f+1, when n=5 it gives 1, which is clearly wrong, for the same reason I think it's wrong for 2f+1, except we use it for the more difficult to understand subquorum, where nobody who sees it first is really sure if it's an optimisation or a required for safety. 

Here are the values for the different values of `n`:
```python
>>> for t in [(n, f, 2*f+1, 2*n//5+1, n-3*f) for (n, f) in  [(n, (n-1)//5) for n in range(1, 15)]]: print("n={} f={} sq_2f1={} sq_25n1={} sq_n3f={}".format(*t))
... 
n=1 f=0 sq_2f1=1 sq_25n1=1 sq_n3f=1
n=2 f=0 sq_2f1=1 sq_25n1=1 sq_n3f=2
n=3 f=0 sq_2f1=1 sq_25n1=2 sq_n3f=3
n=4 f=0 sq_2f1=1 sq_25n1=2 sq_n3f=4
n=5 f=0 sq_2f1=1 sq_25n1=3 sq_n3f=5
n=6 f=1 sq_2f1=3 sq_25n1=3 sq_n3f=3
n=7 f=1 sq_2f1=3 sq_25n1=3 sq_n3f=4
n=8 f=1 sq_2f1=3 sq_25n1=4 sq_n3f=5
n=9 f=1 sq_2f1=3 sq_25n1=4 sq_n3f=6
n=10 f=1 sq_2f1=3 sq_25n1=5 sq_n3f=7
n=11 f=2 sq_2f1=5 sq_25n1=5 sq_n3f=5
n=12 f=2 sq_2f1=5 sq_25n1=5 sq_n3f=6
n=13 f=2 sq_2f1=5 sq_25n1=6 sq_n3f=7
n=14 f=2 sq_2f1=5 sq_25n1=6 sq_n3f=8
```

I think if `n-3f` is correct and `2f+1` is correct then the middle ground must be correct too. You can see they don't give the exact same results all across the board, but of course the starkest difference is below n=6. 

If it's a safety feature, ie. if there is exactly one highest subquorum then you _must_ repropose it, then I'm a bit worried that if the sq=1, it's more likely that we _don't_ repropose because we see two of them, and lose this security. I don't have an example for it from the top of my head, but it seems more intuitive for the other values of 2f+1 that if there are two of them then there surely wasn't any quorum. For example if n=5 and a leader go isolated and voted for their own proposal, but missed the fact that there was a quorum for a value already in the previous round.

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,171 @@
...
+        (self.num_replicas - 1) / 5
+    }
+
+    /// The number of signatures required for a quorum, ie. `4 * f + 1` or more generally `n - f`, which in this case equals `n*4/5+1`
+    pub fn quorum_size(&self) -> usize {
+        self.num_replicas() - self.num_faulty()
+    }
+
+    /// The number of signatures required for a high-vote, ie. `2 * f + 1`
+    pub fn subqourum_size(&self) -> usize {
```
brunoffranca, line 73:
```suggestion
    pub fn subquorum_size(&self) -> usize {
```

node/actors/bft/src/testonly/twins/scenario.rs
brunoffranca, line 74: Use n-3*f

node/actors/bft/src/testonly/twins/scenario.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    /// Generate a single run for the agreed upon number of rounds.
+    pub fn generate_one(&self, rng: &mut impl Rng) -> Scenario<'a, T> {
+        let mut rounds = Vec::new();
+        // We could implement this with or without replacement.
+        // In practice there probably are so many combinations that it won't matter.
+        // For example to tolerate 2 faulty nodes we need 11 replicas, with 2 twins that's 13,
+        // which results in hundreds of thousands of potential paritionings, multiplied
```
brunoffranca, line 152:
```suggestion
        // which results in hundreds of thousands of potential partitionings, multiplied
```

brunoffranca (approved):
Looks good to me. Approved!

node/actors/bft/src/testonly/twins/scenario.rs
brunoffranca: The best approach IMO is to unpartition every scenario at the end. Partitions are a temporary network condition anyway, with these tests we want to see if the consensus can recover from a partition.

PR #117, opened 2024-05-24
BFT-465: Twins tests

aakoshh:
## What ❔

Add tests that run multiple Twins scenarios with random number of replicas as twins.

- [x] Add `Network::Twins` type with a partition schedule
- [x] Add `run_nodes_twins` to execute the test with arbitrary network partitions and potentially multiple ports per validator
- [x] Buffer messages that can't be delivered and deliver them sometime later (potentially unless superseded by a newer message from source to target)
- [x] Simulate the gossiping of finalized blocks
- [x] Run test with no twins and no partitions
- [x] Run test with no twins but random partitions 
- [x] Run test with twins and random partitions 
### Investigation

```shell
cargo nextest run -p zksync_consensus_bft twins_network --no-capture
```

Initially the test failed when partitions are introduced. I'll try to understand if this is because the leader got isolated and an important message got lost. Would like to understand if eventual delivery is an absolute requirement even if all partitions are at least as large as the quorum size.

🔍  I think the reason for the test failing is because it looks for all nodes having persisted a certain number of blocks, but if a proposal with the payload is missed due to a partition preventing the message from being delivered, then there is no mechanism in the machinery instantiated by the test to procure these later, and those nodes are stuck.

🔧 I implemented a message stashing mechanism in the mock network but it still doesn't finalise blocks 👀 

🔍 The problem seems to be that my unstashing mechanism only kicked in when node A tried to send a new message to node B and wasn't blocked any more. However if A didn't try to send, then the stashed messages to B weren't delivered, which causes longer and longer timeouts as nobody is making progress for one reason or another. Meanwhile for example C can be already in a new view, so if we see that, we could conclude that A-to-B should be unstashed as well even if there are no messages from A to B in that round. 

🔧 I'll try testing after merging https://github.com/matter-labs/era-consensus/pull/119 which should trigger unstashes in each round.

🔍 It's still failing after adding the replica-to-replica broadcast. For example one replica A is isolated in a round 1 and doesn't get a LeaderCommit; then in the next round replica B is isolated, and A gets all the missing messages from round 1, plus the new LeaderPrepare, but it doesn't respond with a ReplicaCommit because it doesn't have the block from round 1, and therefore cannot even store proposal 2. The consensus relies on the external gossiping mechanism to eventually propagate the FinalBlock to it; until then the node is stuck. I need to simulate the effect of gossiping in the test.

🔧 I implemented a simulated gossip using the following mechanism: if one node is about to send/unstash a message to another and they are in a gossip relationship, and the message contains a CommitQC, and the sender has the finalized block for the committed height, then it inserts the block directly into the target store.

🔍 The tests now work without twins, but fail with 1 or 2 twins, albeit not on every scenario

🔧 Changed the simulated gossip to push all ancestors of a finalized block into the target blockstore, not just the one in the CommitQC that is in the latest message. This simulates the ability of the target to fetch all missing blocks.

## Why ❔

To check that the consensus does not fail as long as the number of twins does not exceed the tolerable number of Byzantine nodes.

### DIFF
--- a/node/actors/bft/src/testonly/run.rs
+++ b/node/actors/bft/src/testonly/run.rs
@@ -1,4 +1,5 @@
 use super::{Behavior, Node};
+use anyhow::bail;
 use network::{io, Config};
 use rand::prelude::SliceRandom;
 use std::{
@@ -377,6 +378,12 @@ async fn twins_receive_loop(
         // If the view is higher than what we have planned for, assume no partitions.
         // Every node is guaranteed to be present in only one partition.
         let partitions_opt = splits.get(view_number);
+
+        if partitions_opt.is_none() {
+            bail!(
+                "ran out of scheduled rounds; most likely cannot finalize blocks even if we go on"
+            );
+        }
 
         let msg = || {
             io::OutputMessage::Consensus(io::ConsensusReq {
--- a/node/actors/bft/src/testonly/run.rs
+++ b/node/actors/bft/src/testonly/run.rs
@@ -220,6 +220,7 @@ async fn run_nodes_twins(
         // with `async move`, which in turn is necessary otherwise it the spawned process could not borrow `port`.
         let sends = &sends;
         let validator_ports = &validator_ports;
+        let start = &ctx.now();
         // Run networks by receiving from all consensus instances:
         // * identify the view they are in from the message
         // * identify the partition they are in based on their network id
@@ -244,8 +245,16 @@ async fn run_nodes_twins(
                     //   an actual timeout loop and might rely on the network trying to re-broadcast forever.
                     // * OTOH if all partitions are at least quorum sized, then we there will be a subset of nodes that
                     //   can move on to the next view, in which a new partition configuration will allow them to broadcast
-                    //   to previously isolated peers, which will indicate that the buffered messages can be sent as the
-                    //   partition has been "healed".
+                    //   to previously isolated peers. 
+                    // * One idea is to wait until replica A wants to send to replica B in a view when they are no longer 
+                    //   partitioned, and then unstash all previous A-to-B messages. This would _not_ work with HotStuff
+                    //   out of the box, because replicas only communicate with their leader, so if for example B missed
+                    //   a LeaderCommit from A in an earlier view, B will not respond to the LeaderPrepare from C because
+                    //   they can't commit the earlier block until they get a new message from A. However since 
+                    //   https://github.com/matter-labs/era-consensus/pull/119 the ReplicaPrepare messages are broadcasted, 
+                    //   so we shouldn't have to wait long for A to unstash its messages to B.
+                    // * If that wouldn't be acceptable then we could have some kind of global view of stashed messages
+                    //   and unstash them as soon as someone moves on to a new view.
                     let mut stashes: HashMap<Port, Vec<io::OutputMessage>> = HashMap::new();
 
                     // Either stash a message, or unstash all messages and send them to the target along with the new one.
@@ -253,6 +262,7 @@ async fn run_nodes_twins(
                         |can_send: bool, target_port: Port, msg: io::OutputMessage| {
                             let stash = stashes.entry(target_port).or_default();
                             let view = output_msg_view_number(&msg);
+                            let kind = output_msg_label(&msg);
 
                             if can_send {
                                 let s = &sends[&target_port];
@@ -261,13 +271,15 @@ async fn run_nodes_twins(
                                 stash.shuffle(rng);
 
                                 for unstashed in stash.drain(0..) {
-                                    eprintln!("^-> unstashed view={view} unstashed-view={} from={port} to={target_port}", output_msg_view_number(&unstashed));
+                                    let view = output_msg_view_number(&unstashed);
+                                    let kind = output_msg_label(&unstashed);
+                                    eprintln!("   ^^^ unstashed view={view} from={port} to={target_port} kind={kind}");
                                     s.send(unstashed);
                                 }
-                                eprintln!("--> sending view={view} from={port} to={target_port}");
+                                eprintln!("   >>> sending view={view} from={port} to={target_port} kind={kind}");
                                 s.send(msg);
                             } else {
-                                eprintln!("--V stashed view={view} from={port} to={target_port}");
+                                eprintln!("   VVV stashed view={view} from={port} to={target_port} kind={kind}");
                                 stash.push(msg)
                             }
                         };
@@ -297,7 +309,7 @@ async fn run_nodes_twins(
                                 Some(ps) => {
                                     for p in ps {                                        
                                         let can_send = p.contains(&port);
-                                        eprintln!("broadcasting view={view_number} from={port} target={:?} can_send={can_send}", p);
+                                        eprintln!("broadcasting view={view_number} from={port} target={p:?} can_send={can_send} t={}", start.elapsed().as_seconds_f64());
                                         for target_port in p {
                                             send_or_stash(can_send, *target_port, msg());
                                         }
@@ -319,7 +331,7 @@ async fn run_nodes_twins(
                                             let can_send = p.contains(&port);
                                             for target_port in target_ports {
                                                 if p.contains(target_port) {
-                                                    eprintln!("unicasting view={view_number} from={port} target={target_port } can_send={can_send}");
+                                                    eprintln!("unicasting view={view_number} from={port} target={target_port } can_send={can_send} t={}", start.elapsed().as_seconds_f64());
                                                     send_or_stash(
                                                         can_send,
                                                         *target_port,
@@ -346,5 +358,11 @@ async fn run_nodes_twins(
 fn output_msg_view_number(msg: &io::OutputMessage) -> u64 {
     match msg {
         io::OutputMessage::Consensus(cr) => cr.msg.msg.view().number.0,
+    }
+}
+
+fn output_msg_label(msg: &io::OutputMessage) -> &str {
+    match msg {
+        io::OutputMessage::Consensus(cr) => cr.msg.msg.label()
     }
 }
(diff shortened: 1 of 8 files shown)


node/actors/network/src/testonly.rs
```diff
@@ -13,7 +13,10 @@ use std::{
     sync::Arc,
 };
 use zksync_concurrency::{ctx, ctx::channel, io, limiter, net, scope, sync};
-use zksync_consensus_roles::{node, validator};
+use zksync_consensus_roles::{
+    node,
+    validator::{self, SecretKey},
```
pompon0, line 18: we use the convention of using qualified names for types in roles crate: things like SecretKey get easily ambiguous without knowing which role it belong to.

node/actors/network/src/testonly.rs
```diff
@@ -77,7 +80,21 @@ pub fn new_configs(
...
+/// Construct configs for `n` validators of the consensus.
+///
+/// This version allows for repeating keys used in Twins tests.
+pub fn new_configs_for_validators<'a, I>(
+    rng: &mut impl Rng,
+    validator_keys: I,
+    gossip_peers: usize,
+) -> Vec<Config>
+where
+    I: Iterator<Item = &'a SecretKey>,
```
pompon0, line 95: nit: you can use `I: impl Iterator<Item = &'a SecretKey>`

node/actors/bft/src/testonly/run.rs
```diff
@@ -1,18 +1,41 @@
 use super::{Behavior, Node};
-use std::collections::HashMap;
+use anyhow::bail;
+use network::{io, Config};
+use rand::prelude::SliceRandom;
```
pompon0, line 4: nit `rand::seq::SliceRandom as _`. `prelude` is only for `use rand::prelude::*` and we don't use wildcard imports.

node/actors/bft/src/testonly/run.rs
```diff
@@ -1,18 +1,41 @@
...
+    ctx::{
+        self,
+        channel::{self, UnboundedReceiver, UnboundedSender},
+    },
+    oneshot, scope,
+};
 use zksync_consensus_network as network;
-use zksync_consensus_roles::validator;
-use zksync_consensus_storage::testonly::new_store;
+use zksync_consensus_roles::validator::{
```
pompon0, line 18: please use validator qualifier.

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+                async {
+                    let mut network_pipe = dispatcher_pipe;
+                    spec.run(ctx, &mut network_pipe).await
                 }
+                .instrument(tracing::info_span!("node", i)),
+            );
+        }
+        // Taking these references is necessary for the `scope::run!` environment lifetime rules to compile
+        // with `async move`, which in turn is necessary otherwise it the spawned process could not borrow `port`.
+        // Potentially `ctx::NoCopy` could be used with `port`.
```
pompon0, line 256: nit: NoCopy would be preferrable, as `scope::run` is intended to be supplied with `async`, not `async move`.

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+#[allow(clippy::too_many_arguments)]
+async fn twins_receive_loop(
+    ctx: &ctx::Ctx,
+    splits: &PortSplitSchedule,
+    validator_ports: &HashMap<PublicKey, Vec<Port>>,
+    sends: &HashMap<Port, UnboundedSender<io::OutputMessage>>,
+    gossip_targets: &HashSet<Port>,
+    gossip_send: UnboundedSender<(Port, Port, BlockNumber)>,
+    port: Port,
+    mut recv: UnboundedReceiver<io::InputMessage>,
```
pompon0, line 309: nit: that's a lot of args, are you sure you don't want to turn it into a method of some struct?

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+                s.send(msg);
+            };
+
+            // Messages can be delivered in arbitrary order.
+            stash.shuffle(rng);
+
+            for unstashed in stash.drain(0..) {
+                let view = output_msg_view_number(&unstashed);
+                let kind = output_msg_label(&unstashed);
+                eprintln!("   ^^^ unstashed view={view} from={port} to={target_port} kind={kind}");
```
pompon0, line 367: nit: we use `tracing::info` for that stuff

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+                send(unstashed);
+            }
+            eprintln!("   >>> sending view={view} from={port} to={target_port} kind={kind}");
+            send(msg);
+        } else {
+            eprintln!("   VVV stashed view={view} from={port} to={target_port} kind={kind}");
+            stash.push(msg);
+        }
```
pompon0, line 375: nit: you would get less indentation, by treating "stashing path" as early exit path, i.e:
```
if !can_send {
  stash.push(msg);
  return;
}
...
```

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+/// referenced finalized block in its store, then assume that the target
+/// could fetch this block if they wanted, and insert it directly into their store.
+///
+/// This happens concurrently with the actual message passing, so it's just an
+/// approximation. It also happens out of order. This method only contains the
+/// send loop, to deal with the spawning of store operations.
+async fn twins_gossip_loop(
+    ctx: &ctx::Ctx,
+    stores: &HashMap<Port, Arc<BlockStore>>,
+    mut recv: UnboundedReceiver<(Port, Port, BlockNumber)>,
```
pompon0, line 457: that type deserves to be a struct

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+    Ok(())
+}
+
+/// Simulate the effects of gossip: if the source node can gossip to the target,
+/// and the message being sent contains a CommitQC, and the sender has the
+/// referenced finalized block in its store, then assume that the target
+/// could fetch this block if they wanted, and insert it directly into their store.
+///
+/// This happens concurrently with the actual message passing, so it's just an
+/// approximation. It also happens out of order. This method only contains the
```
pompon0, line 452: would it be hard to implement gossip by looking at the current partitions and syncing any missing blocks directly, rather than do syncing on connections induced by other messages?

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+                    break;
+                };
             }
         }
         Ok(())
     })
     .await
 }
+
+fn output_msg_view_number(msg: &io::OutputMessage) -> u64 {
```
pompon0, line 495: afaict there is no need to strip the type safety here. Returning ViewNumber would be more readable, no?

node/actors/bft/src/testonly/run.rs
```diff
@@ -71,77 +110,407 @@ impl Test {
...
+    let block_to_gossip = |target_port: Port, msg: &io::OutputMessage| {
+        if !gossip_targets.contains(&target_port) {
+            return None;
+        }
+        output_msg_commit_qc(msg).map(|qc| qc.header().number)
+    };
+
+    // We need to buffer messages that cannot be delivered due to partitioning, and deliver them later.
+    // The spec says that the network is expected to deliver messages eventually, potentially out of order,
+    // caveated by the fact that the actual implementation only keeps retrying the last message..
```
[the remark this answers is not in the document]
brunoffranca: +1

brunoffranca (approved):
Very nice work! As a side-note, really appreciate the quality of the documentation in the code and PR.

node/actors/bft/src/testonly/run.rs
aakoshh: At the moment there is no concept of "current" in the simulation, the partitions are only consulted in reaction to a message being passed from one actor to another, and for that the view is determined from the message. This is why I left comments to remind that if ChonkyBFT stayed in the same view until a quorum is reached, and we created partitions which have less members than required for a quorum, then we would need another mechanism to lift the partitions in the view "after a while". I left that for a future exercise because I didn't want to introduce timeouts in this test, and the current implementation goes on to the next view on its own anyway.

If I understand your question correctly you're also asking if it would be better to have a sync loop periodically looking at pulling missing blocks, without being triggered by a different message being sent. 

I don't have a strong opinion on that. IIUC the way syncing works is it first waits to be notified via gossip that the remote peer has a range of blocks we are interested in, and then pulls the block via RPC, so a probing loop would still be just an approximation, as it would be missing the initial message from the "have" to the "have not". 

It would be great if the actual gossip messages could be part of the simulation itself, but short of that I don't see one way being much better than the other, as long as we don't introduce stronger assumptions than what is realistic, e.g. by always inserting into the target store before even sending the message, so the target never has to wait for missing blocks to appear.

node/actors/bft/src/testonly/run.rs
```diff
@@ -1,18 +1,41 @@
...
+use zksync_consensus_storage::{testonly::new_store, BlockStore};
 use zksync_consensus_utils::pipe;
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub(crate) enum Network {
     Real,
     Mock,
+    Twins(PortSplitSchedule),
 }
```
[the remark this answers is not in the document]
pompon0: That depends, on whether we make the Twins network use the real network or not. In particular the current implementation of how the message retransmission is too fragile IMO.

pompon0 (approved):

node/actors/bft/src/testonly/run.rs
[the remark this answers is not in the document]
brunoffranca: Twins should use only the mock network. The purpose is to find issues with the consensus algorithm, not to be an integration test.

PR #118, opened 2024-05-24
[BFT-445] L1 batch gossip sync

IAvecilla:
## What ❔

Enable syncing L1 batches by gossip sync.

## Why ❔

Just like blocks, nodes needs to fetch batches of blocks from other peers.


node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -1,25 +1,153 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 /// A batch number.
 pub struct BatchNumber(pub u64);
 
 impl BatchNumber {
     /// Increment the batch number.
     pub fn next(&self) -> BatchNumber {
-        BatchNumber(self.0.checked_add(1).unwrap())
+        BatchNumber(self.0 + 1)
```
pompon0, line 47: ?

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -1,25 +1,153 @@
 use super::{Signed, Signers};
 use crate::{attester, validator::Genesis};
 use anyhow::{ensure, Context as _};
+use zksync_consensus_crypto::{keccak256::Keccak256, ByteFmt, Text, TextFmt};
 
-#[derive(Clone, Debug, PartialEq, Eq, Hash, Default, PartialOrd)]
+/// Payload of the batch. Consensus algorithm does not interpret the payload
+/// (except for imposing a size limit for the payload). Proposing a payload
+/// for a new batch and interpreting the payload of the finalized batches
+/// should be implemented for the specific application of the consensus algorithm.
```
pompon0, line 9: I think there is some confusion here.
Batches will be represented in 2 entirely separate formats in era-consensus:
1. the batch commitment that will be signed and sent to L1. This format needs to be L1 friendly both the signature scheme and format will be dictated by zksync-era repo. era-consensus doesn't need a concept of payload for this format.
2. the batch as a sequence of blocks + some merkle paths (opening the batch commitment from point 1) proving the correctness of the batch. The format of the proof is up to zksync-era and shouldn't be interpreted in era-consensus. Whether the batch content should be fully opaque, or partially structured (sth like `struct { payloads: []Payload, proof: Vec<u8> }`) is TBD.

pompon0: in particular there is no such a thing as FinalBatch, because the `proof` in the 2nd format is NOT the qc in format `1`.

node/tools/src/tests.rs
```diff
@@ -33,6 +33,7 @@ impl Distribution<AppConfig> for EncodeDist {
             genesis: genesis.with_hash(),
             max_payload_size: rng.gen(),
             validator_key: self.sample_opt(|| rng.gen()),
+            attester_key: self.sample_opt(|| rng.gen()),
```
pompon0, line 36: also add attester key to the key generator tool.

node/libs/storage/src/testonly/mod.rs
```diff
@@ -29,17 +32,66 @@ impl Distribution<ReplicaState> for Standard {
     }
 }
 
+/// Test-only memory storage for blocks and batches.
+pub struct TestMemoryStorage {
+    /// In-memory block store with its runner.
```
pompon0, line 37: There is no point in exposing multiple runners - wrap them into a single runner which just spawns those 2 inside.

node/libs/roles/src/attester/messages/batch.rs
brunoffranca: @IAvecilla I didn't find any other issue with this PR, so I'll just comment here. The batch syncing method that @pompon0 wants to add does confuse matters a bit in the code.
There are two types like @pompon0 says. The first type is what we are implementing with the whole L1 batch signing task. The current struct just contains a BatchNumber, and you can assume that in the future it will also contain a BatchHash, but that's it. A number, a hash, it gets signed with the attester key and aggregated into a QC.
The type that we are using for this syncing only exists while nodes are syncing, and is completely separate from the L1 batch signing task. IMO shouldn't be called a Batch (it just creates confusion) and shouldn't be in the roles crate (1. all nodes sync, not just attesters and 2. these are not signed with any key).

pompon0 (approved):

brunoffranca (approved):

PR #119, opened 2024-05-27
Make validators broadcast the high Commit QC (BFT-458)

brunoffranca:
## What ❔
We now broadcast the Replica Prepare message to all replicas. Replicas then use it to update their high Commit QCs and change views id necessary.

## Why ❔
The high Commit QC broadcast should be independent from the rest of the algorithm, so that view synchronization is independent from the bft logic - it simplifies reasoning a lot and prevents potential deadlocks that our implementation currently has

### DIFF
--- a/node/actors/bft/src/lib.rs
+++ b/node/actors/bft/src/lib.rs
@@ -19,7 +19,8 @@ use crate::io::{InputMessage, OutputMessage};
 use anyhow::Context;
 pub use config::Config;
 use std::sync::Arc;
-use zksync_concurrency::{ctx, scope};
+use zksync_concurrency::{ctx, oneshot, scope};
+use zksync_consensus_network::io::ConsensusReq;
 use zksync_consensus_roles::validator;
 use zksync_consensus_utils::pipe::ActorPipe;
 
@@ -93,7 +94,19 @@ impl Config {
                 let InputMessage::Network(req) = pipe.recv.recv(ctx).await?;
                 use validator::ConsensusMsg as M;
                 match &req.msg.msg {
-                    M::ReplicaPrepare(_) | M::ReplicaCommit(_) => leader_send.send(req),
+                    M::ReplicaPrepare(_) => {
+                        // This is a hacky way to do a clone. This is necessary since we don't want to derive
+                        // Clone for ConsensusReq. When we change to ChonkyBFT this will be removed anyway.
+                        let (ack, _) = oneshot::channel();
+                        let new_req = ConsensusReq {
+                            msg: req.msg.clone(),
+                            ack,
+                        };
+
+                        replica_send.send(new_req);
+                        leader_send.send(req);
+                    }
+                    M::ReplicaCommit(_) => leader_send.send(req),
                     M::LeaderPrepare(_) | M::LeaderCommit(_) => replica_send.send(req),
                 }
             }
--- a/node/actors/bft/src/replica/replica_prepare.rs
+++ b/node/actors/bft/src/replica/replica_prepare.rs
@@ -2,7 +2,7 @@
 use super::StateMachine;
 use tracing::instrument;
 use zksync_concurrency::{ctx, error::Wrap};
-use zksync_consensus_roles::validator::{self};
+use zksync_consensus_roles::validator;
 
 /// Errors that can occur when processing a "replica prepare" message.
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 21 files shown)


node/actors/bft/src/replica/replica_prepare.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        if qc_view > self.high_qc.clone().map(|x| x.view().number) {
+            self.high_qc.clone_from(&message.high_qc)
+        }
+
+        // Skip to a new view, if necessary.
+        if qc_view > Some(self.view) {
+            self.view = qc_view.unwrap();
+            self.start_new_view(ctx).await.wrap("start_new_view()")?;
+        }
```
brunoffranca, line 98: Technically this should be `qc_view >= Some(self.view)`, but for some reason the bft tests with mock network become real flaky when I do that. Since this is a temporary solution, maybe this is enough and debugging the mock tests is unnecessary?

node/actors/bft/src/replica/replica_prepare.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl StateMachine {
+    #[instrument(level = "trace", skip(self), ret)]
+    pub(crate) async fn process_replica_prepare(
+        &mut self,
+        ctx: &ctx::Ctx,
+        signed_message: validator::Signed<validator::ReplicaPrepare>,
+    ) -> Result<(), Error> {
+        // ----------- Checking origin of the message --------------
```
pompon0, line 54: I think we can simplify handling this message to just extracting the high_qc and verifying it, while skipping everything else.

node/actors/bft/src/lib.rs
```diff
@@ -93,7 +94,19 @@ impl Config {
                 let InputMessage::Network(req) = pipe.recv.recv(ctx).await?;
                 use validator::ConsensusMsg as M;
                 match &req.msg.msg {
-                    M::ReplicaPrepare(_) | M::ReplicaCommit(_) => leader_send.send(req),
+                    M::ReplicaPrepare(_) => {
+                        // This is a hacky way to do a clone. This is necessary since we don't want to derive
+                        // Clone for ConsensusReq. When we change to ChonkyBFT this will be removed anyway.
```
pompon0, line 99: this is hacky indeed. The right way would be to assign a fresh ack channel to each of the copies and spawn a background task to wait for both of them to get processed, and then ack the received channel.

pompon0 (changes requested):
the output gets spammed by WARN log about "leader receiving prepare when it is not a leader for the view" and about "receiving a message for the past view". This is unacceptably verbose, because both of those events is a common and expected occurrence. Please adjust the logs appropriately.

node/actors/bft/src/replica/replica_prepare.rs
pompon0: you can skip verifying aggregated signatures, if high_qc is useless anyway.

pompon0: also mock network doesn't have p2p block fetching, therefore this flakiness might only appear on mock network tests (we might want to improve that actually, like using a shared block store).

node/actors/bft/src/replica/replica_prepare.rs
brunoffranca: Done.

node/actors/bft/src/replica/replica_prepare.rs
brunoffranca: The problem was not trying to finalize the block. Since mock network doesn't have block fetching, replicas would skip to a higher view but would never receive the previous blocks thus stalling. It's working now.

node/actors/bft/src/lib.rs
brunoffranca: Agree that it's hacky, but shouldn't cause any problems, correct?

pompon0: it shouldn't break anything, but it escapes inflight request limiting

brunoffranca: We only have this until ChonkyBFT, so it will only be for PoA single leader.

node/actors/bft/src/replica/replica_prepare.rs
pompon0: nit: I meant literally all validation, except for CommitQC.verify, and only after checking that this CommitQC has actually higher view than our high_qc, so that the signature validation is skipped entirely otherwise.

node/actors/bft/src/replica/replica_prepare.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        // ----------- All checks finished. Now we process the message. --------------
+
+        let qc_view = high_qc.view().number;
+
+        // Try to create a finalized block with this CommitQC and our block proposal cache.
+        self.save_block(ctx, &high_qc).await.wrap("save_block()")?;
+
+        // Update our high QC, if necessary.
+        if Some(qc_view) > self.high_qc.clone().map(|x| x.view().number) {
```
pompon0, line 97: save_block is responsible for updating the high_qc

pompon0 (approved):

PR #120, opened 2024-05-27
Fix ambiguous reproposal bug (BFT-454)

brunoffranca:
## What ❔
Now if there are >1 subquorums in the PrepareQC, leader will create a new proposal and replicas will require a new proposal.

## Why ❔
It is possible to have 2 subquorums: vote A and vote B, each with >2f weight, in a single PrepareQC (even in the unweighted case, because QC contains n-f signatures, not 4f+1). In such a situation from the POV of the BFT algorithm, leader is eligible to do any of the following:

- reproposal for A
- reproposal for B
- a new proposal

Since the choice here is ambiguous, this can break liveness.

### DIFF
--- a/node/libs/roles/src/validator/messages/leader_prepare.rs
+++ b/node/libs/roles/src/validator/messages/leader_prepare.rs
@@ -90,10 +90,10 @@ impl PrepareQC {
         let mut high_votes: Vec<_> = count.into_iter().filter(|x| x.1 >= min).collect();
 
         if high_votes.len() == 1 {
-            return high_votes.pop().map(|x| x.0);
+            high_votes.pop().map(|x| x.0)
         } else {
-            return None;
-        };
+            None
+        }
     }
 
     /// Get the highest CommitQC.
--- a/node/libs/roles/src/validator/messages/leader_prepare.rs
+++ b/node/libs/roles/src/validator/messages/leader_prepare.rs
@@ -75,6 +75,9 @@ impl PrepareQC {
 
     /// Get the highest block voted and check if there's a quorum of votes for it. To have a quorum
     /// in this situation, we require 2*f+1 votes, where f is the maximum number of faulty replicas.
+    /// Note that it is possible to have 2 quorums: vote A and vote B, each with >2f weight, in a single
+    /// PrepareQC (even in the unweighted case, because QC contains n-f signatures, not 4f+1). In such a
+    /// situation we say that there is no high vote.
     pub fn high_vote(&self, genesis: &Genesis) -> Option<BlockHeader> {
         let mut count: HashMap<_, u64> = HashMap::new();
         for (msg, signers) in &self.map {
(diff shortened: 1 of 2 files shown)


aakoshh (commented):
This looks correctly implemented in line with the new specification. Since it's part of the spec, I'd expect there to be some unit test to enshrine it and prevent any regressions. 

Just to confirm my understanding, here's what I believe this mechanism achieves:
* If there is exactly 1 subquorum of 2f+1, then there could have been a Commit QC for the proposal in the earlier round, and we _must_ repropose that value, otherwise the safety property is violated. Or vice versa if there _was_ a Commit QC worth of votes in the previous round, there _will_ be exactly one subquorum in the next round.
* If there are 2 subquorums of 2f+1, then there could not have been a Commit QC and we are safe to propose a new value, or to repropose one of them. These subquorums can be from the same view, or different views. 
* As a coincidental side effect it can make the consensus algorithm more efficient by not requiring the broadcast of a full payload again, but this is not an explicit goal.

As another side note, I think this only works as long as the quorum size is n-f, not 4f+1 as it was in the spec. As a counter example, say n=100, which gives f=(100-1)/5=19, and we wanted to use 4f+1=77 as quorum size; say we have a QC over block A signed by 77 validators, of which 58 are honest, 19 dishonest, but the adversary prevented these messages from being delivered, and everybody timed out; furthermore, the dishonest leader presented block B to the other 100-77=23 honest validators, who cast their vote on it. In the next round, another dishonest leader can propose a new block C by presenting a two subquorums of size 2f+1=39: 1) over B by 23+19=42 validators, and 2) over A by choosing 39 of the 58 signatures. This gives 39+42=82 signatures which is enough for a timeout certificate, and now C is rightfully proposed because we have 2 subquorums. It doesn't matter if we repropose or not as we'd never know which one had an actual QC. But with q=n-f this doesn't work, because at most f honest nodes can be made to sign an alternative proposal, so if there is a QC for something, it's true that at least 2f+1 of those votes will be present in the next timeout certificate, and no more than 2f can support anything else.

aakoshh:
Perhaps could you re-state your reasoning for _not_ reproposing any of the subquorums, just for posterity? For example the one with the higher view? I read that the validator would like to be the proposer and not the reproposer to reap more rewards, but why shouldn't the others try to enforce a reproposal to make it more efficient?

pompon0 (approved):
lgtm, please add a unit test.

brunoffranca:
@aakoshh There's not a strong reason either way. We just needed to resolve the ambiguity.

aakoshh:
I remember now why I was focussing so much on the optimisation aspect of re-proposals. It's because the ChonkyBFT spec [highlights this](https://github.com/dnkolegov/chonkybft-spec/blob/9a5a72d52fe1e8fd917eebc4d74adc6ec00af12b/spec/replica.rs#L108-L112) and suggests this is a speed up.

PR #121, opened 2024-05-31
BFT-457: Use ECDSA key for attesters

aakoshh:
## What ❔

Adds `secp256k1` keys and signatures and uses it in the `attester` role instead of `bn254`. 
- [x] Add `secp256k1::{SecretKey, PublicKey, Signature, AggregateSignature}`
- [x] Replace `bn254` with `secp256k1` in the `attester`
- [x] Write unit tests for serialization roundtrips
- [x] Write unit tests for signing and verifying messages
- [x] Get some test vectors for Ethereum to ensure that the signatures are compatible with the Solidity semantics (e.g. shifting the recovery ID by 27; see [ECDSA.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol) for reference).
- [x] Investigate failing genesis and other tests
- [x] Remove `bn254` - deferred to https://github.com/matter-labs/era-consensus/pull/123
- [x] Investigate failing Twins test

#### Inefficient aggregate signature verification

The aggregate signature is just a list of bytes, but it mimics the API of the `bn256` version which really supported aggregation and was insensitive to message and key ordering during verification. To fulfil that expectation it has to match every message against every signature until a matching pair is found, which can result in repeated futile signature verifications. 

The API also supports every member signing different messages, which was utilised to aggregate the signatures of validators into a quorum certificate even when they all signed different justifications. In practice we have attesters signing the same batch, so we could get rid of this option. If we did, then then aggregate signature in the batch could become a `Vec<(PublicKey, Signature)>`, which would be less costly to verify in Solidity than what is currently happening in Rust (matching every signatory against every possible signature), although a final Solidity payload transformation can be done by the L1 batch submitter as well. 

On a similar note, the seralized signature aims to be compatible with Solidity by shifting the recovery ID by 27, but this could also be deferred for later, to keep this closer to being standard.

#### Change in protobuf

~~I did not rename the `bn254` field in the protobuf format to `secp256k1`, just copied the same comment that the field name is wrong from the validator role proto file where we have a similar situation. I'll leave it up to you if breaking the JSON schema is worse than misleading names.~~

As requested, a new field has been added. 

### Unrelated test failures

I encountered some test failures that seemingly have nothing to do with my changes. I reckon the reason is that a different number of calls are made to `rng`, which affects subsequently random values. For example during the generation of `Genesis` the new attester keys generation, as a side effect, changes the leader schedule which is generated a few lines down from it, which then causes a test to fail due to the value of the schedule, not the attesters. Notably this happened to tests which have no repetition, ie. they only exercise one random sample.

We discussed once that testing with a deterministic but slowly changing random seed could surface such test failures on CI, perhaps on a nightly run, so as not to disrupt people's workflow. This was considered undesireable as it 1) might provide a false sense of confidence to some test authors that coverage will increase with time, enticing them to write lower quality tests and 2) annoy fellow developers when it fails on CI. Nevertheless this is an example that even with a static seed, unrelated test failures can pop up as we evolve the code. One could even argue that the static seed gives test authors a false sense of "mission accomplished".

#### Twins 

One of the tests that failed was the twins network. It generated a scenario in which node A could not gossip a block 98 to node B, and node B got blocked indefinitely waiting for block 98 to appear in its store. Since the setup dictated that every node has only 1 gossip peer, no other node tried to send block 98 to B (or vice versa B would not have tried to get it from anybody else). For some reason A didn't have the block and possibly was also blocked waiting for it; it didn't seem to participate actively in the rounds. I'm not sure why nobody gossiped the block to A, but what is sure is that the current mock gossip mechanism in the test would be insufficient to keep things going even if they did, because a gossip from A to B is only triggered once, when A tries to send a message to B. 

As a seemingly sensible workaround I increased the number of gossip peers to be larger than the amount of twins in the network, ie. more than the faulty replicas, which seems like a good policy anyway. 

It works now but we could potentially turn the gossip to be more independent from messaging than it currently is. Or a simpler alternative would be to asynchronously wait until the source has the block, and then send it to the target, rather than give up immediately.

## Why ❔

So that we can send batch signatures to L1 and verify them in smart contracts, where the Secp256k1 curve is readily supported by a precompile.


node/libs/roles/src/proto/attester.proto
```diff
@@ -28,10 +28,12 @@ message Signed {
 }
 
 message PublicKey {
+  // The name is wrong, it should be secp256k1.
   optional bytes bn254 = 1; // required
 }
 
 message Signature {
+  // The name is wrong, it should be secp256k1.
```
pompon0, line 36: just add a new field and deprecate the old one. Since it is not deployed anywhere yet, we don't need to maintain support for bn254.

node/actors/bft/src/testonly/run.rs
```diff
@@ -359,6 +359,7 @@ async fn twins_receive_loop(
         // Send after taking note of potentially gossipable blocks.
         let send = |msg| {
             if let Some(number) = block_to_gossip(target_port, &msg) {
+                tracing::info!("   ~~~ gossip from={port} to={target_port} number={number}");
```
pompon0, line 362: nit: you want these logs to be merged?

node/libs/crypto/src/secp256k1/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// not rely on message and key ordering; it supported different messages from each signatory as well.
+    ///
+    /// Here we have a simple list of signatures, and we cannot rely on them having been added in the same
+    /// order in which they are going to be verified, therefore we have to try every message against every
+    /// signature.
+    ///
+    /// The method assumes that there are no repeated pairs in the input, ie. that every signature is used exactly once.
+    pub fn verify_hash<'a>(
+        &self,
+        hashes_and_pks: impl Iterator<Item = (&'a [u8], &'a PublicKey)>,
```
pompon0, line 202: Is this function directly called from outside the crate? If it is, consider exposing a non-generic version of this function instead - we have compilation optimizations enabled for the zksync_consensus_crypto crate, even in debug builds - if this function gets inlined within other crate the optimizations won't get applied.

node/libs/crypto/src/secp256k1/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+                verified.len()
+            );
+        }
+        Ok(())
+    }
+
+    /// Verify messages after hashing them with Keccak256
+    pub fn verify<'a>(
+        &self,
+        msgs_and_pks: impl Iterator<Item = (&'a [u8], &'a PublicKey)>,
```
pompon0, line 232: ditto

node/libs/crypto/src/secp256k1/testonly.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.0.fill_bytes(dest)
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand::Error> {
+        self.0.try_fill_bytes(dest)
+    }
+}
```
pompon0, line 28: do we need these? For bn254 we needed them, because it was using an older version of rand.

node/libs/roles/src/attester/keys/secret_key.rs
```diff
@@ -36,7 +36,11 @@ impl SecretKey {
 
     /// Sign a message hash.
     pub fn sign_hash(&self, msg_hash: &MsgHash) -> Signature {
-        Signature(self.0.sign(&ByteFmt::encode(msg_hash)))
+        let sig = self
+            .0
+            .sign_hash(&ByteFmt::encode(msg_hash))
+            .expect("message hash is sufficient in size");
```
pompon0, line 42: you mean "has insufficient size"?

node/libs/roles/src/proto/attester.proto
```diff
@@ -41,6 +43,9 @@ message WeightedAttester {
 }
 
 message AggregateSignature {
+  // The name is wrong, it should be secp256k1.
+  // This is aggregate signature itself encoded as bytes, which is why it's not `repeated`,
+  // even though for secp256k1 it's just a list of signatures.
```
pompon0, line 48: ditto

node/libs/roles/src/validator/testonly.rs
```diff
@@ -288,7 +288,24 @@ impl Distribution<GenesisRaw> for Standard {
             validators: rng.gen(),
             attesters: rng.gen(),
             leader_selection: rng.gen(),
+        };
+
+        // In order for the genesis to be valid, sticky/rota leaders need to be in the validator committee.
+        if let LeaderSelectionMode::Sticky(_) = genesis.leader_selection {
+            let i = rng.gen_range(0..genesis.validators.len());
+            genesis.leader_selection =
+                LeaderSelectionMode::Sticky(genesis.validators.get(i).unwrap().key.clone());
```
pompon0, line 297: I assume this caused some encoding test to fail. Please add more iterations for this test to get better coverage.

node/actors/bft/src/testonly/run.rs
aakoshh: Yes they were helpful to remind what was happening and that gossiping happens from only a select few.

node/libs/crypto/src/secp256k1/mod.rs
aakoshh: NB this is how it is in [bn254](https://github.com/matter-labs/era-consensus/blob/90b893ceead1aabff34713a152376fffbf2b9e0a/node/libs/crypto/src/bn254/mod.rs#L281-L283)

node/libs/crypto/src/secp256k1/testonly.rs
aakoshh: `k256::SecretKey::random` required `CryptoRngCore`

node/libs/roles/src/attester/keys/secret_key.rs
aakoshh: I mean I know it has sufficient size, which is why I expect that this will not fail, but I see how it is confusing, will change it.

node/libs/crypto/src/secp256k1/mod.rs
aakoshh: It is called, yes. It would be awkward if it's a public type and it couldn't be called. I made it non-generic the same way as bn254 is.

node/libs/crypto/src/secp256k1/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+                if verified[j] {
+                    continue 'signatures;
+                }
+                if sig.verify_hash(hash, pk).is_ok() {
+                    verified[j] = true;
+                    continue 'inputs;
+                }
+            }
+            bail!("failed to verify message {i} against any of the signatures in the aggregate");
+        }
```
brunoffranca, line 226: Indeed, this does not work. We can't have quadratic work verifying this. Just change to accept a single message. It's enough for signing L1 batches.

brunoffranca (changes requested):

aakoshh:
As a side note I'm not convinced that [verifying Genesis in its ProtoFmt](https://github.com/matter-labs/era-consensus/blob/90b893ceead1aabff34713a152376fffbf2b9e0a/node/libs/roles/src/validator/conv.rs#L73) is a good idea. Given how loose protobuf generally is, with most struct fields being optional, I'd be more comfortable if `ProtoFmt` only focussed on data format. By mixing it with domain model verification, it precludes us from even looking at what's wrong with it (well, we can inspect the proto DTO itself), and one cannot even be sure without looking at the implementation whether verification happened or not. For most types it doesn't happen. Genesis is an outlier in that it _can_ be verified on its own right, but others that for example require genesis itself to be verified cannot have this luxury.

This is an example where further static typing could help: this particular `ProtoFmt` could return a `Verified<Genesis>`, while others could return `Unverified<Self>` or something similar, to make us aware what we are dealing with.

node/actors/network/src/rpc/tests.rs
```diff
@@ -25,12 +25,14 @@ fn test_capability_rpc_correspondence() {
...
 fn test_schema_encode_decode() {
     let rng = &mut ctx::test_root(&ctx::RealClock).rng();
-    test_encode_random::<consensus::Req>(rng);
-    test_encode_random::<consensus::Resp>(rng);
-    test_encode_random::<push_validator_addrs::Req>(rng);
-    test_encode_random::<push_block_store_state::Req>(rng);
-    test_encode_random::<get_block::Req>(rng);
-    test_encode_random::<get_block::Resp>(rng);
+    for _ in 0..10 {
+        test_encode_random::<consensus::Req>(rng);
```
pompon0, line 29: nit: we can get it one step further and put the loop inside test_encode_random.

node/libs/crypto/src/secp256k1/mod.rs
pompon0: oh right. As long as we make the generic call the non-generic version, we are good.

node/actors/network/src/rpc/tests.rs
aakoshh: Argh, I didn't even check this was the same method, thanks

pompon0 (approved):

brunoffranca:
Let's merge this one first and then we'll merge the other PRs.

PR #122, opened 2024-06-10
Added PayloadVersion to the genesis.

pompon0:
It will allow us to upgrade the payload encoding in zksync-era.
This is just one solution, alternatives are:
1. allow to actually include custom data in consensus Genesis, not just a version number.
2. same as 1, except we include just a hash of the custom data.
3. bump the zksync-era protocol version instead. That might be even preferrable, however I don't know whether we can afford bumping protocol version just for that.


brunoffranca:
Option 3 would indeed be preferable. Why do you think we can't bump protocol version for this? And I can't imagine that the encoding will change that often.

brunoffranca (approved):

pompon0:
we will go with zksync-era ProtocolVersion instead

PR #123, opened 2024-06-10
BFT-457: Remove bn254

aakoshh:
## What ❔

Removes the `bn254` crypto module. It's a separate PR so the code is easy to restore, should we need this module again.

## Why ❔

Because https://github.com/matter-labs/era-consensus/pull/121 introduces ECDSA for attester signatures and bn254 will not be used for anything any more.


brunoffranca (approved):
Looks good. Let's merge after the other PR.

pompon0 (commented):
If you need my review, please merge main into this branch first.

PR #124, opened 2024-06-11
Http debug page (BFT-259, BFT-451)

ElFantasma:
## What ❔
Added an HTTP endpoint with basic debug information:
- It is disabled by default. To enable it, a line defining it's address in config is required:
`  "debugAddr": "<IP Address>:<port>"`
- Basic authentication implemented, also optional. Configured by
`  "debugCredentials": "<user>:<password>"`
in config file. No authentication required if that line is not present.
- TLS encryption: 
certificate and key files are required in case authentication is enabled. Default names are `cert.pem`and `key.pem`but file paths can be specified in config.json:
```
  "debugCertPath": "<cert path>",
  "debugKeyPath": "<key path>",
```
- Current look:
image
 

## Why ❔

Requested at:
- [BFT-259 - implement an http debug page for a node (with network data)](https://linear.app/matterlabs/issue/BFT-259/implement-an-http-debug-page-for-a-node-with-network-data)
- [BFT-451 - Do not expose HTTP servers for metrics and debugging if possible](https://linear.app/matterlabs/issue/BFT-451/do-not-expose-http-servers-for-metrics-and-debugging-if-possible)


dnkolegov:
> ## What ❔
> Added an HTTP endpoint with basic debug information:
> 
> * It is disabled by default. To enable it, a line defining it's address in config is required:
>   `  "debugAddr": "<IP Address>:<port>"`
> * Basic authentication implemented, also optional. Configured by
>   `  "debugCredentials": "<user>:<password>"`
>   in config file. No authentication required if that line is not present.
> * TLS encryption:
>   `local.cert` and `local.key` files are required.

Is it possible to enable Basic authentication without TLS?

ElFantasma:
> > ## What ❔
> > Added an HTTP endpoint with basic debug information:
> > 
> > * It is disabled by default. To enable it, a line defining it's address in config is required:
> >   `  "debugAddr": "<IP Address>:<port>"`
> > * Basic authentication implemented, also optional. Configured by
> >   `  "debugCredentials": "<user>:<password>"`
> >   in config file. No authentication required if that line is not present.
> > * TLS encryption:
> >   `local.cert` and `local.key` files are required.
> 
> Is it possible to enable Basic authentication without TLS?

It is not possible with current state of the code. Is that something desirable?

dnkolegov:
> > 
> > Is it possible to enable Basic authentication without TLS?
> 
> It is not possible with current state of the code. Is that something desirable?

No. It is very good. It should not be allowed to use basic authentication without TLS.

ElFantasma:
I finally managed to get the server to work with our ctx in case of cancellation, but still have my doubts that this is the best way to solve it.
In any case I'm setting the PR as ready for review, while I'll keep cleaning up a bit some stuff. In particular I'll work in making the cert and private key paths configurable

node/libs/concurrency/src/net/http/mod.rs
```diff
@@ -0,0 +1,46 @@
+//! Http Server configuration structs
```
pompon0, line 1: http has nothing to do with the concurrency crate, please move it to zksync_consensus_network, or to zksync_consensus_utils if that is not feasible

node/libs/concurrency/src/net/http/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::{net::SocketAddr, path::PathBuf};
+
+use anyhow::{Context, Result};
```
pompon0, line 4: nit: don't import Result, use `anyhow::Result` directly

node/libs/concurrency/src/net/http/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Debug Page credentials (user:password)
```
pompon0, line 5: don't derive Debug on a struct with secrets. You can implement Debug manually with the secrets redacted instead.

node/libs/concurrency/src/net/http/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct DebugCredentials {
```
pompon0, line 7: DebugPageCredentials? DebugCredentials is ambiguous IMO

node/actors/network/src/metrics.rs
```diff
@@ -170,3 +185,35 @@ impl NetworkGauges {
         }
     }
 }
+
+/// Metrics reported for TCP connections.
+#[derive(Debug)]
+pub struct StreamValues {
```
pompon0, line 191: MeteredStreamStats? StreamValues is very vague IMO

node/actors/network/src/metrics.rs
```diff
@@ -170,3 +185,35 @@ impl NetworkGauges {
...
+    /// Total bytes sent over the Stream.
+    pub sent: AtomicU64,
+    /// Total bytes received over the Stream.
+    pub received: AtomicU64,
+    /// TCP connections established since the process started.
```
pompon0, line 196: outdated comment?

node/actors/network/src/metrics.rs
```diff
@@ -170,3 +185,35 @@ impl NetworkGauges {
...
+    pub established: SystemTime,
+    /// Ip Address and port of current connection.
+    pub address: Option<SocketAddr>,
```
pompon0, line 199: why is it optional?

node/actors/network/src/metrics.rs
```diff
@@ -170,3 +185,35 @@ impl NetworkGauges {
...
+/// Metrics reported for TCP connections.
+#[derive(Debug)]
+pub struct StreamValues {
+    /// Total bytes sent over the Stream.
+    pub sent: AtomicU64,
+    /// Total bytes received over the Stream.
+    pub received: AtomicU64,
+    /// TCP connections established since the process started.
+    pub established: SystemTime,
+    /// Ip Address and port of current connection.
```
pompon0, line 198: local addr? peer addr?

node/actors/network/src/metrics.rs
```diff
@@ -17,6 +21,8 @@ use zksync_concurrency::{ctx, io, net};
 pub(crate) struct MeteredStream {
     #[pin]
     stream: net::tcp::Stream,
+    // Collects values to be shown on the Debug http page
+    values: Arc<StreamValues>,
```
pompon0, line 25: stats?

node/actors/network/src/metrics.rs
```diff
@@ -17,6 +21,8 @@ use zksync_concurrency::{ctx, io, net};
 pub(crate) struct MeteredStream {
     #[pin]
     stream: net::tcp::Stream,
+    // Collects values to be shown on the Debug http page
```
pompon0, line 24: ///

node/tools/src/bin/deployer.rs
```diff
@@ -51,6 +51,10 @@ fn generate_consensus_nodes(nodes: usize, seed_nodes_amount: Option<usize>) -> V
                 gossip_dynamic_inbound_limit: 2,
                 gossip_static_inbound: [].into(),
                 gossip_static_outbound: [].into(),
+                debug_addr: None,
```
pompon0, line 54: please combine these fields into a single message

node/libs/concurrency/src/net/http/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/// Http debug page configuration.
+#[derive(Debug, Clone)]
+pub struct DebugPageConfig {
+    /// Public Http address to listen incoming http requests.
+    pub addr: SocketAddr,
+    /// Debug page credentials.
+    pub credentials: Option<DebugCredentials>,
+    /// Cert file path
```
pompon0, line 42: why paths? Can we just keep them loaded in memory instead?

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+use tokio_rustls::{
+    rustls::{
+        pki_types::{CertificateDer, PrivateKeyDer},
+        ServerConfig,
+    },
+    TlsAcceptor,
+};
+use zksync_concurrency::net::http::{DebugCredentials, DebugPageConfig};
+
+use crate::{consensus, ctx, scope, Network, StreamValues};
```
pompon0, line 33: import ctx and scope from the concurrency crate.

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+
+const STYLE: &str = include_str!("style.css");
+
+/// Http Server.
+pub struct Server {
+    addr: SocketAddr,
+    credentials: Option<DebugCredentials>,
```
pompon0, line 40: document why it is optional

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    cert_path: PathBuf,
+    key_path: PathBuf,
+    network: Arc<Network>,
+}
+
+impl Server {
+    /// Creates a new Server
+    pub fn new(config: DebugPageConfig, network: Arc<Network>) -> Server {
+        Server {
+            addr: config.addr,
```
pompon0, line 50: if you just embed the fields of configs, it would be cleaner to have a `config` field in the Server instead.

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            // if this signal completes, start shutdown
+            let mut cancel_signal = std::pin::pin!(async {ctx.canceled().await});
+
+            let mut listener =
+                TlsListener::new(self.tls_acceptor(), TcpListener::bind(self.addr).await?);
+
+            let http = http1::Builder::new();
+
+            // Start a loop to continuously accept incoming connections
+            loop {
```
pompon0, line 74:
```
while Ok(res) = ctx.wait(listener.accept()).await {
  let steam = match res {
    Ok((stream,_)) => stream,
    Err(err) => { ...; continue }
  };
  ...
}
graceful.shutdown().await;
```

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    async fn handle(
+        &self,
+        request: Request<hyper::body::Incoming>,
+    ) -> anyhow::Result<Response<Full<Bytes>>> {
+        let mut response = Response::new(Full::default());
+        match self.basic_authentication(request.headers()) {
+            Ok(_) => *response.body_mut() = self.serve(request),
+            Err(e) => {
+                *response.status_mut() = StatusCode::UNAUTHORIZED;
+                *response.body_mut() = Full::new(Bytes::from(format!("{}", e)));
```
pompon0, line 124: `e.to_string()`?

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+                .context("The 'Authorization' header was not a valid UTF8 string.")?;
+            let base64encoded_segment = header_value
+                .strip_prefix("Basic ")
+                .context("The authorization scheme was not 'Basic'.")?;
+            let decoded_bytes = base64::engine::general_purpose::STANDARD
+                .decode(base64encoded_segment)
+                .context("Failed to base64-decode 'Basic' credentials.")?;
+            let incoming_credentials = String::from_utf8(decoded_bytes)
+                .context("The decoded credential string is not valid UTF8.")?;
+            if Into::<String>::into(credentials) == incoming_credentials {
```
pompon0, line 150: `String::from`, but you should prefer comparing structured data instead (not strings).

node/actors/network/src/http/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+        let cert_der = load_certs(&self.cert_path).expect("Invalid certificate");
+        let key_der = load_private_key(&self.key_path).expect("Invalid private key");
+        Arc::new(
+            ServerConfig::builder()
+                .with_no_client_auth()
+                .with_single_cert(cert_der, key_der)
+                .unwrap(),
+        )
+        .into()
+    }
```
pompon0, line 257: these should have been loaded to the config directly and if possible verified upfront.

node/actors/network/src/http/mod.rs
ElFantasma: Wow, nice.

node/actors/executor/src/lib.rs
```diff
@@ -102,6 +98,21 @@ impl Executor {
...
 
+    /// Extracts a debug page config from http crate.
+    fn debug_page_config(&self) -> Option<http::DebugPageConfig> {
+        self.config
+            .debug_page_config
+            .as_ref()
+            .map(|debug_page_config| http::DebugPageConfig {
+                addr: debug_page_config.addr,
+                credentials: debug_page_config.credentials.clone(),
+                certs: http::load_certs(&debug_page_config.cert_path)
```
pompon0, line 109: that better, but please move loading and parsing even further away, so that executor doesn't need access to the file system

node/actors/executor/src/lib.rs
```diff
@@ -113,17 +124,17 @@ impl Executor {
...
             let (net, runner) =
                 network::Network::new(network_config, self.block_store.clone(), network_actor_pipe);
             net.register_metrics();
             s.spawn(async { runner.run(ctx).await.context("Network stopped") });
 
-            if let Some(debug_config) = &self.config.debug_page {
-                s.spawn(async {
-                    http::Server::new(debug_config.clone(), net)
+            if let Some(debug_config) = self.debug_page_config() {
+                s.spawn(async move {
```
pompon0, line 136: don't use "async move" - for moving values inside the closures, use "let x = x;" syntax, or ctx::NoCopy wrapper for passing Copyable objects.

node/actors/executor/src/lib.rs
```diff
@@ -57,7 +53,7 @@ pub struct Config {
     pub gossip_static_outbound: HashMap<node::PublicKey, net::Host>,
     /// Http debug page configuration.
     /// If None, debug page is disabled
-    pub debug_page: Option<DebugPageConfig>,
+    pub debug_page_config: Option<DebugPageConfig>,
```
pompon0, line 56: I'm not sure if I didn't say the opposite in the previous review round (sry, if that's been the case), but "_config" here is redundant, since it is a field of the Config struct.

node/actors/network/src/http/mod.rs
```diff
@@ -61,53 +66,40 @@ impl Server {
...
-                                tracing::error!("Error accepting connection: {}", err);
+            // Start a loop to accept incoming connections
+            while let Ok(res) = ctx.wait(listener.accept()).await {
+                match res {
+                    Ok((stream, _)) => {
+                        let io = TokioIo::new(stream);
+                        let conn = http.serve_connection(io, service_fn(|req| self.handle(req)));
+                        // watch this connection
+                        let fut = graceful.watch(conn);
+                        s.spawn_bg(async move {
```
pompon0, line 84: ditto, avoid async move

node/tools/src/config.rs
```diff
@@ -103,9 +100,7 @@ pub struct AppConfig {
     pub gossip_static_inbound: HashSet<node::PublicKey>,
     pub gossip_static_outbound: HashMap<node::PublicKey, net::Host>,
 
-    pub debug_credentials: Option<DebugCredentials>,
-    pub debug_cert_path: Option<PathBuf>,
-    pub debug_key_path: Option<PathBuf>,
+    pub debug_page_config: Option<DebugPageConfig>,
```
pompon0, line 103: ditto, "_config" is redundant here

pompon0 (approved):

brunoffranca (approved):

PR #125, opened 2024-06-12
BFT-465: Phase level partitioning

aakoshh:
## What ❔

Changes the `ScenarioGenerator` to create multiple splits per view and apply different partitions in different phases. 

I ended up with 2 phases instead of 4, otherwise the tests time out: `ReplicaPrepare`, `LeaderPrepare` and `ReplicaCommit` are considered phase 0 and then `LeaderCommit` is phase 1. 

## Why ❔

Because there are certain scenarios that the current granularity could not produce. For example consider the following attempt at trying to capture a security violation:

Say we have 6 nodes, 2 of which (node 1 and 2) are dishonest and colluding (which is more than the maximum 1 tolerable faulty replica with only 6 nodes). 
1. In view 1 node 1 is leader and proposes block A to {1,2,3,4,5} and block B to {1,2,6}. Block A has a QC but node 1 does not disseminate the LeaderCommit. Block B has 3 signatures. 
2. In view 2 node 2 is leader and proposes block C to everyone, with the following justifications from the previous view: {A: [3,4,5], B: [1,2,6]}. So there are 2 high votes and the algorithm says it can propose a new block. This would be impossible if we only had 1 dishonest node, but we have 2. C gathers QC. 
3. Now node 1 can present two alternative finalised blocks at the same height: A and C, which violates the safety property. 

How do we model this with the current Twins setup?

1. We create nodes [1,2,3,4,5,6,1',2'], ie. two twins, but otherwise all honest nodes. 
2. In view 1 we partition the nodes into P1={1,2,3,4,5} and P2={1',2',6}. The leaders are [1, 1'] and they propose block A and B in P1 and P2, respectively. 
3. In view 2 we would partition nodes into P3={1',2',6,3,4,5} and P4={1,2}, so that we have the right number of votes for A and B in P3 to allow a new proposal. Alas, it wouldn't work because in step 2 all members of P1 already got the LeaderCommit and have the HighQC for block A, they finalized it, and thus the leader of view 2 will have it as well. 

With the new change we can re-partition the nodes during view 1 to allow a scenario where node 1 has all the ReplicaCommit messages but fails to deliver the LeaderCommit to anybody. 

### In the paper

I missed it during the implementation, but this is possible in the paper as well, for example:

> **View 1**: Initialize D and D′ with different inputs v1 and v2.
> * **Create the partitions** P1 = {D, E, G}, P2 = {D′, F }.
> * Let D and D′ run as leaders for one round. D proposes v1 to P1 and gathers votes from P1 creating QC(v1). D′ proposes v2 to P2 and gathers votes but not a QC. 
> * **Create the following partitions**: P1 = {D, E}, P2 = {D′, F }, P3 = {G}. D broadcasts QC(v1), which only reaches P1 i.e., (D, E).
>
> **View 2**: Drop all proposals from D and D′ until View 2 starts.
> * Remove all partitions, i.e., P = {D, D′, E, F , G}.
> ...

The Python example I found seems to have done something slightly different: they consider [permutations of message types](https://github.com/druuuu/DiemBFT-Twins/blob/a08e6ad328431962abeb4a08e3972a1316dd4560/src/ScenarioGenerator.py#L9-L12) which I assume govern which messages can make through within a view.


pompon0 (approved):

brunoffranca (approved):

aakoshh:
@brunoffranca I added a test to check that the tests fail when we run them with more twins than the maximum faulty number. I managed to make it fail with n=5 and twins=1; with n=6 and twins=2 it timed out. 

I looked at the logs to see how exactly it fails and it's this new mechanism that makes it happen: 
1. one of the twins is the leader in a view but it can't broadcast the LeaderCommit
2. in the next view it's isolated and can't broadcast the ReplicaPrepare
3. the other twin proposes a different block and it gets finalized 

For a short moment I was wondering: why doesn't the other 4 who signed the QC have a high vote to indicate that it should be reproposed?

Then I remembered that for n<6 we have a sub-quorum size of 1 with the 2f+1 formula, which means basically any dissent precludes any reproposals. In this case the other twin which was isolated when the QC was made is enough to have 2 alternative high votes with a signature ratio of 4:1, so it can propose something new. 

Then I also remembered that you wanted to change this to sq=n-3f, which in this case would be 5, and that also prevents a reproposal. 

Finally the formula I [proposed](https://github.com/matter-labs/era-consensus/pull/116#discussion_r1615922354), sq=2/5n+1, would give 3, which would allow a reproposal in this case. 

I actually tried all 3 of these:
```rust
        let min = 2 * genesis.validators.max_faulty_weight() + 1;        
        let min = genesis.validators.total_weight() - 3 * genesis.validators.max_faulty_weight();
        let min = 2 * genesis.validators.total_weight() / 5 + 1;
```

FWIW with the last version the test does not encounter equivocation. That of course doesn't make it safe with n=5 and f=1, but I thought I'd mention it that it doesn't only affect a percieved efficiency, but is _marginally_ safer, if that makes any sense.

brunoffranca:
Any idea why the n=6 test timed out? I would expect it to also equivocate.

aakoshh:
I can't remember the reason for this timeout exactly as this is a bit atypical in that I'm running 100 scenarios in hope of finding one that fails, and in the case of 5 replicas it was scenario 25 that failed, but with 6 replicas it just doesn't get to the end within the allocated 30 seconds. Just tried now and it took 45 seconds to go through 100 scenarios but it didn't find a counter example.

From earlier logs I can say that when consensus didn't happen for multiple rounds (say, 5 rounds with no finalised block), then I saw the wall clock time between new views be as much as 12 seconds, even with the 10x speedup in the simulated clock time. I suppose with 6 nodes and 2 twins it can play enough havoc that timeouts can make everything go slower in general.

PR #128, opened 2024-06-14
BFT-466: LeaderPrepare wait for previous block deadlock

aakoshh:
## What ❔

- [x] Create a test to artificially create a situation in which some nodes have the proposal payload but not the CommitQC that would finalize it and allow it to be gossiped, while others have the HighQC received as part of a new proposal, but they don't have the previous payload and wait for it to be gossiped to them, creating a deadlock and stalling consensus. 
- [x] The solution is to timeout waiting for the block to appear in the store and move on to the new view, broadcast the HighQC as part of the ReplicaPrepare, which should allow the others to finalise the block.

## Why ❔

To deal with the loss of liveness caused by multiple leaders who owned both the payload and the CommitQC going down, and further replicas stop participating in consensus until they have the previous payload which never comes.

### DIFF
--- a/node/actors/bft/src/leader/state_machine.rs
+++ b/node/actors/bft/src/leader/state_machine.rs
@@ -94,11 +94,16 @@ impl StateMachine {
                         Err(err) => {
                             match err {
                                 super::replica_prepare::Error::Internal(e) => {
+                                    tracing::error!(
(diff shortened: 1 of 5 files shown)


brunoffranca (approved):
Approved!

PR #129, opened 2024-06-15
made tests more deterministic

pompon0:
Some of the tests were flaky because of cpu contention affecting interleavings. This fix makes the passing time (almost) deterministic, by the use of `tokio::time::pause()`. The other consequence is that runtime used in multiple tests has been changed from multi-threaded to single-threaded. This also helps the determinism because the nextest parallelism is bounded by the number of cores - if we make all tests single-threaded, behavior of the test will be the same when executed separately and in a test suite.

Also converted a should_panic test to regular test for consistency.
Tuned size of some of the tests to reasonable execution times.

### DIFF
--- a/node/Cargo.lock
+++ b/node/Cargo.lock
@@ -3825,6 +3825,7 @@ dependencies = [
  "once_cell",
  "pretty_assertions",
  "rand 0.8.5",
+ "test-casing",
  "thiserror",
  "tokio",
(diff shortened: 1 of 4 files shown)


aakoshh (approved):
Nice, `tokio::time::pause` seems to be exactly what I have been looking for 👍

PR #134, opened 2024-06-24
CLIPPY: Explain RwLock around persistent DB

aakoshh:
## What ❔

~~Remove the `RwLock` around the RocksDB instance.~~

Silence the `clippy` warning around the use of an `RwLock` around RocksDB and add some comments as to why it's there.

## Why ❔

`clippy` started complaining in Rust version 1.79 that we don't actually need write access to the `db` variable after acquiring the write lock. AFAIR RocksDB manages its threads internally, and while the instance we have is a `DbCommon<SingleThreaded>`, according to the docs that only means that the column family operations require `&mut` access on the instance, while for `DbCommon<MultiThreaded>` it uses an `RwLock` inside around the column family registry. 

I can't see us using any column family operations, and if we did the compiler would complain about needing write access, so I thought it was safe to remove the `RwLock`. 

On closer inspection it's there to protect an invariant with the in-memory data that shows what the next expected block number is: we take the lock, check that what we're inserting is the expected number, then insert and update the memory; if we didn't take the lock, somebody else could insert, and then we might overwrite something already inserted.

### DIFF
--- a/node/tools/src/store.rs
+++ b/node/tools/src/store.rs
@@ -1,11 +1,7 @@
 //! RocksDB-based implementation of PersistentBlockStore and ReplicaStore.
 use anyhow::Context as _;
 use rocksdb::{Direction, IteratorMode, ReadOptions};
-use std::{
-    fmt,
-    path::Path,
-    sync::{Arc, RwLock},
-};
+use std::{fmt, path::Path, sync::Arc};
 use zksync_concurrency::{ctx, error::Wrap as _, scope, sync};
 use zksync_consensus_roles::validator;
 use zksync_consensus_storage::{BlockStoreState, PersistentBlockStore, ReplicaState, ReplicaStore};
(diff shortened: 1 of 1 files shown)


pompon0:
We don't use rocksdb transactions (which are relatively expensive), so to avoid race conditions in db access, we use an inmemory RwLock instead (only 1 process can access a rocksdb data anyway) - i.e. whenever a consistent reads/read-write transaction is needed, a read/write lock is taken in memory. It could have been better documented though. Also, given that we use rocksdb only for testing, we could switch to rocksdb transactions instead. Alternatively we can just silence this specific warning in this file.

pompon0 (approved):
thank you

aakoshh:
On a similar note, I like using Software Transactional Memory, where a similar pattern arises: the variables in STM can maintain invariants about the state and compose arbitrarily, but at some point we need to persist the data as well, which can be achieved in multiple ways:
1. Atomically check in STM whether the update should happen, update the in-memory if yes, return an error if not; then if the results is successful update the database as well, which now follows the in-memory contents _unless_ it fails to update, in which case the application should handle the inconsistency by restarting.
2. Create database transactions that gets committed or rolled back together with the STM one. 
3. Use an STM queue to write to the database in the background.

PR #137, opened 2024-06-26
BFT-475: Add BatchVotePublisher and expose it to the Executor

aakoshh:
## What ❔

Adds a `BatchVotePublisher` which will be used in the `Executor` to push L1 batch votes to the gossip layer.

## Why ❔

So that we can insert the nodes attestations over L1 batches into vote collection, which [automatically gets gossiped](https://github.com/matter-labs/era-consensus/blob/f696992d77aa788fb228c4068a49f97868d8b531/node/actors/network/src/gossip/runner.rs#L208-L223) to connected peers in `Network::run_stream`.


node/actors/executor/src/lib.rs
```diff
@@ -132,6 +132,12 @@ impl Executor {
             net.register_metrics();
             s.spawn(async { runner.run(ctx).await.context("Network stopped") });
 
+            if let Some(_attester) = self.attester {
+                tracing::info!("Running the node in attester mode.");
+                let _publisher = net.batch_vote_publisher();
+                // TODO: Start a background task that polls the store for new L1 batches and publishes votes.
```
pompon0, line 138: please try to avoid active polling if possible

pompon0 (approved):

PR #138, opened 2024-06-26
FIX: Add missing colon to spec.yaml

aakoshh:
## What ❔

Fixes invalid syntax in `spec.yaml`

## Why ❔

I don't know why it didn't fail on `main` but on my PR after merging `main` into it CI complained about `spec.yaml` having no workflow due to invalid syntax.


dnkolegov (approved):
Thank you Akosh!

PR #144, opened 2024-07-02
Refactored TimeoutQC to contain just one CommitQC

brunoffranca:
## What ❔

Refactored (in the informal spec) `TimeoutQC`, `SignedTimeoutVote` and `TimeoutVote` so that `TimeoutQC` contains only one `CommitQC` (the high QC).

## Why ❔

It reduces the state space which should help with model checking.

### DIFF
--- a/spec/informal-spec/types.rs
+++ b/spec/informal-spec/types.rs
@@ -178,6 +178,7 @@ impl SignedTimeoutVote {
         // If we wish, there are two invariants that are easy to check but aren't required for correctness:
         // self.view() >= self.high_vote.view() && self.high_vote.view() >= self.high_commit_qc_view
         self.vote.high_commit_qc_view == self.high_commit_qc.map(|x| x.view()) && self.verify_sig()
+          && self.high_commit_qc.map(|qc| qc.verify())
     }
 }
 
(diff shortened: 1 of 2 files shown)


spec/informal-spec/types.rs
```diff
@@ -172,9 +175,9 @@ impl SignedTimeoutVote {
     }
 
     fn verify(self) -> bool {
-        // Technically, only the signature needs to be verified. But if we wish, there are two invariants that are easy to check:
-        // self.view() >= self.high_vote.view() && self.high_vote.view() >= self.high_commit_qc.view()
-        self.verify_sig()
+        // If we wish, there are two invariants that are easy to check but aren't required for correctness:
+        // self.view() >= self.high_vote.view() && self.high_vote.view() >= self.high_commit_qc_view
+        self.vote.high_commit_qc_view == self.high_commit_qc.map(|x| x.view()) && self.verify_sig()
```
konnov, line 180: The model checker has just found an omission in the updated spec #146:

```suggestion
        self.vote.high_commit_qc_view == self.high_commit_qc.map(|x| x.view()) && self.verify_sig()
          && self.high_commit_qc.map(|qc| qc.verify())
```

Otherwise, a signed timeout vote by a faulty replica may propagate an invalid `CommitQC`.

konnov (commented):
Adding a few comments after implementing the updates in the Quint specification #146

PR #145, opened 2024-07-04
feat: Collect votes for batch QC and save into store (BFT-477)

aakoshh:
## What ❔

- [x] Added `BatchHash` and `Batch::hash` to give meaning to the attestation
- [x] Refactored `BatchVotes` to look for a quorum over a height+hash combo
- [x] Added a background task to listen to changes in the votes, look for a quorum and save it to the DB
- [x] Test the quorum logic
- [x] Test the propagation

### Caveat

`BatchVotes` still only allows 1 vote per attester and thus doesn't fully handle multiple outstanding L1 batches. That will come later when we'll have subscribed to notifications about the highest finalised batch number, coming from the main node API which doesn't exist yet. That will serve as our beacon for what kind of ranges of votes to accept. Currently for defensive reasons we only accept one vote per attester, which means if attesters keep voting on different heights, always out of sync with each other, they might never form a quorum. That is considered okay at this point; we save what we can and see how it works. 1 minute batches should be long enough not to cause trouble for gossip.

## Why ❔

With these changes the node will react to votes coming in over gossip and forming a quorum over the latest L1 batch.

### DIFF
--- a/node/actors/network/src/gossip/batch_votes.rs
+++ b/node/actors/network/src/gossip/batch_votes.rs
@@ -25,14 +25,17 @@ pub(crate) struct BatchVotes {
     /// the quorum to be reached, but eventually we'll have to allow multiple
     /// votes across different heights.
     pub(crate) votes: im::HashMap<attester::PublicKey, Arc<attester::Signed<Batch>>>,
+
     /// Total weight of votes at different heights and hashes.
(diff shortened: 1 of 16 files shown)


brunoffranca (approved):
Looks good to me!

PR #148, opened 2024-07-05
feat: Batch vote publish loop (BFT-474)

aakoshh:
## What ❔

- [x] Add `BatchStore::unsigned_batch_numbers` to query `BatchNumber`s that still need a QC
- [x] Add `BatchStore::batch_to_sign` to fetch the `Batch` with the correct `BatchHash` that needs to be signed
- [x] Add `PersistentBatchStore::unsigned_batch_numbers` to implement the above
- [x] Add `PersistentBatchStore::batch_to_sign` to implement the above
- [x] Add `AttesterRunner` to periodically query `unsigned_batch_numbers`, retrieve the data with `batch_to_sign` and publish the `Signed<Batch>` to the gossip network.
- [x] Run the `AttesterRunner` in `Executor` if the node has an `Attester` key.
- [x] Add a minimum batch number to sign to prevent repeated signing

The code will re-submit the node's signature after a restart to a number of batches to be determined by the database query. We can say we'll re-publish the top 10 unsigned batches, or we can say we'll only look for the ones which follow the last finalised batch observed the Main node. It is enough to publish a vote once because of the way `BatchVotes` works (gossip peers will pull the data they need when they connect). Because we'll always have the batches in order, we can take note of the highest number we have published a vote for and never publish it again (until a restart), to save ourselves the trouble of repeatedly retrieving and signing the payload. That said it is assumed that `batch_to_sign` is cheap as it just needs to look up a hash in the DB.

### Follow up steps

The ticket will need a counterpart in `zksync-era` to implement the new `PersistentBatchStore` operations by querying Postgres. Need figuring out which column needs to go into the hash.

## Why ❔

To publish the attestations over L1 batches to the gossip network, so that a QC can be formed in https://github.com/matter-labs/era-consensus/pull/145

### DIFF
--- a/node/actors/executor/src/attestation.rs
+++ b/node/actors/executor/src/attestation.rs
@@ -22,7 +22,7 @@ pub(super) struct AttesterRunner {
 }
 
 impl AttesterRunner {
-    /// Crete a new instance of a runner.
+    /// Create a new instance of a runner.
     pub(super) fn new(
         block_store: Arc<BlockStore>,
         batch_store: Arc<BatchStore>,
--- a/node/actors/executor/src/attestation.rs
+++ b/node/actors/executor/src/attestation.rs
@@ -1,0 +1,73 @@
+//! Module to publish attestations over batches.
+
+use std::sync::Arc;
+
+use anyhow::Context;
+use zksync_concurrency::ctx;
+use zksync_concurrency::time;
+use zksync_consensus_network::gossip::BatchVotesPublisher;
+use zksync_consensus_storage::{BatchStore, BlockStore};
+
+use crate::Attester;
+
+const POLL_INTERVAL: time::Duration = time::Duration::seconds(1);
+
+/// Polls the database for new batches to be signed and publishes them to the gossip channel.
+pub(super) struct AttesterRunner {
+    block_store: Arc<BlockStore>,
+    batch_store: Arc<BatchStore>,
+    attester: Attester,
+    publisher: BatchVotesPublisher,
+}
+
+impl AttesterRunner {
+    /// Crete a new instance of a runner.
+    pub(super) fn new(
+        block_store: Arc<BlockStore>,
+        batch_store: Arc<BatchStore>,
+        attester: Attester,
+        publisher: BatchVotesPublisher,
+    ) -> Self {
+        Self {
+            block_store,
+            batch_store,
+            attester,
+            publisher,
+        }
+    }
+    /// Poll the database for new L1 batches and publish our signature over the batch.
+    pub(super) async fn run(self, ctx: &ctx::Ctx) -> ctx::Result<()> {
+        let public_key = self.attester.key.public();
+        loop {
+            // Pretend that the attesters can evolve.
+            let Some(attesters) = self.block_store.genesis().attesters.as_ref() else {
+                continue;
+            };
+            if !attesters.contains(&public_key) {
+                continue;
+            }
+
+            let unsigned_batch_numbers = self
+                .batch_store
+                .unsigned_batch_numbers(ctx)
+                .await
+                .context("unsigned_batch_numbers")?;
+
+            for bn in unsigned_batch_numbers {
+                if let Some(batch) = self
+                    .batch_store
+                    .batch_to_sign(ctx, bn)
+                    .await
+                    .context("batch_to_sign")?
+                {
+                    self.publisher
+                        .publish(attesters, &self.attester.key, batch)
+                        .await
+                        .context("publish")?;
+                }
+            }
+
+            ctx.sleep(POLL_INTERVAL).await?;
+        }
+    }
+}
--- a/node/actors/executor/src/attestation.rs
+++ b/node/actors/executor/src/attestation.rs
@@ -6,6 +6,7 @@ use anyhow::Context;
 use zksync_concurrency::ctx;
 use zksync_concurrency::time;
 use zksync_consensus_network::gossip::BatchVotesPublisher;
+use zksync_consensus_roles::attester::BatchNumber;
 use zksync_consensus_storage::{BatchStore, BlockStore};
 
 use crate::Attester;
@@ -38,6 +39,9 @@ impl AttesterRunner {
     /// Poll the database for new L1 batches and publish our signature over the batch.
     pub(super) async fn run(self, ctx: &ctx::Ctx) -> ctx::Result<()> {
         let public_key = self.attester.key.public();
+        // The first batch number we want to publish our vote for. We don't have to re-publish a vote
+        // because once it enters the vote register even future peers can pull it from there.
+        let mut min_batch_number = BatchNumber(0);
         loop {
             // Pretend that the attesters can evolve.
             let Some(attesters) = self.block_store.genesis().attesters.as_ref() else {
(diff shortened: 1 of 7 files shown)


node/actors/executor/src/attestation.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+    /// Poll the database for new L1 batches and publish our signature over the batch.
+    pub(super) async fn run(self, ctx: &ctx::Ctx) -> ctx::Result<()> {
+        let public_key = self.attester.key.public();
+        // The first batch number we want to publish our vote for. We don't have to re-publish a vote
+        // because once it enters the vote register even future peers can pull it from there.
+        let mut min_batch_number = BatchNumber(0);
+        loop {
+            // Pretend that the attesters can evolve.
+            let Some(attesters) = self.block_store.genesis().attesters.as_ref() else {
```
RomanBrodetski, line 47: not very relevant now, but doesn't this one need to pass the l1 batch number for which we need the list of attesters?

aakoshh: Today was the first time I heard about this plan to have attesters be queries that way. I knew looking for attesters in genesis probably wasn't going to be the final form, but it seems pervasive in the codebase, so I didn't see a reason to start deviating from it here. The same goes for the list of validators. 

What I thought was that there might be some kind of configuration that lives on until it's replaced with a new one, those _epochs_ that we saw in the Solidity code, so that one might have to know the current and previous and potentially the next epochs, but not arbitrary batch/block numbers.

This is potentially one of those places where an STM approach would help as well, to have an in-memory representation of what we believe is _the_ current state, without having to deal with eventual consistency between every component of the system.

So you are right that this is a design smell, but it's such a gaping hole that I don't know what to fill it with at the moment that looks more plausible.

node/actors/executor/src/attestation.rs
```diff
@@ -0,0 +1,89 @@
...
+                    continue;
+                }
+
+                if let Some(batch) = self
+                    .batch_store
+                    .batch_to_sign(ctx, bn)
+                    .await
+                    .context("batch_to_sign")?
+                {
+                    min_batch_number = batch.number.next();
```
RomanBrodetski, line 77: So I understand we always expect the `unsigned_batch_numbers` to be consecutive - you also left a comment that 
> // Batches appear in the store in order, even if we might have QC for a newer and not for an older batch

This makes sense, but then the API and the overall loop flow seems hard to follow: if `unsigned_batch_numbers` is supposed to be consecutive, why do we represent it as an array, not range? Now on the type level it's possible to represent errounous sequence like `4, 5, 7`. 

Also, feels confusing to have both an array of values to process and a mutable variable `min_batch_number` that allows for filtering such sequence.

Could we maybe keep a mutable range instead and operate over it? And thus only request the `last_persisted_batch` from `batch_store` to update the upper limit. 

There is probably a good reason why it is made like it is but then we should question the unerlying APIs/design decisions

node/actors/executor/src/attestation.rs
RomanBrodetski: Makes sense... So maybe we should request it one and not in a loop since it will not work correctly in the loop in its current form..

node/actors/executor/src/attestation.rs
aakoshh: Thanks for taking a close look at this; if it's confusing, I might have misunderstood something. 

> we always expect the `unsigned_batch_numbers` to be consecutive

Not necessarily. By saying "batches appear in order" I simply meant that the actual records in `l1_batches` will be be consecutive, we won't see batch 6 appear in that table before batch 5. However the result of `unsigned_batch_numbers` can have gaps in it, because the records in `l1_batches_consensus` can appear in arbitrary order, once we have the `l1_batch`. That's exactly why it's an array and not a range. 

Say the last finalized batch is batch 3, and we have batches [4,5,6,7,8] in the database, and just so it happens that our node has already collected and inserted batch QC for batch 6 and 8. When we query `unsigned_batch_numbers` it will return [4,5,7], as per your example. This just means that we are still interested in collecting signatures over 4, 5 and 7, but we don't need to sign 6 and 8 as we already have the QC. 

So, we start by retrieving the `Batch` with the `hash` for 4, sign it, publish it, and raise the `min_batch_number` to 4+1. We never have to sign 4 again until we restart. Then, we sign and publish 5, and finally 7. Next time we look for unsigned numbers, we can look from `min_block_number=8`, since everything that was before that we already signed. The query won't return 8 because it has a QC, but when 9 appears, we can sign it. 

The only thing that can throw this off at the moment is if `batch_to_sign` happened to return `None` for 4 and `Some` for 5, but I believe/hope that would make no sense, as it would mean we have the commitment for 5 earlier than 4, and I'm pretty sure that would break some recursive property of proofs.

You may be right that it's a code smell to query the `unsigned_batch_numbers` array and _then_ filter it. In fact in https://github.com/matter-labs/zksync-era/pull/2399 a very similar `min_batch_number` was passed to the query. It means something slightly different though: in the query its purpose is to represent the "last finalized batch", whatever that might be - currently it's just the last batch minus 10, but eventually it should come from the main node or L1. But from the outside perspective I thought the `PersistentBatchStore` (or the `BatchStore`, depending on where we want to put an L1 API client) itself is a self-contained authority on what needs to be signed. This particular filter is just a way to prevent repeatedly retrieving the payload from the DB to hash and sign it, when we have already just done this. Only this loop knows that, though. So it's just a separation of concerns based on who is supposed to possess what information.

Say we passed this filter into the `unsigned_batch_numbers` right here. The inner query might know what the latest finalized batch is (either because it's flagged in the DB, or there is an L1 client, or the main node has been queried), but now it doesn't know if it should override the value passed in by this loop, or not. That's why it's actually the maximum of two numbers: `where l1_batch.number > last_signed_by_us and l1_batch.number > last_finalized`.

aakoshh: > Could we maybe keep a mutable range instead and operate over it? And thus only request the last_persisted_batch from batch_store to update the upper limit.

IIUC that would be something like using `PersistentBatchStore::last_batch` for the upper limit, and introducing a `PersistentBatchStore::first_unsigned_batch() -> BatchNumber` or `PersistentBatchStore::last_finalized_batch() -> Option<BatchNumber>`, and then signing everything in between, periodically asking whether the upper limit can move on to the next step. The difference would be that it wouldn't skip all the numbers in between for which we already have a QC, unless we query those explicitly as we go to skip them, or retrieve and sign the payload regardless. 

I don't have a strong opinion on which one is better. For sure the main node API will need a `last_finalized_batch` query, which we discussed is one that doesn't have a gap after it. Implementing it now will mean we can have 1 less query methods.

aakoshh: I think another reason I went with this because it deferred most of the logic to the store, and I couldn't really come up with a good definition for a `last_finalized_batch` method. Does it mean it's on L1 and L1 is final? Surely just because it's signed doesn't actually mean it's "final". So last "quorumed"? 

How about `get_earliest_unsigned_batch_number`, returning `Some` or `None`? Not as nice as something that would return a finalized height, but it might be the only thing we can do. 

And then it's just extended to return all of them, but I do see the appeal of not repeatedly returning 10 items just to discard all of them, but simply querying the `last_batch`, which saves a few roundtrips in all except the first loop.

If you agree with this "earliest unsigned" I'll rewrite the two PRs in terms of that.

RomanBrodetski: hey thanks for all this context! It makes sense and I understand the reasoning. I marginally prefer `get_earliest_unsigned_batch_number` over the current implementation, but also feel free to keep as is - we may revisit all of this after we have the first testing results.

aakoshh: Okay, let's rewrite. It will save an extra query because the current way of determining the minimum height to consider for signing requires querying the maximum first.

node/actors/executor/src/attestation.rs
brunoffranca: For now I think it's fine to do it like this. We are going to deploy first without attester rotation anyway. There's a lot of places where we use genesis() so it's better if we do it separately.

node/actors/executor/src/attestation.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Polls the database for new batches to be signed and publishes them to the gossip channel.
+pub(super) struct AttesterRunner {
+    block_store: Arc<BlockStore>,
+    batch_store: Arc<BatchStore>,
+    attester: Attester,
+    publisher: BatchVotesPublisher,
+}
+
+impl AttesterRunner {
+    /// Crete a new instance of a runner.
```
brunoffranca, line 25:
```suggestion
    /// Create a new instance of a runner.
```

brunoffranca (approved):
Approved!

node/actors/executor/src/attestation.rs
aakoshh: https://github.com/matter-labs/era-consensus/pull/150

node/actors/executor/src/attestation.rs
aakoshh: This was moved out of the loop in #150

PR #150, opened 2024-07-08
refactor: Use earliest batch number to sign (BFT-474)

aakoshh:
## What ❔

Followup for some design discussions in https://github.com/matter-labs/era-consensus/pull/148

* Removes `BatchStore::unsigned_batch_numbers`
* Add `BatchStore::earliest_batch_number_to_sign`
* Add `BatchStore::latest_batch_number`
* Changes `AttesterRunner` to have a mutable range of `earliest_batch_number` and `latest_batch_number` instead of querying `unsigned_batch_numbers` and filtering it with a `min_batch_number`

## Why ❔

This might be more intuitive than the original version. 

Another motivation was that the current implementation of `unsigned_batch_numbers` in https://github.com/matter-labs/zksync-era/pull/2399 has to do two queries: 1) to figure out the latest batch number and 2) to find all batches without certificates within a limited range of the latest batch. 

Here we only query the latest batch number in each loop, and we forego the check on whether a QC already exists, otherwise we'd be back at doing two queries, or some ugly book keeping to only check it on the first loop after a restart. It should be cheap enough to publish a duplicate vote.


node/libs/storage/src/batch_store.rs
```diff
@@ -258,23 +258,55 @@ impl BatchStore {
...
+        }
+
+        let batch = self
+            .persistent
+            .last_batch(ctx)
+            .await
+            .context("persistent.last_batch()")?;
+
+        Ok(batch)
+    }
```
brunoffranca, line 283: For my education, is there any situation where we reach line 276? It seems like inner.persisted should always have a last batch.

brunoffranca (approved):

node/libs/storage/src/batch_store.rs
aakoshh: Great question, it should be obviously true, but it I suspect it's tricky at the moment because of how https://github.com/matter-labs/zksync-era/pull/2340 implements it:

* indeed we [initialise](https://github.com/matter-labs/zksync-era/blob/d12fbd3a380cbd58414342884a1adb146a46c42d/core/node/consensus/src/storage/store.rs#L92) the state when we create the `PersistentBatchStore`, so you'd think it will never be empty if we have a batch
* then, the `BatchStoreRunner` [copies the state](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/libs/storage/src/batch_store.rs#L153-L165) from the persistent inner to the wrapper 
* however, the `PersistentBatchStore` implementation returns a [hardcoded](https://github.com/matter-labs/zksync-era/blob/d12fbd3a380cbd58414342884a1adb146a46c42d/core/node/consensus/src/storage/store.rs#L455) `None` instead of the thing it keeps updated, because this mechanism would trigger the `SyncBatch` gossip which isn't configured correctly and causes tests to fail

So actually it will never be updated and we always reach line 276 at the moment. 

Your question also made me think whether it's okay to look at `inner.queued` at all. From what I can tell the way that gets [updated](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/actors/network/src/gossip/runner.rs#L364) is in `Network::run_stream` _iff_ our node asked the remote peer for the batch, and the precondition for that is that the remote peer reports these as [available](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/actors/network/src/gossip/fetch.rs#L149) to us. What I don't understand is why it reports the [queued](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/actors/network/src/gossip/runner.rs#L260) state rather than the persisted one, it looks like a chicken and egg problem: if I only report what's queued, and the only thing that gets queued is what I pull from others, and they only report to me as available what they managed to pull, then who reports availability first to get it going? Yet we know that if persisted is enabled it will gossip 😵 

What I was afraid of is that it will allow others to push one batche to us that we didn't want, and then force this thing to forego the check to the DB at all, causing it to stall by not sending anything else. 

Unrelated but we should keep in mind that it looks like the batch any peer gives us will enter our cache and prevent anything else from entering [without validation](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/libs/storage/src/batch_store.rs#L249-L255). We talked about Merkle proofs, they should be validated as early as possible, not during the storing operation asynchronously, because this node will start spreading the lie over gossip as soon as the batch is in the cache.

I will remove the `queued` from this because I don't trust it. It's also more likely that we'll have a payload to sign if we rely on the persisted batches.

aakoshh: For the blocks we can see that `queue_block` is called when the replica [saves a block](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/actors/bft/src/replica/block.rs#L47) during consensus. Can't see the same for a batch.

aakoshh: Found it, `queued` is [initialised](https://github.com/matter-labs/era-consensus/blob/96e5e692c064ad254dc5f7ec7b6ecdefdeb30f58/node/libs/storage/src/batch_store.rs#L199) to be the same as `persisted`

aakoshh: And here's another chicken and egg problem or potential deadlock:

The updating of the `batches_persisted` uses the `BatchStoreState` which has a `last: Option<SyncBatch>` field. This gets [updated in a loop](https://github.com/matter-labs/zksync-era/blob/c0f2329d47321d0d95f6e176a01e0bca5eeb03c5/core/node/consensus/src/storage/store.rs#L213) whenever we have a new batch. Currently we are ignoring the `SyncBatch::proof` field, so once we see an increase in the latest batch number, we can retrieve the batch. The update loop should do something to check more often for `SyncBatch` otherwise we won't be signing the batch until the _next_ one arrives. 

But!

The bigger problem is that `SyncBatch::proof` requires inclusion on L1, which means the batch has to be signed first, which means if we use the appearance of a new persisted batch in the data structure as a signal for signing it, we will never sign anything as soon as we introduce proofs.

I reckon that means I have to poll the database for the payload, there's no way around it, and can't use any of the existing in-memory reflections which are based on `SyncBatch`.

node/actors/executor/src/attestation.rs
```diff
@@ -39,51 +38,68 @@ impl AttesterRunner {
...
+
+            // Refresh the upper end of the range.
+            if earliest_batch_number > last_batch_number {
+                last_batch_number = self
+                    .batch_store
+                    .last_batch_number(ctx)
+                    .await
+                    .context("last_batch_number")?
+                    .unwrap_or(last_batch_number);
+            }
```
RomanBrodetski, line 102: it's indeed easier to follow now, but I find the execution flow here peculiar... 

We maintain these two pointers in memory (`mut earliest_batch_number` and `mut last_batch_number`), we go through batches in between those two numbers in a `while` loop and sign them, until either:
1. needed batch commitment is missing in the DB (we `break`)
2. we reached `last_batch_number + 1` (`while` condition doesn't hold true)

then we sleep for `POLL_INTERVAL` and conditionally bump `last_batch_number` - only if condition `2.` from above triggered - otherwise we return to the loop, waiting for `batch_to_sign` to become `batch_to_sign(ctx, earliest_batch_number)`. Why do we need both? We essentially have two restrictions that prevent event loop from continuing:
* `batch_to_sign(ctx, earliest_batch_number).is_some()`
* `last_batch_number(ctx)`

we check them in a different way and in different places, and also treat them differently - when (1) is false, we poll, whereas when we do (2), we update `last_batch_number`, but they do effectively the same - makes this component wait.

Could we achieve the same result in a way that's easier to follow? Something like:
```
let last_persisted_batch = self
                    .batch_store
                    .last_batch_number(ctx)
                    .await
                    .context("last_batch_number")?
                    .unwrap_or(last_batch_number);
let mut current_batch_number = self
            .batch_store
            .earliest_batch_number_to_sign(ctx)
            .await
            .context("earliest_batch_number_to_sign")?
            .unwrap_or(last_persisted_batch);
          
// log the current and last_persisted_batch numbers

loop {
             let batch = get_batch_to_sign_blocking(current_batch_number);
             // log the batch number and time spent waiting (although will be apparent from log lines)
            // also consider sending this timing to prometheus
             
            self.publisher
                    .publish(attesters, &self.attester.key, batch)
                    .await
                    .context("publish")?;
                  
           current_batch_number = current_batch_number.next()
}

// blocks until the batch is available in the DB for signing
    fn get_batch_to_sign_blocking(batch_number) -> attester::Batch {
    loop {
        let res =self
                    .batch_store
                    .batch_to_sign(ctx, batch_number)
                    .await
                    .context("batch_to_sign")?
        match res {
            Some(batch) => {
                return batch;
            }
            None => {
                sleep(polling_interval);
            }
        }
    }
}
```
the blocking logic can be done in `PersistentBatchStore` directly - will need to look deeper to see what makes more sense but I think we have a lot of polling there already.

Am I missing something or is the code above equivalent? If so, I like it more since the control flow is very clear (albeit still imperative). State space is much smaller - the only mutable var is `current_batch_number`.

node/libs/storage/src/batch_store.rs
RomanBrodetski: So what is **currently** the exact meaning of `queued` and `persisted` fields in `BatchStoreL::Inner`? The comments are not helpful:
```
/// Inner state of the `BatchStore`.
#[derive(Debug)]
struct Inner {
    /// Batches that are queued to be persisted.
    queued: BatchStoreState,
    /// Batches that are already persisted.
    persisted: BatchStoreState,
    cache: VecDeque<attester::SyncBatch>,
}
```

node/actors/executor/src/attestation.rs
aakoshh: Yours looks equivalent in what it achieves, just doing different queries in the background. I'll rewrite it that way.

I'd rather avoid blocking inside the `PersitentBatchStore` by polling inside the call. Easier to compose if individual methods do just simple things.

aakoshh: Done in https://github.com/matter-labs/era-consensus/pull/150/commits/49a9b94d2d8b76e7713bc6be3edf7fc3ad97ac84

I didn't add timing info - it's going to be on average the same as any recurring event between batches - but I'll add metrics as part of BFT-486 if that's okay.

RomanBrodetski: > Yours looks equivalent in what it achieves, just doing different queries in the background

for me the important part here is that we never invoke `last_batch_number` (less things to keep in mind) and instead of two mutable variables just stores one. The state in the original version is represented by a `range`, whereas here it's just one number. For me it's substantially easier to reason about, do you agree?

aakoshh: Yes I agree it's easier to follow, and thank you dedicating you time to make it nicer! 

I did it in a range because that's what you asked [earlier](https://github.com/matter-labs/era-consensus/pull/148#discussion_r1668747408): 

> Could we maybe keep a mutable range instead and operate over it? And thus only request the last_persisted_batch from batch_store to update the upper limit.

I just focussed on making that happen without looking for further opportunities 😅

RomanBrodetski: oh yes agree could have proposed this the first time 👍

node/libs/storage/src/batch_store.rs
brunoffranca: > I reckon that means I have to poll the database for the payload, there's no way around it, and can't use any of the existing in-memory reflections which are based on SyncBatch.

I agree as well. SyncBatches really shouldn't be in the data flow for L1 batch signing since they serve a different purpose. But at this point I'm also confused with the data flows for all these different tasks. I need to create some flow diagrams for this.

> So what is currently the exact meaning of queued and persisted fields in BatchStoreL::Inner?

AFAIK, `persisted` means it should be in the DB, `queued` is either in the DB or waiting to get in the DB. The name is a bit confusing because you expect them to be disjoint, but `queued` is a superset of `persisted`.

aakoshh: Yeah they only become apparent when you look at the way they are maintained. The `BlockStore` is a more authoritative example, as the batch syncing story is incomplete. 

### `queued` 
`queued` basically means stuff that has been added to the `cache`, which happens when the node _pulls_ batches from its peers. The `BatchRunner` is subscribed to changes in `queued` and will asynchronously push them to the `PersistentBatchStore`. 

Things that appear in `queued` are also _pushed_ to gossip peers. The precondition of pulling from a peer is a push from that peer to show that data is available.

However, the gossiping of `SyncBatch` is currently disabled:
* The `PersistentBatchStore::persisted()` implementation returns `None` for `last`, which causes `queued` to be initialised to `None` as well, and therefore no node is trying to push anything.
* Nothing seems to call `BatchStore::queue_batch` from anywhere outside gossip, unlike `BlockStore::queue_block`, which is called when a block is finalised. I suppose we could change the `BatchStoreRunner` to not just call `truncate_cache` but also append to the cache when a new thing is available in the DB, which would cause things to be gossiped. It's a snake that bites its own tail because we might have persisted it because it appeared in queued. 


### `persisted`

This is maintained by `BatchStoreRunner::run` which copies changed state from `PersistentBatchStore::persisted()` to the `inner.persisted`, to reflect what is in the DB. 

The content here currently is maintained by polling the database and putting in all the payload from all the blocks in the batch _without_ the inclusion proof. That will have to change in the future to wait for the proof.

aakoshh: Added docstrings.

PR #156, opened 2024-07-22
fixed a bunch of nits

pompon0:
tweaked names a bit, replaced context() with wrap() to propagate cancellation, fixed gossipnet msg size limits

### DIFF
--- a/node/actors/network/src/gossip/batch_votes.rs
+++ b/node/actors/network/src/gossip/batch_votes.rs
@@ -1,8 +1,8 @@
 //! Global state distributed by active attesters, observed by all the nodes in the network.
 use crate::watch::Watch;
-use std::{collections::HashSet, sync::Arc};
+use std::{collections::HashSet, fmt, sync::Arc};
 use zksync_concurrency::sync;
-use zksync_consensus_roles::attester::{self, Batch};
+use zksync_consensus_roles::attester;
 
 /// Represents the currents state of node's knowledge about the attester votes.
 ///
@@ -24,7 +24,7 @@ pub(crate) struct BatchVotes {
     /// for now, hoping that with 1 minute batches there's plenty of time for
     /// the quorum to be reached, but eventually we'll have to allow multiple
     /// votes across different heights.
-    pub(crate) votes: im::HashMap<attester::PublicKey, Arc<attester::Signed<Batch>>>,
+    pub(crate) votes: im::HashMap<attester::PublicKey, Arc<attester::Signed<attester::Batch>>>,
 
     /// Total weight of votes at different heights and hashes.
     ///
@@ -40,7 +40,7 @@ pub(crate) struct BatchVotes {
 
 impl BatchVotes {
     /// Returns a set of votes of `self` which are newer than the entries in `b`.
-    pub(super) fn get_newer(&self, b: &Self) -> Vec<Arc<attester::Signed<Batch>>> {
+    pub(super) fn get_newer(&self, b: &Self) -> Vec<Arc<attester::Signed<attester::Batch>>> {
         let mut newer = vec![];
         for (k, v) in &self.votes {
             if let Some(bv) = b.votes.get(k) {
@@ -61,7 +61,7 @@ impl BatchVotes {
     pub(super) fn update(
         &mut self,
         attesters: &attester::Committee,
-        data: &[Arc<attester::Signed<Batch>>],
+        data: &[Arc<attester::Signed<attester::Batch>>],
     ) -> anyhow::Result<bool> {
         let mut changed = false;
 
@@ -133,7 +133,7 @@ impl BatchVotes {
                                 sigs
                             });
                         attester::BatchQC {
-                            message: Batch {
+                            message: attester::Batch {
                                 number: *number,
                                 hash: *hash,
                             },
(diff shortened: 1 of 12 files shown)


node/actors/network/src/gossip/runner.rs
```diff
@@ -107,23 +86,23 @@ impl rpc::Handler<rpc::push_block_store_state::Rpc> for &PushBlockStoreStateServ
...
     }
 }
 
 #[async_trait]
-impl rpc::Handler<rpc::push_batch_store_state::Rpc> for &PushBatchStoreStateServer {
+impl rpc::Handler<rpc::push_batch_store_state::Rpc> for &PushServer<'_> {
     fn max_req_size(&self) -> usize {
-        self.max_batch_size.saturating_add(kB)
+        10 * kB
     }
```
aakoshh, line 98: I notice you removed this from the server side but not from the [client side](https://github.com/matter-labs/era-consensus/blob/1afde79e3891c14b080d35b3f69593843fc54970/node/actors/network/src/gossip/runner.rs#L338). 

Without this change the tests in `zksync-era` failed as the payload size was larger than 10kB, and after this change the gossip of batches was actually enabled. Am I missing something that keeps requests small now?

node/actors/network/src/lib.rs
```diff
@@ -131,12 +131,7 @@ impl Runner {
             });
 
             // Update QC batches in the background.
-            s.spawn(async {
-                match self.net.gossip.run_batch_qc_finder(ctx).await {
-                    Err(ctx::Error::Canceled(_)) => Ok(()),
-                    other => other,
-                }
-            });
+            s.spawn(self.net.gossip.run_batch_qc_finder(ctx));
```
aakoshh, line 134: Okay, I understand now why it behaved differently: a `.context` wrapped a `Cancel` into an `anyhow::Error` which then became a `ctx::Error::Internal` and caused the scope to fail during shutdown. 

This is a bit obscure; it would be helpful if `Wrap` or at least the [implementation of it](https://github.com/matter-labs/era-consensus/blob/cf01b6add5cb8993411b29bdc1ba2627bb993e0f/node/libs/concurrency/src/ctx/mod.rs#L269) would mention the significance of consistently using `.wrap` with `ctx::Result`.

brunoffranca (approved):

node/actors/network/src/gossip/batch_votes.rs
```diff
@@ -233,13 +232,20 @@ impl BatchVotesWatch {
 /// Wrapper around [BatchVotesWatch] to publish votes over batches signed by an attester key.
 pub struct BatchVotesPublisher(pub(crate) Arc<BatchVotesWatch>);
 
+impl fmt::Debug for BatchVotesPublisher {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("BatchVotesPublisher")
+            .finish_non_exhaustive()
+    }
+}
```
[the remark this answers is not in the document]
pompon0: because the nested field doesn't implement Debug.

node/actors/network/src/gossip/runner.rs
pompon0: There are many different messages here:
* PushBatchStoreState request: just range of BatchNumbers (begin and end)
* PushBAtchStoreState response: empty
* GetBatch request: just BatchNumber
* GetBatch response: the actual large batch.

Now that you mention it, I see that The PushBatchStoreState request is incorrectly defined. Let me fix it.

pompon0: done

node/actors/network/src/lib.rs
pompon0: Good point, I've added documentation to both Wrap and ctx::Error to cover that. If you think it is still too obscure, we can for example consider removing implicit conversion from anyhow::Error to ctx::Error.

aakoshh:
Great, changing `BatchStoreState::last` from a `SyncBatch` to a `BatchNumber` makes a whole lot more sense then unconditionally pushing all the L2 blocks to gossip peers after each L1 batch whether they need it or not 👍 We discussed this with @brunoffranca and @RomanBrodetski on standups but none of us were sure whether it's intentional, although the fact that the `.proto` file said it was a "QC" made it suspicious that it was a copy paste error from the similar block related artifacts. 

Perhaps the title can be updated to reflect this change as it will also have to be fixed on the era side, ie. a bit more significant than a nit.

node/actors/network/src/proto/gossip.proto
```diff
@@ -37,10 +37,12 @@ message PushBlockStoreState {
...
 // and actively fetch newest batch.
 message PushBatchStoreState {
+  reserved 2;
+  reserved "last";
   // First batch that the node has locally.
   optional uint64 first = 1; // required; BatchNumber
-  // Last batch QC that the node has locally.
-  optional roles.attester.SyncBatch last = 2; // optional
+  // Last batch that the node has locally.
+  optional uint64 last_v2 = 3; // optional
```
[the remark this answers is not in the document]
pompon0: the point here is that it doesn't matter - it is supposed to be a range of batches that the peer can send us and we don't care if it has them persisted or not (we wouldn't trust them anyway).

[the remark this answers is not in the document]
aakoshh: Okay so persisted or queued doesn't matter, but for example the fact that a proof is generated means that the asynchronously produced `commitment` is available, and not just the L1 batch itself. Never mind, just thought there is something better than v2.

PR #157, opened 2024-07-23
feat: Batch vote metrics (BFT-486)

aakoshh:
## What ❔

Adds metrics around the gossiping of batch votes and the storage of batches.

Added `gossip::metrics::BATCH_VOTES_METRICS`:
- [x] `committee_size` to show how many attesters we have, which can be correlated with the number of votes we receive over time. Wanted to add a weight based one as well, but not sure if counting the weights monotonically would not overflow the counter.
- [x] `votes_added` counts the votes received since start; with its rate-of-change we can get an idea of how many attesters are producing votes
- [x] `weight_added` counts the total weight of votes added since start, each normalized to [0; 1] range with the total committee weight; this way it shouldn't overflow, and its rate of growth should be 0.8-1.0 per batch.
- [x] `min_batch_number` shows the minimum batch we expect votes for; going up will indicate that QCs are formed
- [x] `last_added_vote_batch_number` is the number from the last vote we registered; apart from Byzantine votes it should average to go up as L1 batches are created and votes are received, even if a QC cannot be formed
- [x] `last_signed_batch_number` indicates that the current node is publishing its votes and keeping up with the L1 batches

Also implement metrics for the `BatchStore` similar to how the `BlockStore` has them:
- [x] `next` for `queued` and `persisted`
- [x] latencies for the `PersistedBatchStore` operations
- [x] latencies for the `wait_for_...` methods
- [x] last batch QC saved

The latencies for some of the `BatchStore` methods could potentially be less cumbersome if we used instrumentation.

## Why ❔

To make the voting process observable.


node/actors/network/src/gossip/batch_votes.rs
```diff
@@ -57,13 +72,14 @@ impl BatchVotes {
...
+    ///
     /// Returns true iff some new entry was added.
     pub(super) fn update(
         &mut self,
         attesters: &attester::Committee,
         data: &[Arc<attester::Signed<attester::Batch>>],
-    ) -> anyhow::Result<bool> {
-        let mut changed = false;
+    ) -> anyhow::Result<BatchUpdateStats> {
+        let mut stats = BatchUpdateStats::default();
```
pompon0, line 82: why not report these stats directly?

aakoshh: I made these [tests](https://github.com/matter-labs/era-consensus/blob/99df3002508d65f49ebb829b9a104dee4e577ea6/node/actors/network/src/gossip/tests/mod.rs#L611) to look for quorum and `BatchVotes` is easy to test in isolation. Adding global metrics into it feels disturbing. It even gets cloned IIRC by the subscribers. 

By contrast there is only one instance of `BatchVotesWatch` and it's meant to coordinate changes to the state by managing locks and subscriptions, so it felt like a better place to maintain the metrics. 

The [vise](https://crates.io/crates/vise) page recommends two ways to make metrics testable:
> * You may pass around references to the metric type(s) in the logic under test so that these types can be injected and then checked by tests.
> * Alternatively, your logic may produce statistics that are then reported as metrics (this may be beneficial for performance as well). In this case, produced statistics can be checked by tests

Because I didn't want to actually test that metrics are emitted, the stats approach seemed cleaner.

pompon0: ok

pompon0 (approved):

node/actors/network/src/gossip/metrics.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub(crate) min_batch_number: vise::Gauge<u64>,
+
+    /// Batch number in the last vote added to the register.
+    ///
+    /// This should go up as L1 batches are created, save for any temporary
+    /// outlier from lagging attesters or ones sending votes far in the future.
+    pub(crate) last_added_vote_batch_number: vise::Gauge<u64>,
+
+    /// Batch number of the last batch signed by this attester.
+    pub(crate) last_signed_batch_number: vise::Gauge<u64>,
```
RomanBrodetski, line 33: can postpone this, but can we have a `histogram` that would show the distribution of votes delay after an l1 batch was produced?

RomanBrodetski (approved):

node/actors/network/src/gossip/metrics.rs
aakoshh: Sounds interesting, I will create a ticket for this. There are a few challenges:
* the timestamp of the batch isn't available here
* currently we can receive votes earlier than we have the batch itself, so we would have to store a timestamp with the vote, and the retrospectively emit negative latencies when the batch is produced on the local node

We can have some specific workarounds: 
* For example when the `last_signed_batch_number` above changes it can act as an approximation of when the L1 batch was produced, but it's only an approximation because it depends on the polling frequency, and also it's only available on nodes which are attesters. 
* We could observe the time elapsed between the first and last vote. For attester nodes this would indicate the speed of propagation of votes over gossip, compared to the vote produced locally, although again the local vote might come later than the first gossiped one.
* We could implement Grzegorz's idea about only gossiping votes to nodes which have indicated they have the batch already, which would deal with the negative values; this simplifies the metrics in that we don't have to buffer timestamps and observe them the first time we learn about the batch.
* We could observe the difference between `l1_batches.created_at` and `l1_batches_consensus.created_at` when we insert the QC, which would be the overall latency of gossiping votes.

aakoshh: https://linear.app/matterlabs/issue/BFT-497/add-metric-for-l1-batch-vote-latency

aakoshh:
Added the following metrics to the [Consensus](https://grafana.matterlabs.dev/d/eba66cd9-5335-4521-9777-ad34a379f35c/era-consensus?orgId=1&var-datasource=mIPIusa4k) dashboard on Grafana: 

image

The queries behind them are the following:
* `max (network_gossip_batch_votes_last_signed_batch_number) by (pod) != 0`: last vote published by attesters
* `max (zksync_consensus_storage_batch_store_last_persisted_batch_qc) by (pod)`: last batch QC saved by nodes
* `max (network_gossip_batch_votes_last_added_vote_batch_number) by (pod)`: last vote received and added by nodes
* `min (network_gossip_batch_votes_min_batch_number) by (pod)`: minimum accepted vote by node
* `max(rate(network_gossip_batch_votes_votes_added[10m]) * 600) by (pod)`: number of votes added per minute over the last 10 minutes
* `max(network_gossip_batch_votes_committee_size) by (pod)`: committee size  to compare against number of votes added

PR #161, opened 2024-07-26
feat: Poll the main node for batch to vote on (BFT-496)

aakoshh:
## What ❔

Poll the main node for which batch height to vote on.

- [x] Return an `Option` rather than a `Vec` of `BatchQC` from `BatchVotes::find_quorum`: we decided not to implement voting on multiple heights for now, which makes it just confusing that a vector is returned.
- [x] Create an `AttestationStatusClient` trait with a method to poll the next `BatchNumber` to vote on, which can be injected as a `dyn` dependency
- [x] Create an `AttestationStatusWatch` that contains the next `BatchNumber` to vote on; this is updated by a single background task and listened to by multiple subscribers
- [x] Create an `AttestationStatusRunner` which is supposed to be started along with the `BatchStore` in `zksync-era` and polls the client; the point of this is to push out the poll interval configuration to the edge and make the `AttestationStatusWatch` the integration point
- [x] Change `AttesterRunner` to wait for the `AttestationStatusWatch` to change, then the payload to appear, then cast the vote; it doesn't need to worry about reverts because the node will restart if that happens
- [x] Change `run_batch_qc_finder` to wait for the `AttestationStatusWatch` to change, then wait for the next available QC, and save it; this might produce gaps which is normal on the external nodes, but undersireable on the main node - on the main node it only produces gaps if the attesters vote on higher heights despite the main node telling them not to, which if they are majority honest should not happen, and if they are not then what can we do anyway
- [x] Rename `PersistentBatchStore::earliest_batch_number_to_sign` to `next_batch_to_attest` in accordance with the new method on `consensus_dal`. 
- [ ] ~~Initialise `BatchNumber::min_batch_number` to by asking the main node where to resume from~~ This initialisation isn't necessary because 1) we decided that we'll only allow 1 vote per attester, so there is no attack vector of casting votes from batch 0 and 2) `run_batch_qc_finder` first waits for the API status to appear and then prunes all preceding data, so an older QC will not be saved. It was also undesirable to have to initialise from an API that might return nothing (or errors) for an unknown amount of time.

### Poll interval

The `AttesterStatusRunner` polls the API at fixed intervals without any exponential backoff. I thought 5s would be a reasonable default value. With 60s batches this seems fine because the *next* batch to sign will be signalled as soon as the current batch QC is inserted into the database, which is well ahead of time of when  the next batch and its commitment will be available. If we think we'll need to catch up with historical batch signatures it might be a bit slow. We generally still have to account for gossiping around the votes though, which in a large enough network could be on the order of several seconds as well.

### Potential deadlock

There is a pathological sequence of events that could prevent the feature from starting on a fresh database with no prior batch QCs:
1. Say we start from batch 0, and a QC is gathered, but saving it to the database (which is an async process) takes a long time on the main node.
2. Say it takes so long that batch 1 and batch 2 are created, but none of them have a QC yet.
3. Say we have two attesters A and B, and A polled the API when it showed batch 2, and the main node set its minimum batch number in the vote registry to batch 2 as well. 
4. Now the QC for batch 0 is saved to the database and the API indicates the next one to vote on is batch 1. 
5. Attester B casts its vote on batch 1 but it goes ignored by the main node vote registry because its looking for a quorum with at least batch number 2. 
6. Having missed an attestation the main node will never save a QC and never move on until it's restarted. 

Note that even if the registry minimum batch number was adjusted _down_ that might happen _after_ the vote for batch 1 has already been discarded, and because new votes are only pushed once through gossip there is no guarantee that it will get it again. 

The solution is coming in the form of fixing the starting point of gossip to be where the genesis starts, even if that means filling a potentially long history of batches in the beginning. See https://github.com/matter-labs/zksync-era/pull/2539

## Why ❔

We want to collect batch QCs without gaps, and for now put the main node in charge of what to vote on. The API to serve this information is in https://github.com/matter-labs/zksync-era/pull/2480 and this PR is the follow up to make polling that information part of the signing process.

### DIFF
--- a/node/actors/network/src/gossip/attestation_status.rs
+++ b/node/actors/network/src/gossip/attestation_status.rs
@@ -1,0 +1,85 @@
+use std::sync::Arc;
+
+use zksync_concurrency::{ctx, sync};
+use zksync_consensus_roles::attester;
+use zksync_consensus_storage::BatchStore;
+
+use crate::watch::Watch;
+
+/// An interface which is used by attesters and nodes collecting votes over gossip to determine
+/// which is the next batch they are all supposed to be voting on, according to the main node.
+#[async_trait::async_trait]
+pub trait AttestationStatusClient: 'static + std::fmt::Debug + Send + Sync {
+    /// Get the next batch number for which the main node expects a batch QC to be formed.
+    ///
+    /// The API might return an error while genesis is being created, which we represent with `None`
+    /// here and mean that we'll have to try again later.
+    async fn next_batch_to_attest(
+        &self,
+        ctx: &ctx::Ctx,
+    ) -> ctx::Result<Option<attester::BatchNumber>>;
+}
+
+/// Implement the attestation status for the main node by returning the next to vote on from the [BatchStore].
+#[derive(Debug, Clone)]
+pub struct LocalAttestationStatus(Arc<BatchStore>);
+
+impl LocalAttestationStatus {
+    /// Create local attestation client form a [BatchStore].
+    pub fn new(store: Arc<BatchStore>) -> Self {
+        Self(store)
+    }
+}
+
+#[async_trait::async_trait]
+impl AttestationStatusClient for LocalAttestationStatus {
+    async fn next_batch_to_attest(
+        &self,
+        ctx: &ctx::Ctx,
+    ) -> ctx::Result<Option<attester::BatchNumber>> {
+        self.0.next_batch_to_attest(ctx).await
+    }
+}
+
+/// Coordinate the attestation by showing the status as seen by the main node.
+pub struct AttestationStatus {
+    /// Next batch number where voting is expected.
+    ///
+    /// Its value is `None` until the background process polling the main node
+    /// can establish a value to start from.
+    pub next_batch_to_attest: Option<attester::BatchNumber>,
+}
+
+/// The subscription over the attestation status which votes can monitor for change.
+pub type AttestationStatusReceiver = sync::watch::Receiver<AttestationStatus>;
+
+/// A [Watch] over an [AttestationStatus] which we can use to notify components about
+/// changes in the batch number the main node expects attesters to vote on.
+pub(crate) struct AttestationStatusWatch(Watch<AttestationStatus>);
+
+impl Default for AttestationStatusWatch {
+    fn default() -> Self {
+        Self(Watch::new(AttestationStatus {
+            next_batch_to_attest: None,
+        }))
+    }
+}
+
+impl AttestationStatusWatch {
+    /// Subscribes to AttestationStatus updates.
+    pub(crate) fn subscribe(&self) -> AttestationStatusReceiver {
+        self.0.subscribe()
+    }
+
+    /// Set the next batch number to attest on and notify subscribers it changed.
+    pub(crate) async fn update(&self, next_batch_to_attest: attester::BatchNumber) {
+        let this = self.0.lock().await;
+        this.send_if_modified(|status| {
+            if status.next_batch_to_attest == Some(next_batch_to_attest) {
+                return false;
+            }
+            status.next_batch_to_attest = Some(next_batch_to_attest);
+            true
+        });
+    }
+}
--- a/node/actors/network/src/gossip/attestation_status.rs
+++ b/node/actors/network/src/gossip/attestation_status.rs
@@ -1,6 +1,7 @@
+use std::fmt;
 use std::sync::Arc;
 
-use zksync_concurrency::{ctx, sync};
+use zksync_concurrency::{ctx, sync, time};
 use zksync_consensus_roles::attester;
 use zksync_consensus_storage::BatchStore;
 
@@ -9,7 +10,7 @@ use crate::watch::Watch;
 /// An interface which is used by attesters and nodes collecting votes over gossip to determine
 /// which is the next batch they are all supposed to be voting on, according to the main node.
 #[async_trait::async_trait]
-pub trait AttestationStatusClient: 'static + std::fmt::Debug + Send + Sync {
+pub trait AttestationStatusClient: 'static + fmt::Debug + Send + Sync {
     /// Get the next batch number for which the main node expects a batch QC to be formed.
     ///
     /// The API might return an error while genesis is being created, which we represent with `None`
@@ -20,28 +21,8 @@ pub trait AttestationStatusClient: 'static + std::fmt::Debug + Send + Sync {
     ) -> ctx::Result<Option<attester::BatchNumber>>;
 }
 
-/// Implement the attestation status for the main node by returning the next to vote on from the [BatchStore].
-#[derive(Debug, Clone)]
-pub struct LocalAttestationStatus(Arc<BatchStore>);
-
-impl LocalAttestationStatus {
-    /// Create local attestation client form a [BatchStore].
-    pub fn new(store: Arc<BatchStore>) -> Self {
-        Self(store)
-    }
-}
-
-#[async_trait::async_trait]
-impl AttestationStatusClient for LocalAttestationStatus {
-    async fn next_batch_to_attest(
-        &self,
-        ctx: &ctx::Ctx,
-    ) -> ctx::Result<Option<attester::BatchNumber>> {
-        self.0.next_batch_to_attest(ctx).await
-    }
-}
-
 /// Coordinate the attestation by showing the status as seen by the main node.
+#[derive(Debug, Clone)]
 pub struct AttestationStatus {
     /// Next batch number where voting is expected.
     ///
@@ -50,12 +31,19 @@ pub struct AttestationStatus {
     pub next_batch_to_attest: Option<attester::BatchNumber>,
 }
 
-/// The subscription over the attestation status which votes can monitor for change.
+/// The subscription over the attestation status which voters can monitor for change.
 pub type AttestationStatusReceiver = sync::watch::Receiver<AttestationStatus>;
 
 /// A [Watch] over an [AttestationStatus] which we can use to notify components about
 /// changes in the batch number the main node expects attesters to vote on.
-pub(crate) struct AttestationStatusWatch(Watch<AttestationStatus>);
+pub struct AttestationStatusWatch(Watch<AttestationStatus>);
+
+impl fmt::Debug for AttestationStatusWatch {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("AttestationStatusWatch")
+            .finish_non_exhaustive()
+    }
+}
 
 impl Default for AttestationStatusWatch {
     fn default() -> Self {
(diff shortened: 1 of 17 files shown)


node/actors/network/src/gossip/attestation_status.rs
```diff
@@ -0,0 +1,85 @@
...
+#[async_trait::async_trait]
+pub trait AttestationStatusClient: 'static + std::fmt::Debug + Send + Sync {
+    /// Get the next batch number for which the main node expects a batch QC to be formed.
+    ///
+    /// The API might return an error while genesis is being created, which we represent with `None`
+    /// here and mean that we'll have to try again later.
+    async fn next_batch_to_attest(
+        &self,
+        ctx: &ctx::Ctx,
+    ) -> ctx::Result<Option<attester::BatchNumber>>;
```
pompon0, line 20: How about we shift the abstraction a bit and make the AttestationReceiver abstract? Sth like/equivalent to
```
trait AttestationReceiver {
  fn subscribe() -> watch::Receiver<Option<attester::BatchNumber>>
}
```
I would like to push the active polling as far out as possible, to avoid hardcoded delays which the outer logic has to be aware of. In this case the polling still will be performed, but in zksync-era code.

node/actors/network/src/gossip/mod.rs
```diff
@@ -12,16 +12,22 @@
...
+    AttestationStatusClient, AttestationStatusReceiver, LocalAttestationStatus,
+};
 pub use self::batch_votes::BatchVotesPublisher;
 use self::batch_votes::BatchVotesWatch;
 use crate::{gossip::ValidatorAddrsWatch, io, pool::PoolWatch, Config, MeteredStreamStats};
+use attestation_status::AttestationStatusWatch;
 use fetch::RequestItem;
 use std::sync::{atomic::AtomicUsize, Arc};
 pub(crate) use validator_addrs::*;
+use zksync_concurrency::time::Duration;
```
pompon0, line 25: nit: please use qualified names for concurrency imports - in this case "use zksync_concurrency::time;"

node/actors/network/src/gossip/mod.rs
```diff
@@ -148,37 +161,91 @@ impl Network {
...
+        let mut recv_votes = self.batch_votes.subscribe();
+        let mut recv_status = self.attestation_status.subscribe();
+
+        let mut prev_batch_number = None;
+
+        'status: loop {
+            // Wait until the status indicates that we're ready to sign the next batch.
+            // This is not strictly necessary but avoids repeatedly finding the same quorum, or having to skip it until it changes.
+            let next_batch_number =
+                sync::wait_for_some(ctx, &mut recv_status, |s| match s.next_batch_to_attest {
```
pompon0, line 182: nit `sync::changed` should do

node/actors/network/src/gossip/mod.rs
```diff
@@ -148,37 +161,91 @@ impl Network {
...
+
+            // Now wait until we find the next quorum, whatever it is:
+            // * on the main node, if attesters are honest, they will vote on the next batch number and the main node will not see gaps
+            // * on external nodes the votes might be affected by changes in the value returned by the API, and there might be gaps
+            // What is important, though, is that the batch number does not move backwards while we look for a quorum, because attesters
+            // (re)casting earlier votes will go ignored by those fixed on a higher min_batch_number, and gossip will only be attempted once.
+            // The possibility of this will be fixed by deterministally picking a start batch number based on fork indicated by genesis.
+            loop {
+                let quorum_opt = {
+                    let votes = sync::changed(ctx, &mut recv_votes).await?;
```
pompon0, line 204: Are you sure you don't need "recv_votes.mark_changed()" before this await? In particular `subscribe()` returns a receiver which is *seen*.

node/actors/network/src/gossip/mod.rs
```diff
@@ -148,37 +161,91 @@ impl Network {
...
+                };
+
+                if let Some(qc) = quorum_opt {
+                    self.batch_store
+                        .persist_batch_qc(ctx, qc)
+                        .await
+                        .wrap("persist_batch_qc")?;
+
+                    continue 'status;
+                }
```
pompon0, line 215: nit: I think this inner loop is equivalent to:
```
let qc = sync::wait_for_some(ctx, &mut recv, |v| v.find_quorum(...)).await?;
self.batch_Store.persist_batch_qc(ctx,qc).await?;
```

node/actors/network/src/gossip/mod.rs
```diff
@@ -148,37 +161,91 @@ impl Network {
...
+                    // happened to decrease. But we decided to fix it at the source,
+                    // so the only place that is adjusted is before looking for a QC.
+                }
+                Ok(None) => tracing::debug!("waiting for attestation status..."),
+                Err(error) => tracing::error!(
+                    ?error,
+                    "failed to poll attestation status, retrying later..."
+                ),
             }
+            ctx.sleep(POLL_INTERVAL).await?;
```
pompon0, line 248: that's the sleep that I would like to avoid in era-consensus

node/actors/network/src/gossip/mod.rs
aakoshh: Hm, I wouldn't want to call it here because then it would become a tight loop. I didn't know `.subscribe()` returns it seen, that's good to keep in mind, however in this case it doesn't matter because of the call to `set_min_batch_number` which happens before this loop, which will cause the whole `BatchVotes` to change, so this call will get something immediately, and then every time a vote is added.

node/actors/network/src/gossip/mod.rs
aakoshh: Yeah, that looks neat 👍

node/actors/network/src/gossip/mod.rs
aakoshh: I don't fancy hardcoded polls either, just following the pattern. They might have been originally more prevalent in `zksync-era`, and I started adding them to places like the `AttestationRunner`, introducing them to `era-consensus. Maybe we could just make them a parameter, so at least they are easy to change from `zksync-era`? Although getting them to where they need to be might be awkward, with miscellaneous poll interval defaults being added to the `Config` of the `Executor.

node/actors/network/src/gossip/attestation_status.rs
aakoshh: You mentioned before that you might want to extend the `AttestationStatus` with further things like the number of concurrent votes, which I expected would not change during the lifetime of the program, but may be announced by the main node. I suppose we can add further async methods to `AttestationReceiver` above that one can call.

pompon0: or we can have one large watch, it should work either way

node/actors/network/src/gossip/mod.rs
pompon0: so the patter that I was trying to maintain so far was to have the polls in zksync-era, where they are just unavoidable because you cannot subscribe to postgres db afaik.

node/actors/network/src/gossip/attestation_status.rs
pompon0: technically we could even use the watch `instead` of the `AttestationReceiver` trait, except that it would limit extensionality of the API. But I guess we can revert to having a trait at any point.

aakoshh: I do wonder though if we should just expose the `AttestationStatusWatch` itself as a parameter to the `Explorer`. There isn't much magic there, it has `update` and `subscribe`, which can be driven from `zksync-era`.

pompon0: yes, and imo a piece of shared state with a well designed API is more maintainable API surface than a trait interface.

aakoshh: I agree with encapsulating things and having shared memory be a good integration pattern. I have a Software Transactional Memory library that could be adopted to work with ctx to make this composable as well.

aakoshh: I am a bit worried though that moving the poll onto the era side will make it more cumbersome to compose the application now. The Network is now self contained so the vote propagation tests did not have to change. If it just sits there doing nothing, and not even the Executor has the stuff we just make it more fragmented. I’ll do it and we’ll see. After all the PermanentStore variants do it too, maybe this is a general component to watch the main node.

aakoshh: Created an `AttestationStatusRunner` to encapsulate the polling, created by `let (status, runner) = AttestationStatusWatch::new(client, poll_interval);`, which should be done where we instantiate the `BlockStore`, the `BatchStore` and their respective runners. This will need to be duplicated for the main and external node setups, and then only the `AttestationStatusWatch` is passed to `Executor` and then onto the `Network`.

pompon0: We shouldn't need to define the AttestationStatusClient now, no?

pompon0: like, AttestationStatusWatch is our API, and how it will be fed information is up to the caller.

aakoshh: It's for convenience really, to be used by the `AttestationStatusRunner` so you don't have to write that loop, just hook in either the database or the API based query method. If one doesn't want to use the runner, they don't have to.

aakoshh: Moved.

pompon0: thanks

pompon0 (approved):
LGTM! thx

node/actors/network/src/gossip/attestation_status.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::watch::Watch;
+
+/// Coordinate the attestation by showing the status as seen by the main node.
+#[derive(Debug, Clone)]
+pub struct AttestationStatus {
+    /// Next batch number where voting is expected.
+    ///
+    /// Its value is `None` until the background process polling the main node
+    /// can establish a value to start from.
+    pub next_batch_to_attest: Option<attester::BatchNumber>,
```
RomanBrodetski, line 15: are we 100% sure `None` should be part of the state space? Is it possible to delay initialization before it's accessible?

aakoshh: In https://github.com/matter-labs/zksync-era/pull/2544 even the API can return `None` until `Genesis` is available. I'm not sure how long that is, probably not that long in practice. I'm also finding it difficult to understand what a delayed initialisation would mean for an external node: it doesn't run until the main node API is available _and_ returns `Some`, okay, but how about the main node itself? If I can't start the `Executor` until we have `Some` value for this, can it even create the `Genesis` in the first place?

pompon0: If the genesis.first_block points to a block in the future, then it may happen that the initial batch number is not known. Since now the made the AttestationStatusWatch the API, we can require that it is initialized before the executor is started (i.e.wait for the first full response from the attestationStatus rpc).

pompon0: In https://github.com/matter-labs/zksync-era/pull/2544 API may return None, because I don't understand the design principles behind the Web3Error, so I wasn't sure how to embed the NotFound response in the error space.

brunoffranca (approved):

node/actors/network/src/gossip/attestation_status.rs
aakoshh: Okay, if it is indeed safe to delay starting the Executor on the main node until we have this information by polling the store, we can get rid of the None from the Watch. 

On the external node it’s safe because it will have to pull genesis anyway

I will do it as a separate PR.

PR #162, opened 2024-07-26
Added support for splitting proto packages into multiple files

pompon0:
Without it it was impossible to break dependency cycles in a backward compatible way.
Also I've split the roles proto files to allow attester messages to include GenesisHash message.


node/libs/roles/src/proto/attester/messages.proto
```diff
@@ -0,0 +1,50 @@
...
+  optional uint64 number = 1; // required
+  optional BatchHash hash = 2; // required
+}
+
+message BatchQC {
+  reserved 2, 3; 
+  reserved "signers", "sig";
+
+  // TODO: de-deprecate these when we move back to using BLS aggregates, 
+  // or consider the ones at the bottom.
```
aakoshh, line 28:
```suggestion
  // or consider using `AggregateMultiSig`
```

aakoshh (approved):
Nice!

PR #163, opened 2024-07-28
refactor: update/add committee data types (BFT-443)

moshababo:
## What ❔

Adding minimally needed changes to support https://github.com/matter-labs/zksync-era/pull/2518, as part of committees rotation (BFT-484).

* Adding `pop: validator::Signature` (proof-of-possession) field to `WeightedValidator`.
* Adding `ValidatorCommittee` & `AttesterCommittee` types to provide protobuf formatting for DB store.

## Why ❔

These types cannot be placed in `zksync_node_consensus` because they are required by `zksync_dal::consensus_dal`, which is a dependency of `zksync_node_consensus`.

### DIFF
--- a/node/libs/roles/src/proto/validator.proto
+++ b/node/libs/roles/src/proto/validator.proto
@@ -198,4 +198,9 @@ message AggregateSignature {
 message WeightedValidator {
   optional PublicKey key = 1; // required
   optional uint64 weight = 2; // required
+  optional Signature pop = 3; // required
+}
+
+message ValidatorCommittee {
+   repeated WeightedValidator members = 1; // required
 }
--- a/node/libs/roles/src/proto/validator.proto
+++ b/node/libs/roles/src/proto/validator.proto
@@ -198,7 +198,7 @@ message AggregateSignature {
 message WeightedValidator {
   optional PublicKey key = 1; // required
   optional uint64 weight = 2; // required
-  optional Signature pop = 3; // required
+  optional Signature proof_of_possession = 3; // optional
 }
 
 message ValidatorCommittee {
--- a/node/libs/roles/src/attester/conv.rs
+++ b/node/libs/roles/src/attester/conv.rs
@@ -1,7 +1,4 @@
-use super::{
-    AggregateSignature, Batch, BatchHash, BatchNumber, BatchQC, Msg, MsgHash, MultiSig, PublicKey,
-    Signature, Signed, Signers, SyncBatch, WeightedAttester,
-};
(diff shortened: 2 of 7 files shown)


node/libs/roles/src/proto/validator.proto
```diff
@@ -198,4 +198,9 @@ message AggregateSignature {
 message WeightedValidator {
   optional PublicKey key = 1; // required
   optional uint64 weight = 2; // required
+  optional Signature pop = 3; // required
```
pompon0, line 201: use the full name: proof_of_possession.

pompon0: It has to stay optional until we bump the consensus protocol version. Also we need to think through how do we do the rollout, given that we already have external users of the consensus.

moshababo: Changes here: https://github.com/matter-labs/era-consensus/pull/163/commits/1305572a8bc9e3535671be5a6d246133efc1f9f1

node/libs/roles/src/proto/validator.proto
```diff
@@ -198,4 +198,9 @@ message AggregateSignature {
 message WeightedValidator {
   optional PublicKey key = 1; // required
   optional uint64 weight = 2; // required
+  optional Signature proof_of_possession = 3; // optional
```
pompon0, line 201: Sorry, I got confused for a moment: so here you are introducing the *new* types for committees which are *not* placed in the Genesis. Also, afaict these types are supposed to be used solely on the zksync-era side, in which case you should define them in zksync-era, not here.

pompon0: let's discuss it on the standup

moshababo: `WeightedValidator` _is_ placed in Genesis (part of `validator::Committee`), that's why I thought it's fine to make it optional, as part of the rollout (although it's not ideal).

The list types (`ValidatorCommittee`, etc.) are indeed about introducing a new type which isn't part of Genesis. In addition, They're simpler types, minimal for (de)serializing for DB store (AFAIK a simpler, `Vec<WeightedValidator>` isn't viable in protobuf).

As for the location in the codebase, since it will eventually need to be also passed in-mem from `zksync-era` to `era-consensus`, placing it here makes sense. Otherwise, another (de)serialization will need to happen.

moshababo:
Closing in favor of not introducing cross-repo integration, and instead having https://github.com/matter-labs/zksync-era/pull/2518 to add temporary data type clones.

PR #164, opened 2024-07-29
removed panicking when peer_addr fails

pompon0:
peer_addr() may fail if the network socket is closed. It needs to be handled gracefully.
I've also refactored some of the error handling to propagate cancellation better.

### DIFF
--- a/node/actors/network/src/frame.rs
+++ b/node/actors/network/src/frame.rs
@@ -1,6 +1,7 @@
 //! Simple frame encoding format (length ++ value) for protobuf messages,
 //! since protobuf messages do not have delimiters.
 use crate::{mux, noise::bytes};
+use anyhow::Context as _;
 use zksync_concurrency::{ctx, io};
 
 /// Reads a raw frame of bytes from the stream and interprets it as proto.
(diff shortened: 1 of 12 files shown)


node/actors/network/src/frame.rs
```diff
@@ -53,28 +54,41 @@ pub(crate) async fn recv_proto<T: zksync_protobuf::ProtoFmt, S: io::AsyncRead +
...
     let msg_size = u32::from_le_bytes(msg_size);
-    anyhow::ensure!(
-        msg_size as usize <= max_size,
-        "message too large: max = {max_size}, got {msg_size}",
-    );
+    if msg_size as usize > max_size {
+        return Err(
+            anyhow::format_err!("message too large: max = {max_size}, got {msg_size}",).into(),
+        );
+    }
```
aakoshh, line 67: What was wrong with `anyhow::ensure!`?

pompon0: anyhow::ensure needs the function to return exactly anyhow::Result (with ctx::Result it won't compile). It is possible to write a more general macro which will work with any error type implementing `From<anyhow::Error>`.

PR #167, opened 2024-08-01
feat: Add GenesisHash to AttestationStatusClient and reject unexpected updates (BFT-496)

aakoshh:
## What ❔
- [x] Add the initial `GenesisHash` to `AttestationStatus` and change `AttestationStatusWatch::update` to take `GenesisHash` and return an error if the latest from the client indicates a change in genesis. This stops the runner and thus should stop the node itself when the error bubbles up. On external nodes this is redundant with the routine that monitors the main node API for changes. On the main node I shouldn't happen.
- [x] Change `next_batch_to_attest() -> Option<BatchNumber>` into `attestation_status() -> Option<AttestationStatus>` on `AttestationStatusClient`, so it now includes the `GenesisHash` as well.
- [x] `LocalAttestationStatusClient` returns the `GenesisHash` it was started with 

## Why ❔

This came out of a discussion here: https://github.com/matter-labs/zksync-era/pull/2544#discussion_r1699598459 

Notably the `ConsensusDal::attestation_status()` now returns the `GenesisHash`, which I thought was only to signal through the API from the main node to the external nodes that a reorg has happened (which they would eventually learn anyway through polling the genesis endpoint, causing them to restart), and they should not attest for batches if they are on a different fork. The comment raised the issue that on the main node I discarded the `genesis` field from the response because I assumed it can only be the same as the one we started the `Executor` with, and second guessing it in the `BatchStore` would be awkward: the original genesis wasn't available to check against (it's cached in the `BlockStore`) and running a query to fetch it (even if the `PersistentBatchStore` supported it) would just compare it to what it already is. 

The way I handled this mismatch for external nodes was to just stop updating the status by returning `None` and wait for the restart, treating it like any other transient RPC error, it was just logged. It does make sense though to elevate it higher and be able to stop the node if it's in an inconsistent state. 

Now it's part of the `AttestationStatusWatch` because that is what we consider to be our interface with the node, (and not the `AttestationStatusRunner` which is a convenience construct). 

### Reorgs without restart?

If an inconsistency happened on the main node it wouldn't necessarily have to be fatal: if the genesis was allowed to change, and the components such as the `BlockStore` were able to handle it at all, then ostensibly the `AttestationStatus*` stuff could recover as well, for example by resetting the `BatchVotes` if they see a change in the `genesis`. The problem currently is that they might have already received and discarded votes for the new fork, which would not be gossiped again until clients are reconnected. 

Apparently we don't plan to handle regenesis on the fly, so `AttestationStatus::genesis` is only present to prevent any attempt at changing it by _causing_ a restart instead.

### Atomicity

In the first version of this PR the `BatchStore` and `PersistentBatchStore` were also changed to have an `attestation_status()` method that returned a `GenesisHash` along with the `BatchNumber`. The only way I could make sense of this was if 1) changes in genesis while running the main node were allowed and 2) they could happen between calls to `BlockStore::genesis()` and `BatchStore::next_batch_to_attest()`, since those are not atomic methods. We discussed that this will not be supported, the `Genesis` cached in `BlockStore` will not change. Since we only start the `Executor` and the `AttestationStatusRunner` [after regenesis](https://github.com/matter-labs/zksync-era/blob/1d206c0af8f28eb00eb1498d6f2cdbb45ffef72a/core/node/consensus/src/mn.rs#L39) in `zksync-era`, we don't have to worry about this changing. In that light it would be counter intuitive to return the genesis hash from `BatchStore`.

(In the future we could consider using Software Transactional Memory to have an in-memory representation of the state where we can do consistent read and writes between multiple Watch-like references. The DAL is capable of transactional reads, and it would be nice if we could combine reading for example here the genesis and the last/next batch and not have to worry about having to compare hashes, when all of these are local reads.)

### DIFF
--- a/node/actors/network/src/gossip/attestation_status.rs
+++ b/node/actors/network/src/gossip/attestation_status.rs
@@ -6,13 +6,27 @@ use zksync_consensus_roles::attester;
 use crate::watch::Watch;
 
 /// Coordinate the attestation by showing the status as seen by the main node.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AttestationStatus {
     /// Next batch number where voting is expected.
     ///
     /// The node is expected to poll the main node during initialization until
     /// the batch to start from is established.
     pub next_batch_to_attest: attester::BatchNumber,
+    // The hash of the genesis of the chain to which the L1 batches belong.
+    //
+    // A change in this value would indicate a reorg on the main node.
+    // On the main node itself this is not expected to change because
+    // a reorg involves a restart, and a regenesis happens before the
+    // executor is started. If it could happen, it could be used to
+    // signal to the `BatchVotes` that votes need to be cleared and
+    // potentially discarded votes received over gossip would need
+    // to be re-acquired (which doesn't happen at the moment unless
+    // the connection is re-established).
+    //
+    // It is not added yet as the system is not expected to be able
+    // to handle changes in the value.
+    //pub genesis: attester::GenesisHash,
 }
 
 /// The subscription over the attestation status which voters can monitor for change.
--- a/node/actors/network/src/gossip/attestation_status.rs
+++ b/node/actors/network/src/gossip/attestation_status.rs
@@ -72,6 +72,17 @@ impl AttestationStatusWatch {
                 status.genesis,
                 genesis
             );
+            // The next batch to attest moving backwards could cause the voting process
+            // to get stuck due to the way gossiping works and the BatchVotes discards
+            // votes below the expected minimum: even if we clear the votes, we might
+            // not get them again from any peer. By returning an error we can cause
+            // the node to be restarted and connections re-established for fresh gossip.
+            anyhow::ensure!(
+                status.next_batch_to_attest <= next_batch_to_attest,
+                "next batch to attest moved backwards: {} -> {}",
+                status.next_batch_to_attest,
+                next_batch_to_attest
+            );
         }
         this.send_if_modified(|status| {
             if status.next_batch_to_attest == next_batch_to_attest {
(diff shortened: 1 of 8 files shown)


node/actors/network/src/gossip/attestation_status.rs
```diff
@@ -6,13 +6,27 @@ use zksync_consensus_roles::attester;
...
+    // a reorg involves a restart, and a regenesis happens before the
+    // executor is started. If it could happen, it could be used to
+    // signal to the `BatchVotes` that votes need to be cleared and
+    // potentially discarded votes received over gossip would need
+    // to be re-acquired (which doesn't happen at the moment unless
+    // the connection is re-established).
+    //
+    // It is not added yet as the system is not expected to be able
+    // to handle changes in the value.
+    //pub genesis: attester::GenesisHash,
```
pompon0, line 29: We are not (ever) going to add support for changing the genesis in runtime. All I meant is that change of AttestationStatus with mismatching genesis hash should not be accepted by AttestationStatusWatch.

aakoshh: Okay, I can move the check into the `AttestationStatusWatch::update` and return an error from there instead. 

But then I'm again not sure that `BatchStore` and `PersistentBatchStore` should be returning `GenesisHash` if we know it can only ever be the same as `BlockStore::genesis()`.

pompon0: true, there is no need for fetching genesis through PersistentBatchStore/BatchStore

aakoshh: Okay great, then I'll revert those two to return `next_batch_to_attest`.

pompon0 (approved):

node/actors/network/src/gossip/attestation_status.rs
```diff
@@ -43,14 +54,32 @@ impl AttestationStatusWatch {
...
+                status.genesis == genesis,
+                "the attestation status genesis changed: {:?} -> {:?}",
+                status.genesis,
+                genesis
+            );
+        }
         this.send_if_modified(|status| {
             if status.next_batch_to_attest == next_batch_to_attest {
                 return false;
             }
```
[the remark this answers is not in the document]
aakoshh: Okay. I didn't think that was a showstopper, in that it could, but not necessarily will lead to getting stuck. I suppose returning an error will force a restart and get us out of a jam.

PR #170, opened 2024-08-02
improve: Relax the signature check in BatchVote

aakoshh:
## What ❔

`BatchVote::update` ignores invalid signatures instead of returning an error.

## Why ❔

Because on `stage-v2` the `moby` variant of the external node runs an earlier version of the code without the `Batch::genesis` field, which causes it to reject all votes coming from the main and the other external nodes. The error causes it to disconnect its gossip peers, which is a bit heavy handed given that this is an unstable feature.

### DIFF
--- a/node/actors/executor/src/attestation.rs
+++ b/node/actors/executor/src/attestation.rs
@@ -5,7 +5,7 @@ use anyhow::Context;
 use std::sync::Arc;
 use zksync_concurrency::{ctx, sync, time};
 use zksync_consensus_network::gossip::{
-    AttestationStatus, AttestationStatusReceiver, AttestationStatusWatch, BatchVotesPublisher,
+    AttestationStatusReceiver, AttestationStatusWatch, BatchVotesPublisher,
 };
 use zksync_consensus_roles::attester;
 use zksync_consensus_storage::{BatchStore, BlockStore};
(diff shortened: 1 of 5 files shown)


node/actors/network/src/gossip/batch_votes.rs
```diff
@@ -125,10 +130,10 @@ impl BatchVotes {
             }
 
             // Check the signature before insertion.
-            d.verify()?;
-
-            self.add(d.clone(), weight);
-            stats.added(d.msg.number, weight);
+            if d.verify().is_ok() {
```
pompon0, line 133: please log the error

pompon0 (changes requested):
added comment

node/actors/network/src/gossip/batch_votes.rs
```diff
@@ -125,10 +130,12 @@ impl BatchVotes {
             }
 
             // Check the signature before insertion.
-            d.verify()?;
-
-            self.add(d.clone(), weight);
-            stats.added(d.msg.number, weight);
+            if let Err(e) = d.verify() {
+                tracing::error!(error =? e, "failed to verify batch vote");
+            } else {
```
pompon0, line 135: nit: I guess this operator logs error with "{:?}" formatting, which afair just shows the outermost message in case of anyhow::Errors (if it indeed is an anyhow error), "{:#}" is more bullet proof, so I recommend using it everywhere.

pompon0 (approved):

node/actors/network/src/gossip/batch_votes.rs
aakoshh: Yeah, it just doesn't work with KV pairs, or I don't know how to use it. I'll just log it as part of the message.

PR #172, opened 2024-08-02
chore: Remove deprecated cargo deny fields

aakoshh:
## What ❔

Follow up for https://github.com/matter-labs/era-consensus/pull/171

Removes fields which are deprecated according to https://github.com/matter-labs/era-consensus/actions/runs/10215658697/job/28265688492?pr=169 

See the link in the error messages to `cargo-deny` PR 611 which describes the new behaviour, and then the link to PR 606 which explains why the fields were removed. 

I don't think the deprecated fields need to be moved, the explanation says they made no sense; except for `[graph].target`.

## Why ❔

So that we can keep using the latest version of `cargo-deny` without failing the build.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,7 +37,6 @@ jobs:
         uses: baptiste0928/cargo-install@v2
         with:
           crate: cargo-deny
-          version: "0.15.1"
       - name: deny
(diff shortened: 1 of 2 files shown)


aakoshh:
Done in https://github.com/matter-labs/era-consensus/pull/173

PR #173, opened 2024-08-02
Fixed cargo deny config

brunoffranca:
## What ❔
Changes the config of cargo deny because a breaking version of deny was published.


brunoffranca:
Yup. You can see it in the template. https://github.com/EmbarkStudios/cargo-deny/blob/main/deny.template.toml

PR #175, opened 2024-08-02
Support for dynamic attester committee

pompon0:
It will support collecting votes for a single batch for now. I've extended the rpc to support both pushing and pulling the votes (with dynamic committee we cannot really collect future votes). I've reworked the API to concentrate the api surface in one place.

I've also moved the rpc registry from rust code to protobuf enum to make enforcing compatibility easier.

I'll prepare a corresponding PR in zksync-era before merging to make sure that the integration is smooth e2e.


brunoffranca:
Couple of notes:

1. If possible, I would prefer to have the attester code into its own actor. Otherwise it will be harder to understand. I also don't want the network actor to keep growing in size until it basically encapsulates all the code.
2. We can have yet another trait (Client) to be implemented on era. But we already have 3 (PayloadManager, PersistentBatchStore, PersistentBlockStore) and I want us to merge all these traits into one. The consensus repo really should have a single interface to interact with era. This will be better discussed on the offsite, but just keep in mind that you might not want to create more traits like this.

node/actors/network/src/testonly.rs
```diff
@@ -195,25 +177,42 @@ impl InstanceRunner {
...
+/// InstanceConfig
+pub struct InstanceConfig {
+    /// cfg
+    pub cfg: Config,
+    /// block_store
+    pub block_store: Arc<BlockStore>,
+    /// batch_store
+    pub batch_store: Arc<BatchStore>,
+    /// attestation_state
+    pub attestation_state: Arc<attestation::StateWatch>,
```
aakoshh, line 189: As far as I see with the removal of the loop in run, nothing will keep this up-to-date, right? It would be helpful to leave a comment of what to expect, as previously that loop simulated having a client to pull data, in this case from the local batch store, so as long as the stores agreed there was a chance that they would all be voting on the same thing, but now I'm not sure what's the deal with this.

node/libs/roles/src/attester/messages/batch.rs
```diff
@@ -160,46 +156,43 @@ impl BatchQC {
...
+        msg: &Signed<Batch>,
+        committee: &attester::Committee,
+    ) -> anyhow::Result<()> {
+        anyhow::ensure!(self.message == msg.msg, "inconsistent messages");
+        anyhow::ensure!(
+            !self.signatures.contains(&msg.key),
+            "signature already present"
         );
-
+        anyhow::ensure!(committee.contains(&msg.key), "not in committee");
```
aakoshh, line 169: What's the reason to stop using `Error::InconsistentMessage` and the other errors, but not to remove the enum?

node/libs/storage/src/batch_store/mod.rs
```diff
@@ -54,21 +54,6 @@ pub trait PersistentBatchStore: 'static + fmt::Debug + Send + Sync {
     /// Range of batches persisted in storage.
     fn persisted(&self) -> sync::watch::Receiver<BatchStoreState>;
 
-    /// Get the next L1 batch for which attesters are expected to produce a quorum certificate.
-    ///
-    /// An external node might never have a complete history of L1 batch QCs. Once the L1 batch is included on L1,
-    /// the external nodes might use the [attester::SyncBatch] route to obtain them, in which case they will not
-    /// have a QC and no reason to get them either. The main node, however, will want to have a QC for all batches.
-    async fn next_batch_to_attest(
```
aakoshh, line 62: What's the new way to query this? 👀

node/libs/storage/src/batch_store/mod.rs
```diff
@@ -365,27 +289,6 @@ impl BatchStore {
         Ok(())
     }
 
-    /// Wait until the database has a batch, then attach the corresponding QC.
-    #[tracing::instrument(skip_all, fields(l1_batch = %qc.message.number))]
-    pub async fn persist_batch_qc(&self, ctx: &ctx::Ctx, qc: attester::BatchQC) -> ctx::Result<()> {
```
aakoshh, line 370: How do we save batches now? 👀

node/actors/network/src/lib.rs
```diff
@@ -133,9 +128,6 @@ impl Runner {
                 Ok(())
             });
 
-            // Update QC batches in the background.
-            s.spawn(self.net.gossip.run_batch_qc_finder(ctx));
```
aakoshh, line 137: What looks for quorums now? 👀

node/actors/executor/src/attestation.rs
```diff
@@ -207,50 +41,23 @@ impl AttestationStatusRunner {
...
+                        .await
+                        .context("update_config")?;
                 }
+                Ok(None) => tracing::debug!("waiting for attestation config..."),
+                Err(error) => tracing::error!(
+                    ?error,
+                    "failed to poll attestation config, retrying later..."
+                ),
+            }
+            if let Err(ctx::Canceled) = ctx.sleep(self.poll_interval).await {
```
aakoshh, line 57: Isn't [this](https://github.com/matter-labs/era-consensus/blob/3bb3e190189dfdcf32e231b06087cec733dee676/node/actors/executor/src/attestation.rs#L36) enough?

node/actors/executor/src/attestation.rs
```diff
@@ -1,201 +1,35 @@
...
     /// The API might return an error while genesis is being created, which we represent with `None`
     /// here and mean that we'll have to try again later.
-    ///
-    /// The genesis hash is returned along with the new batch number to facilitate detecting reorgs
-    /// on the main node as soon as possible and prevent inconsistent state from entering the system.
-    async fn attestation_status(
-        &self,
-        ctx: &ctx::Ctx,
-    ) -> ctx::Result<Option<(attester::GenesisHash, attester::BatchNumber)>>;
+    async fn config(&self, ctx: &ctx::Ctx) -> ctx::Result<Option<Config>>;
```
aakoshh, line 17: I'm not convinced that the main node API should return the committee as part of the `Config` and the attesters should just take it, as opposed to executing the batch and deriving the attester committee from their VM. I might be misunderstanding what the `Client` is supposed to be, but the docs still say it's querying the main node.

I mean you could argue that the ultimate authority about the committee is the L1 contract where the signatures are going to be checked, so even if the main node would try to get it signed by a bogus committee it would be rejected, but it seems odd that we ask the main node for something that we (will) have computed locally.

If this is a temporary measure it's not well documented. I understand we have Proof-of-Authority and centrally set the weights, and that we don't have the integration with the smart contracts yet, so the information has to come from somewhere, but maybe there is a way to show this in the APIs and data paths without bundling up the status and the configuration.

aakoshh (commented):
Big change! 

It got me confused that all the persistence side and the querying of the batch store for the next thing to attest for is gone, and the PR description doesn't mention it at all. I suspect the reason is that you want to only call `wait_for_qc()` from `zksync-era` and push all the code over to that side which retrieves the data from the DB and saves the result. 

It's true that some of the methods were (apart from maintaining metrics) merely passing the data through, and it looks like we just want to take care of the gossiping of the data now, and forget the interface described [here](https://www.notion.so/matterlabs/L1-batches-and-Consensus-7348aef31a514940b7779e2293ba0da0), but one could also argue that so far the interface on the `zksync-era` side was relatively simple: implement the persistent store, start the runners, and you're done, now there will have to be more interaction about which method to call to update, which method to poll for changes, etc. Is there some general guideline in you mind about where to implement stuff when we have a choice? I guess we can discuss more on the offsite too.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+mod tests;
+
+/// Coordinate the attestation by showing the config as seen by the main node.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Config {
+    /// Batch to attest.
+    pub batch_to_attest: attester::Batch,
+    /// Committee that should attest the batch.
+    /// NOTE: the committee is not supposed to change often,
+    /// so you might want to use `Arc<attester::Committee>` instead
```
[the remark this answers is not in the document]
pompon0: This is a note to myself. I'll resolve it

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// to avoid extra copying.
+    pub committee: attester::Committee,
+}
+
+/// A [Watch] over an [AttestationStatus] which we can use to notify components about
+/// changes in the batch number the main node expects attesters to vote on.
```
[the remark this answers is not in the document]
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+                    .iter()
+                    .filter(|(k, _)| !old.votes.contains_key(k))
+                    .map(|(_, v)| v.clone())
+                    .collect(),
+            },
+        }
+    }
+
+    /// Verifies and adds a vote.
+    /// Noop if vote is not signed by a committee memver or already inserted.
```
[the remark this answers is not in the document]
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+            sigs.add(vote.key.clone(), vote.sig.clone());
+        }
+        Some(attester::BatchQC {
+            message: self.config.batch_to_attest.clone(),
+            signatures: sigs,
+        })
+    }
+}
+
+/// Received of state diffs.
```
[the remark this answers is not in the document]
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Noop if vote is not signed by a committee memver or already inserted.
+    /// Returns an error if genesis doesn't match or the signature is invalid.
+    fn insert_vote(&mut self, vote: Arc<attester::Signed<attester::Batch>>) -> anyhow::Result<()> {
+        anyhow::ensure!(
+            vote.msg.genesis == self.config.batch_to_attest.genesis,
+            "Genesis mismatch"
+        );
+        if vote.msg != self.config.batch_to_attest {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
pompon0: The only violation as of now is sending votes with bad genesis (which shouldn't even occur on the gossipnet). My implementation allows for sending future votes, which is slightly easier than monitoring what votes our peer is willing to accept. There will be always a race condition on either sending too old or too new votes, so I wanted to simplify it.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Watch of the attestation state.
+pub struct StateWatch {
+    key: Option<attester::SecretKey>,
+    state: Watch<Option<State>>,
+}
+
+impl fmt::Debug for StateWatch {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("StateWatch").finish_non_exhaustive()
```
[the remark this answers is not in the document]
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+            if !diff.is_empty() {
+                return Ok(diff);
+            }
+        }
+    }
+}
+
+/// Watch of the attestation state.
+pub struct StateWatch {
+    key: Option<attester::SecretKey>,
```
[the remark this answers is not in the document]
pompon0: True, it does more than that. I've renamed it to `attestation::Controller`. SG?

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+            config: Arc::new(config),
+            votes: im::HashMap::new(),
+            weight: 0,
+        };
+        if let Some(key) = self.key.as_ref() {
+            if new.config.committee.contains(&key.public()) {
+                let vote = key.sign_msg(new.config.batch_to_attest.clone());
+                // This is our own vote, so it always should be valid.
+                new.insert_vote(Arc::new(vote)).unwrap();
+            }
```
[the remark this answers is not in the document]
aakoshh: As my other comments show I realise that nothing explicitly says the committee and the batch hash are coming from the main node, so _technically_ we're just signing what the configuration says we need to be voting on, which _could_ be coming from our own database. I know we agreed `Client` is a convenience and `Config` or the `StateController` is the actual API.

So my more general observation is that we should document the expectations around the `Config`.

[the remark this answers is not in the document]
pompon0: afaic it wai: with this PR the semantics of votes gossiping is in fact more restrictive - every node is willing to gossip only votes which are matching the hash that it calculated on its own. It is true that the API has shifted from where it was before, but thanks to that it is smaller. If you want the signing to be done together with verification/hash computation, we can move it out to the zksync-era side (i.e. entirely out of era-consensus), but I disagree that this approach is any more rubber stamping than the previous implementation.

node/actors/network/src/gossip/runner.rs
```diff
@@ -58,21 +57,36 @@ impl rpc::Handler<rpc::push_validator_addrs::Rpc> for &PushServer<'_> {
...
+            .attestation_state
+            .insert_votes(req.votes.into_iter())
+            .await
+            .context("insert_votes()")
+        {
+            // Attestation feature is still evolving, so for forward
+            // compatibility we just ignore any invalid data.
+            // Once stabilized we will drop the connection instead of
+            // logging the error.
+            tracing::warn!("{err:#}");
```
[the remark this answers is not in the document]
pompon0: ok, makes sense, I'll make insert_vote more restrictive.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
pompon0: nvm, I've made insert_vote more restrictive.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+        anyhow::ensure!(
+            vote.msg.genesis == self.config.batch_to_attest.genesis,
+            "Genesis mismatch"
+        );
+        if vote.msg != self.config.batch_to_attest {
+            return Ok(());
+        }
+        let Some(weight) = self.config.committee.weight(&vote.key) else {
+            return Ok(());
+        };
```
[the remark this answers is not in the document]
pompon0: nvm, I've made it more restrictive

node/actors/network/src/gossip/runner.rs
```diff
@@ -181,42 +197,51 @@ impl Network {
...
+            s.spawn::<()>(async {
+                let push_batch_votes_client = push_batch_votes_client;
+                // Subscribe to what we know about the state of the whole network.
+                let mut recv = self.attestation_state.subscribe();
+                loop {
+                    let diff = recv.wait_for_diff(ctx).await?;
+                    let req = rpc::push_batch_votes::Req {
+                        // If the config has changed, we need to re-request all the votes
+                        // from peer that we might have ignored earlier.
+                        want_snapshot: Some(diff.config_changed),
```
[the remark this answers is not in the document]
pompon0: I thought about it, considered that a premature optimization. But if you want it, I'll just add it.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
aakoshh: I agree with the restrictiveness, I have no problem with it - it required some changes to the protocol to request the resubmission of the votes once the node knows what the hash is, so I had no choice but to implement it the way it was within the constraints of the design at the time. 

I am also not arguing with you about whether the implementation is more rubber stamping than before, simply because I haven't seen the full implementation, only this half of it. My point was that it was unclear from what we see here how the rest of the implementation will evolve. This is for historic reasons and the lack of an updated design documents: we have a "client" and an API on the main node serving some data, and it we haven't discussed how this client could evolve: currently it's called `AttestationStatusClient`; it could evolve into a more general consensus related "main node client", or it could stay special to the attestation like here - I know Bruno wanted to avoid fragmentation, so imagine if we needed 3-4 of these special clients. My _assumption_ was that whatever data it gets it will get from the API, because it's a "client" and we already had `BatchStore::batch_to_sign` - which I see we no longer have. Based on those assumptions it was pure rubber stamping. The previous implementation explicitly retrieved the thing to sign, so it was obviously not rubber stamping. 

IMO if the `Client` serves as a convenience then it should ideally be simple to understand and keep the implementer within the intended guard rails of the architecture. As I mentioned it could do this if we broke up the `config` method into 3 different methods to retrieve: 1) the current status from the main API, then if it changed the 2) batch to vote on once its available and 3) the committee for that batch, which should be available from the previous batch execution - this would also allow us to return `Arc` for the committee if we wanted by caching it. The `Runner` could exercise these methods and pass them atomically to the `update` and it could clearly be explained where each piece of data is expected to come from. Technically it would be the same, but more obvious.

node/actors/network/src/gossip/runner.rs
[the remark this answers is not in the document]
aakoshh: In practice it will probably not matter that much, if we only have a handful of attesters and a handful of gossip peers. I only thought it everyone added the main node as a static gossip peer, and it's the one announcing the new status (the first to have it), then it will probably always get all the old votes. 

OTOH I like how `want_snapshot` is just a boolean flag. Let's leave it like this and in the future we can add another field to carry the number, and decide how to interpret it.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
aakoshh: To your specific points: 

>  every node is willing to gossip only votes which are matching the hash 

Yes.

> that it calculated on its own.

This wasn't clear from the PR.

> the API has shifted from where it was before, but thanks to that it is smaller.

I assume we consider the API to include the `Client` and the `BatchStore` as well. It is true that it has less methods, which makes it have a smaller footprint, but it's also "bigger" in terms of having more responsibility of implementing undocumented assumptions in how it produces `Config`.

> If you want the signing to be done together with verification/hash computation, we can move it out to the zksync-era side (i.e. entirely out of era-consensus)

No no, I think it's good that it lives here. Perhaps one way to look at it is that if we can verbally explain what should happen, and we _can_ implement the outline of that algorithm in `era-consensus`, then we _should_, with as many of the rules as possible captured in the code, leaving only the holes that we _cannot_ implement in this repo to be filled in on the `zksync-era` side. That would be clear guideline IMO. Alternatively we could start treating it as a pure gossip channel that just carries data, with case-by-case decisions on where to put the rules.

node/actors/network/src/gossip/runner.rs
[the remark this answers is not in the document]
pompon0: I already updated it. Now it just sends the batch it wants the votes for.

node/actors/network/src/proto/gossip.proto
```diff
@@ -19,10 +19,18 @@ message PushValidatorAddrs {
...
 message PushBatchVotes {
+  // Requesting the peer to respond with votes for the batch. 
+  optional roles.attester.Batch want_votes_for = 2; // optional
   // Signed roles.validator.Msg.votes
   repeated roles.attester.Signed votes = 1;
 }
 
+message PushBatchVotesResp {
+  // Signed roles.validator.Msg.votes
+  // Empty if want_snapshot in request was not set.
```
aakoshh, line 30:
```suggestion
  // Empty if want_votes_for in request was not set.
```
Cool, `want_votes_for` sounds like it can be used as an FYI as well so we know what the remote party thinks even without any votes to send where this could be deducted from.

node/actors/network/src/testonly.rs
pompon0: I've added expected usage to the documentation of controller.

node/libs/roles/src/attester/messages/batch.rs
pompon0: ok, this method was fucked up - it was using and enum and wrapping it immediately into anyhow. Given that there is no use case for distinguishing the error programmatically, I'll remove the enum.

node/libs/storage/src/batch_store/mod.rs
pompon0: attestation::Controller api covers this

node/actors/network/src/lib.rs
pompon0: `Controller::wait_for_qc()` should be used. I've added documentation to `attestation::Controller`.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
pompon0: > Alternatively we could start treating it as a pure gossip channel that just carries data, with case-by-case decisions on where to put the rules.

Yes, this pr is shifting towards this direction. It abstracts from how/if the certificate should be persisted.

node/actors/network/src/proto/gossip.proto
pompon0: done

node/actors/executor/src/attestation.rs
aakoshh: I see the `Client` is gone, so this conversation is moot.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
aakoshh: I think an interface based on shared observable memory is a guideline we can certainly explore, ie. instead of us providing runners to poll the stores and place the data into watches, we can eschew as much as we can of such runners on this side and make it clear that the responsibility to maintain the watchable input state and to subscribe to the watchable output state lies with the zksync-era side. We have a lot of code that we own there, so we can maintain these runners ourselves, but hopefully the choice of where the code should live would be obvious in most cases.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+///     };
+///     ctrl.update_config(Arc::new(config.clone())).unwrap();
+///     s.spawn(async {
+///         if let Some(qc) = ctrl.wait_for_qc(ctx, config.batch_to_attest.number).await?;
+///         // Submit the certificate `qc` to the global registry
+///         ...
+///     });
+///     // Wait for the global registry to include the certificate.
+///     ...
+/// }
```
aakoshh, line 187: This example looks deceptively simple. 

What looked suspicious to me is that `wait_for_qc` might never complete if we miss the votes through gossip _and_ the `config.batch_to_attest.number` never changes to reflect the fact that the main node API now indicates that we should move on to vote on the next batch. Then I noticed the `s.spawn(...)` which makes this run in the background, so the main loop can move on. I assume `Wait for the global registry to include the certificate.` is supposed to prevent a tight loop, but it's not a given that every EN will have the certificate.

aakoshh: Perhaps two concurrent loops would be simpler to understand:
* one to get the next batch number to attest, wait for the batch, wait for the committee, update the status
* another that just keeps calling `wait_for_qc` with a batch it increments by 1 regardless of the status and saves whever it gets `Some`

node/actors/network/src/gossip/mod.rs
```diff
@@ -173,57 +165,4 @@ impl Network {
...
-
-        loop {
-            async {
-                // Wait until the status indicates that we're ready to sign the next batch.
-                let Some(batch_number) = sync::changed(ctx, &mut recv_status)
-                    .instrument(tracing::info_span!("wait_for_attestation_status"))
-                    .await?
-                    .next_batch_to_attest
-                else {
-                    return Ok(());
```
aakoshh, line 200: Oops, this should have been `continue` 🤔

node/actors/network/src/gossip/attestation/mod.rs
pompon0: >  but it's not a given that every EN will have the certificate.

By global registry, I mean L1/main node. LMK if you have a better word for this.

pompon0: I can make it 2 loops, yes.

node/tools/src/config.rs
```diff
@@ -259,24 +260,13 @@ impl Configs {
     pub async fn make_executor(
         &self,
         ctx: &ctx::Ctx,
-    ) -> ctx::Result<(executor::Executor, TestExecutorRunner)> {
+    ) -> ctx::Result<(executor::Executor, TestMemoryStorageRunner)> {
         let replica_store = store::RocksDB::open(self.app.genesis.clone(), &self.database).await?;
         let store = TestMemoryStorage::new(ctx, &self.app.genesis).await;
 
         // We don't have an API to poll in this setup, we can only create a local store based attestation client.
```
aakoshh, line 267:
```suggestion
```

node/actors/network/src/gossip/attestation/mod.rs
aakoshh: I think for the waiting part it could just say L1.

pompon0: > As discussed on the daily, Roman doesn't think waiting for the batch N to be on L1 before starting to collect votes for batch N+1 is feasible.

That won't be enough, because the current implementation doesn't support gossiping certificates for multiple batches in parallel (this doesn't change from the previous implementation), while we need to keep gossiping votes for batch N until it is actually committed to L1. We will address it later though.

node/actors/network/src/gossip/mod.rs
pompon0: :man_shrugging:

node/actors/network/src/gossip/attestation/mod.rs
aakoshh: I might be missing the obvious, but I thought the issue was that we didn't want to wait for N to be on L1 before we start gossiping about N+1. If we only wait until we have the certificate for N in the `l1_batches_consensus` table, then we can immediately start gossiping about N+1 as soon as the commitment is ready, regardless of whether it's submitted to L1 or not. At least this is the case when the main node is the only sender of the batches. You are right if we want anyone to be able to send the data then we have to gossip until the (unknown) sender(s) manage to get the certificate, which will require keeping the votes around until the batch is on L1.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+/// ```
+pub struct Controller {
+    key: Option<attester::SecretKey>,
```
RomanBrodetski, line 190: please add a comment what `None` and `Some` means here

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+        metrics::METRICS
+            .committee_size
+            .set(new.config.committee.len());
+        metrics::METRICS.votes_collected.set(new.votes.len());
+        #[allow(clippy::float_arithmetic)]
+        metrics::METRICS
+            .weight_collected
+            .set(new.weight as f64 / new.config.committee.total_weight() as f64);
+        locked.send_replace(Some(new));
+        Ok(())
```
RomanBrodetski, line 328: let's add comprehensive logging to this. Metrics are good, but they are only scrapped every 30 seconds, we may go through multiple batches during that time - that is, we won't know if we skipped something

RomanBrodetski (commented):
Some preliminary comments - diving deeper to the actual logic now

node/actors/network/src/gossip/attestation/mod.rs
pompon0: That will work as long as we use the main node as the "global registry". When L1 is the global registry, each node should wait until the batch commitment is finalized on L1 before ceasing to gossip the attestations for that batch. But I don't know how committing batches in parallel will work in a multi-submitter setup though.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    fn diff(&self, old: &Option<Self>) -> Diff {
+        let Some(old) = old.as_ref() else {
+            return Diff {
+                config: Some(self.config.clone()),
+                votes: self.votes.values().cloned().collect(),
+            };
+        };
+        if self.config.batch_to_attest.number != old.config.batch_to_attest.number {
+            return Diff {
+                config: Some(self.config.clone()),
```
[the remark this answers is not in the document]
pompon0: we fail even earlier in the update_config.

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// Committee that should attest the batch.
+    pub committee: Arc<attester::Committee>,
+}
+
+// Internal attestation state: config and the set of votes collected so far.
+#[derive(Clone)]
+struct State {
+    config: Arc<Config>,
+    votes: im::HashMap<attester::PublicKey, Arc<attester::Signed<attester::Batch>>>,
+    weight: attester::Weight,
```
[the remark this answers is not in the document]
pompon0: renamed to total_weight, added a comment

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+use zksync_concurrency::{ctx, sync};
+use zksync_consensus_roles::attester;
+
+mod metrics;
+#[cfg(test)]
+mod tests;
+
+/// Coordinate the attestation by showing the config as seen by the main node.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Config {
```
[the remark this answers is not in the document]
pompon0: I've updated the documentation of the struct.

node/actors/network/src/gossip/attestation/mod.rs
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+/// let ctrl = attestation::Controller::new(Some(key));
+/// loop {
+///     // Check the global attestation registry.
+///     // Compute the next expected batch and the committee that should attest it.
+///     ...
+///     let config = attestation::Config {
+///         batch_to_attest: ...,
+///         committee: ...,
+///     };
+///     ctrl.update_config(Arc::new(config.clone())).unwrap();
```
RomanBrodetski, line 179: IMO this is a very confusing interface. Why is the method that communicates the next batch details to the consensus subsystem called `update_config`? Also `Config` is not a good name for a struct belonging to the public interface

Something like
```
  attester_controller.start_batch_attestation(BatchAttesterInfo {number, committe})
```

pompon0: start_batch_attestation, ok. But how is "Info" better than "Config"? Info doesn't even indicate whether data is input or output of the system.

aakoshh: This is why I suggested that `update` could take multiple parameters and the `Config` could be an internal detail, perhaps?

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
aakoshh: Also the Narwhal implementation used in Sui had [reconfiguration channels](https://github.com/MystenLabs/narwhal/blob/e67f91530e6bd4ef7808e42f548f07e58764ec5b/worker/src/processor.rs#L29-L30) in all the services for [similar purposes](https://github.com/MystenLabs/narwhal/blob/e67f91530e6bd4ef7808e42f548f07e58764ec5b/types/src/primary.rs#L704-L708)

node/actors/network/src/gossip/attestation/mod.rs
RomanBrodetski: yes, I'll be much more comfortable with `Config` being internal to consensus. 
`Info` is better because it's more generic - `Config` associates with something configurable, in most cases configurable by operator. This is not the case here. as @aakoshh mentioned elsewhere, this name is common in literature, but not everyone working on era code has this background

node/actors/network/src/gossip/attestation/mod.rs
pompon0: done

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
pompon0: +1 to what @aakoshh says. The config in literature is at the granularity of an epoch, were in this specific piece of code epoch ~ 1 batch.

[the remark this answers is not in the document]
pompon0: throughout the era-consensus codebase I named "Config" everything that specifies how the given component should behave.

node/actors/executor/src/attestation.rs
pompon0: ptal at how the new api will be used in practice: https://github.com/matter-labs/zksync-era/pull/2657

node/actors/network/src/gossip/attestation/mod.rs
pompon0: I'm not convinced, but ok, `Info` it is.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
pompon0: renamed to info

node/actors/network/src/testonly.rs
pompon0: I've updated the example with more details. Added second loop.

node/actors/network/src/gossip/attestation/mod.rs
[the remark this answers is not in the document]
aakoshh: Just trying one more time: the actual interface is the controller which could have an update method to take the commuter and the batch as separate arguments. The other method in it is the wait_for_qc, which returns a BatchQC. The Config doesn’t have to appear in a public API, nor does any Info. Or does it?

Wouldn’t that keep everyone happy, if we used Config internally but hide it from the public eye?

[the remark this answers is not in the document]
pompon0: I'm more happy with Info, than passing a tuple.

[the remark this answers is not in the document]
aakoshh: Okay, for the record I didn't mean a tuple, just an old fashioned method like `Controller::update(&self, committee: Arc<Committee>, batch: Batch)`, and then perhaps the registry contract address we discussed, but if you prefer a single struct then fine.

(FWIW when we discussed putting the registry address into genesis we used the word "configuration" again. To me "info" is some read-only thing that we don't have a good name for, a bunch of fields about something. Don't have a better candidate though, so long live Info!)

aakoshh (approved):
Nice one 👍

PR #176, opened 2024-08-05
bumped capability ID of the push_batch_votes rpc

pompon0:
We added an extra field to the Batch message, without considering the fact that an extra field won't be recognized by old binaries and they will fail to verify signatures on Batch messages. They will consider sending such invalid signatures malicious and will disconnect from the peer. To avoid this we create a new RPC instead (by just bumping the capability id on push_batch_votes RPC) which will not be interpreted by the old binaries at all.

### DIFF
--- a/node/actors/network/src/noise/stream.rs
+++ b/node/actors/network/src/noise/stream.rs
@@ -48,7 +48,7 @@ const MAX_PAYLOAD_LEN: usize = MAX_TRANSPORT_MSG_LEN - AUTHDATA_LEN;
 /// <frame len:u16> ++ <frame data:[u8;len]>.
 ///
 /// Length of the frame len field.
-const LENGTH_FIELD_LEN: usize = std::mem::size_of::<u16>();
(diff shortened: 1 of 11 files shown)


aakoshh (approved):
Is there a chance to accidentally duplicate the capability IDs?

pompon0:
we have a test for this, but you can always forget to add the ID to the test

node/actors/network/src/noise/stream.rs
```diff
@@ -48,7 +48,7 @@ const MAX_PAYLOAD_LEN: usize = MAX_TRANSPORT_MSG_LEN - AUTHDATA_LEN;
 /// <frame len:u16> ++ <frame data:[u8;len]>.
 ///
 /// Length of the frame len field.
-const LENGTH_FIELD_LEN: usize = std::mem::size_of::<u16>();
```
aakoshh, line 51: It looks like `zksync-era` doesn't like this: https://github.com/matter-labs/zksync-era/actions/runs/10249428261/job/28352782628?pr=2583

PR #178, opened 2024-08-06
feat: instrument code with tracing spans

itegulov:
## What ❔

This PR instruments all interesting* places within the system.

**\*** "interesting" here is very subjective and I tried to use my understanding of the code to judge what should be covered and what shouldn't. Feel free to give feedback 

## Why ❔

Better observability


node/actors/bft/src/lib.rs
```diff
@@ -87,10 +90,17 @@ impl Config {
 
             tracing::info!("Starting consensus actor {:?}", cfg.secret_key.public());
 
-            // This is the infinite loop where the consensus actually runs. The validator waits for either
-            // a message from the network or for a timeout, and processes each accordingly.
-            loop {
-                let InputMessage::Network(req) = pipe.recv.recv(ctx).await?;
+            #[tracing::instrument(skip_all)]
```
pompon0, line 93: sry for this being kind of late, but explicit argument passing is in these "*_iter" functions, I actually like the async "{ ... }.instrument(...)" better

node/actors/executor/src/attestation.rs
```diff
@@ -58,56 +59,70 @@ impl AttesterRunner {
...
 
     /// Wait for the batch commitment to become available.
+    #[tracing::instrument(skip_all, fields(l1_batch = %number))]
     async fn wait_for_batch_to_sign(
-        &self,
         ctx: &ctx::Ctx,
         number: attester::BatchNumber,
+        batch_store: &BatchStore,
+        poll_interval: time::Duration,
     ) -> ctx::Result<attester::Batch> {
```
itegulov, line 116: This looks pretty ugly, but I had to remove `self` from here as it prevents us from using it in the `async` closure above

brunoffranca (approved):

node/libs/concurrency/src/ctx/mod.rs
```diff
@@ -185,8 +185,8 @@ impl Ctx {
     ) -> CtxAware<impl 'a + Future<Output = OrCanceled<F::Output>>> {
         CtxAware(async {
             tokio::select! {
-                output = fut => Ok(output),
```
pompon0, line 188: what happened here?

node/libs/concurrency/src/ctx/tests.rs
```diff
@@ -9,7 +9,7 @@ async fn test_run_canceled() {
     scope::run!(ctx, |ctx, s| async {
         s.cancel();
         assert!(!ctx.is_active());
-        Ok::<(), ()>(())
+        std::result::Result::Ok::<(), ()>(())
```
pompon0, line 12: why is it needed?

pompon0 (approved):

node/libs/concurrency/src/ctx/mod.rs
itegulov: I introduced `fn Ok<T>(t: T) -> Result<T>` just so I can use `ctx::Ok(())` for convenience when needed. This messes with the existing references to `Ok` in the same module so I had to explicitly clarify them.

node/libs/concurrency/src/ctx/tests.rs
itegulov: Same here

node/actors/network/src/gossip/mod.rs
```diff
@@ -109,20 +110,32 @@ impl Network {
...
-                });
+                s.spawn(
+                    async {
+                        let _permit = permit;
+                        let number = number.into();
+                        let queue_span = tracing::info_span!("wait_for_block_to_queue");
+                        let queue_span_copy = queue_span.clone();
+                        let _: ctx::OrCanceled<()> = scope::run!(ctx, |ctx, s| async {
+                            let request_span = tracing::info_span!("fetch_block_request");
+                            request_span.follows_from(queue_span_copy.id());
```
pompon0, line 121: is this explicit binding needed? We do the span forwarding here for every spawned task: https://github.com/matter-labs/era-consensus/blob/635d3b79c67fe2f01a91c70883dea864e844f928/node/libs/concurrency/src/scope/mod.rs#L92

pompon0: i.e. nested spawn inherits the span of the task it was spawned from.

itegulov: The way I understood the code originally made me think this closure can outlive the parent span, but looking again that might not be the case.

`follows_from` is a bit different from inheriting as it allows a span to outlive the parent (I believe by default parent closing also closes all of its children spans).

In any case, if you can confirm that they are nested I will revert to plain spans.

pompon0: I see, in this case the request_span won't outlive queue_span, because all scope's tasks have to complete before the scope completes. What would happen if the child span outlived the parent span though? Is that problematic?

itegulov: When you close the parent span, all of the child spans will also close. So nothing "problematic" per se, the child's life would just be misrepresented in the metrics.

PR #182, opened 2024-08-14
Reintroduced the lost span

pompon0:
It was removed by accident during the instrumentation refactor.
The span name is intentionally short, so that the logs don't become too verbose


aakoshh (approved):
These were very useful indeed!

PR #185, opened 2024-08-19
added build_version label to gossip handshake

pompon0:
It will allow us to monitor which nodes are using which version of the binary.
I've added a metric to count peers by their version (it is unbounded though).
I've added it to the http page display as well.
unrelated: I've removed Default impl for committees - it was creating invalid object.
unrelated: Exposed the validator::ProofOfPossession type.


node/actors/network/src/gossip/mod.rs
```diff
@@ -33,14 +33,25 @@ mod testonly;
 mod tests;
 mod validator_addrs;
 
+/// Info about a gossip connection.
+#[derive(Debug)]
+pub(crate) struct Connection {
+    /// Peer's public key.
+    pub(crate) key: node::PublicKey,
+    /// Build version of peer's binary (not verified).
+    pub(crate) build_version: Option<String>,
```
aakoshh, line 42: Could it be verified against semiweekly reflex, perhaps?

pompon0: > semiweekly reflex

what is that?

aakoshh: Sorry, autocomplete. Semver regex

aakoshh (approved):
Looks good but I would implement some protection against spam, like a semver or length limit 

I had to do a similar addition once to an Ethereum client, the format included the executable name, the platform and architecture and the version, eg. mantis/v3.0-cd5ae33/linux-amd64/ubuntu-openjdk64bitservervm-java-11.0.9

We don’t have multiple clients but I wouldn’t let anyone send unconstrained strings

pompon0:
The string is constrained by the Handshake message size. We will restrict it further if the need arises.

node/actors/network/src/gossip/mod.rs
pompon0: done

PR #191, opened 2024-09-06
feat: support for broadcasting unknown data

pompon0:
Re-encoding the protobuf message has a side effect of dropping unknown fields, which changes the hash of the message.
With this change we will be able to add data to the message broadcasted by the validator, in a backward compatible way.


brunoffranca:
Is this abandoned?

pompon0:
there is no use for it yet, but it is not abandoned.

PR #198, opened 2024-09-23
refactor: made TLS certificate optional for the debug page

pompon0:


brunoffranca:
Is it worth it to still keep the option of making the debug page public, or should we just force it to be local and then we can remove the whole TLS cert code?
People can still access the page remotely via SSH.

pompon0:
I think it would be good to be able to inspect debug page of the partner's EN in case they have problems.

brunoffranca (approved):

brunoffranca:
I doubt that will get used. The debug page doesn't even have that much information, and partners can also just let us connect via SSH. But the functionality is already here anyway.

PR #203, opened 2024-09-30
feat: support for syncing pre-genesis blocks

pompon0:
We will sync blocks with merkle proofs for the whole batches, which obsoletes the unused "batch syncing" logic that we had.

I've introduced a concept of PreGenesisBlock which contains a custom Justification, and a verify_pre_genesis_block() method in PersistentBlockStore for verifying those.
Some of the RPC requests required tuning to provide backward compatibility.

I've removed WIRE from protobuf compatibility check, since it is a subset of WIRE_JSON.


brunoffranca (commented):
Approach looks good. There is the inconsistency of post-genesis blocks being verified on era-consensus while pre-genesis blocks are verified on zksync-era. But this is probably only solvable when we move the eth watcher to era-consensus.

node/libs/storage/src/block_store/mod.rs
```diff
@@ -75,56 +116,46 @@ pub trait PersistentBlockStore: 'static + fmt::Debug + Send + Sync {
...
     /// `queue_next_block()` may return BEFORE the block is actually persisted,
     /// but if the call succeeded the block is expected to be persisted eventually.
     /// Implementations are only required to accept a block directly after the previous queued
     /// block, starting with `persisted().borrow().next()`.
-    async fn queue_next_block(
-        &self,
-        ctx: &ctx::Ctx,
-        block: validator::FinalBlock,
-    ) -> ctx::Result<()>;
+    async fn queue_next_block(&self, ctx: &ctx::Ctx, block: validator::Block) -> ctx::Result<()>;
```
brunoffranca, line 126:
```suggestion
    async fn store_next_block(&self, ctx: &ctx::Ctx, block: validator::Block) -> ctx::Result<()>;
```
I think you already renamed this on the metrics.

brunoffranca (approved):

node/libs/storage/src/block_store/mod.rs
pompon0: wdym? This is queue_next_block everywhere (as intended) afaict.

brunoffranca (approved):

PR #205, opened 2024-10-06
feat: Expand metrics in HTTP debug page

brunoffranca:
## What ❔

- Almost all info in Network was added to the debug page. Exceptions are values related to batch gossiping (which is going to be deprecated) and message pools/queues (which I deemed too verbose to include, but can be included if necessary).
- Fixed calculation of connection throughput. It now averages over the last minute.
- Made all units human readable.
- As a side effect, a couple of cargo files were formatted. The only real change was adding human_repr package. Will format the rest of the cargo files in a separate PR.

Fixes BFT-508


node/actors/network/src/debug_page/mod.rs
```diff
@@ -217,115 +220,448 @@ impl Server {
...
-            let received = stats.received.load(Ordering::Relaxed);
-            let sent = stats.sent.load(Ordering::Relaxed);
+            .with_header_row(vec!["", "Number", "Hash", "Genesis hash"]);
+
+        for (key, batch) in votes {
+            table.add_body_row(vec![
+                key.encode(),
+                batch.msg.number.to_string(),
+                batch.msg.hash.encode(),
+                batch.msg.genesis.encode(),
```
pompon0, line 552: genesis hash and msg hash are always the same for each vote, so they would obfuscate the displayed information. You can display the data about the message that we are collecting votes for separately.

also IMO it would be more informative to display the committee and votes tables together: row for attester + if we have their vote, so that it is clearly visible who didn't sent their vote yet.

node/actors/network/src/debug_page/mod.rs
```diff
@@ -217,115 +220,448 @@ impl Server {
...
-            })
+    /// Returns human readable duration. We use this function instead of `Duration::human_duration` because
+    /// we want to show days as well.
+    fn human_readable_duration(seconds: u64) -> String {
+        let days = seconds / 86400;
+        let hours = (seconds % 86400) / 3600;
+        let minutes = (seconds % 3600) / 60;
+        let seconds = seconds % 60;
+
+        format!("{}d {}h {}m {}s", days, hours, minutes, seconds)
```
pompon0, line 665: nit: consider stripping prefix values if they are 0

node/actors/network/src/metrics.rs
```diff
@@ -229,8 +230,12 @@ pub struct MeteredStreamStats {
     pub received: AtomicU64,
     /// System time since the connection started.
     pub established: SystemTime,
-    /// Ip Address and port of current connection.
+    /// IP Address and port of current connection.
     pub peer_addr: SocketAddr,
+    /// Circular buffer for sent bytes in the last minute.
+    sent_buffer: VecDeque<(Instant, u64)>,
```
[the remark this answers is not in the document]
brunoffranca: Precision will suck a bit, but ok.

node/actors/network/src/debug_page/mod.rs
brunoffranca: Is it guaranteed that it is always the same for all votes? Because the data structure doesn't seem to enforce that.

Ok on the tables though, I can merge.

node/actors/network/src/gossip/validator_addrs.rs
```diff
@@ -90,6 +90,13 @@ impl ValidatorAddrsWatch {
         self.0.subscribe()
     }
 
+    /// Copy of the current ValidatorAddrs state.
+    pub(crate) fn current(
+        &self,
```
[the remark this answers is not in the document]
brunoffranca: Not that I can find. It has borrow.

node/actors/network/src/metrics.rs
```diff
@@ -245,58 +253,58 @@ impl MeteredStreamStats {
...
-        }
+        let elapsed_secs = now.saturating_sub(self.current_minute_start.load(Ordering::Relaxed));
 
-        while let Some(&(timestamp, _)) = self.received_buffer.front() {
-            if timestamp < one_minute_ago {
-                self.received_buffer.pop_front();
-            } else {
-                break;
-            }
+        if elapsed_secs >= 60 {
```
pompon0, line 286: nit: I guess this also would work, but fyi I meant more sth like
```
let now = now().as_secs()/60;
if now>current_minute {
  if current_minute == now-1 {
    prev_minute_sent = current_minute_sent  
  } else {
    prev_minute_sent = 0
  }
  current_minute_sent = 0
  current_minute = now
}
```

node/actors/network/src/metrics.rs
```diff
@@ -245,58 +253,58 @@ impl MeteredStreamStats {
...
     }
 
     /// Returns the upload throughput of the connection in bytes per second.
     pub fn sent_throughput(&self) -> f64 {
-        self.sent_buffer
-            .iter()
-            .map(|&(_, amount)| amount)
-            .sum::<u64>() as f64
-            / 60.0
+        self.previous_minute_sent.load(Ordering::Relaxed) as f64 / 60.0
```
pompon0, line 303: note that if no data is transferred for a minute, then this value will not reset to 0. You should compare the current timestamp against the current_minute_start and based on that return either previous_minute_sent/current_minute_sent or 0.

node/actors/network/src/metrics.rs
```diff
@@ -245,58 +253,58 @@ impl MeteredStreamStats {
...
+            .as_secs();
 
-        while let Some(&(timestamp, _)) = self.sent_buffer.front() {
-            if timestamp < one_minute_ago {
-                self.sent_buffer.pop_front();
-            } else {
-                break;
-            }
-        }
+        let elapsed_secs = now.saturating_sub(self.current_minute_start.load(Ordering::Relaxed));
```
pompon0, line 284: if elapsed_secs is negative, then it means that the system time has been reset, in which case you should reset all the counters.

node/actors/network/src/metrics.rs
brunoffranca: Seems equivalent, unless we don't receive any packet for more than a minute. But then a 1m window is not appropriate anyway.

node/actors/network/src/metrics.rs
```diff
@@ -245,58 +253,58 @@ impl MeteredStreamStats {
...
+        self.current_minute_sent
+            .fetch_add(amount, Ordering::Relaxed);
     }
 
-    fn cleanup_buffers(&mut self) {
-        let one_minute_ago = Instant::now() - Duration::from_secs(60);
+    fn update_minute(&self) {
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
```
pompon0, line 281: well this would mean that we went back before 1970 XD

brunoffranca: So the error message is indeed correct!

node/actors/network/src/metrics.rs
pompon0: note that alternatively you can use the monotone clock (i.e. Instant as in previous version) and measure time relative to the time when the socket has been opened (i.e. call time::now() in constructor, and then use this timestamp instead of UNIX_EPOCH).

node/actors/network/src/metrics.rs
brunoffranca: Or I can just call update_minute() in the beginning.

pompon0 (dismissed):

node/actors/network/src/metrics.rs
pompon0: that might cause race condition when running update_minute() minute from read()/write() and http rendering function. If you want to do that, then put a mutex instead of bunch of atomic vars.

PR #209, opened 2024-10-15
feat: ChonkyBFT types

brunoffranca:
Implemented all types for ChonkyBFT. Work on this is all on the roles crate. Tests in roles crate are passing. Note that I'm not keeping the old types, that's because `ReplicaCommit` has no changes to it (I checked and serialization is the same for old and new, old signatures still work) as well as `CommitQC` and `FinalBlock`. Any type that an EN could receive, or that ends on our database, doesn't suffer any alterations. So it seems that the upgrade to ChonkyBFT is backwards-compatible for ENs. Since we only have one validator, we can upgrade the whole network without a planned hard fork.

Part of BFT-452


brunoffranca:
I still need to add tests for some new error types that I introduced.

node/libs/roles/src/proto/validator/messages.proto
```diff
@@ -39,57 +39,69 @@ message Block {
...
-  oneof t {// required
-    ReplicaPrepare replica_prepare = 1;
+  reserved 1, 3, 4;
+  reserved "replica_prepare", "leader_prepare", "leader_commit";
+
+  oneof t { // required
     ReplicaCommit replica_commit = 2;
-    LeaderPrepare leader_prepare = 3;
-    LeaderCommit leader_commit = 4;
+    ReplicaTimeout replica_timeout = 5;
```
pompon0, line 53: Please use a separate message type for chonkybft messages (i.e. sth like ConsensusMsgV2, or ChonkyMsg).

node/libs/roles/src/validator/messages/committee.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub fn total_weight(&self) -> u64 {
+        self.total_weight
+    }
+}
+
+/// Validator representation inside a Committee.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WeightedValidator {
+    /// Validator key
+    pub key: validator::PublicKey,
```
pompon0, line 158: let's add proof-of-possession to this new version of the protocol.

pompon0:
Please do not remove the fastest hotstuff logic, we need both to coexist for the time of the migration.

node/libs/roles/src/proto/validator/messages.proto
brunoffranca: Why?

brunoffranca:
Why do we need the old logic?

pompon0:
for transition?

brunoffranca:
But as I've wrote in the description, the change is backwards compatible for ENs. And we only have one validator, so it doesn't need to be backwards compatible for validators. Unless I'm missing something here, it seems like it should work.

node/libs/roles/src/validator/messages/committee.rs
brunoffranca: That would make it backwards incompatible with ENs, which I'm trying to avoid. Better to add it after we remove the validator committee from Genesis.

node/libs/roles/src/validator/messages/leader_proposal.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// This proposal is being proposed after a view where we finalized a block.
+    /// A commit QC is just a collection of commit votes (with at least
+    /// QUORUM_WEIGHT) for the previous view. Note that the commit votes MUST
+    /// be identical.
+    Commit(CommitQC),
+    /// This proposal is being proposed after a view where we timed out.
+    /// A timeout QC is just a collection of timeout votes (with at least
+    /// QUORUM_WEIGHT) for the previous view. Unlike with the Commit QC,
+    /// timeout votes don't need to be identical.
+    /// The first proposal, for view 0, will always be a timeout.
```
pompon0, line 124: I think it is in fact a good idea to make view 0 always time out, as this will be the initial synchronization point of the consensus. However please make sure that this doesn't create an artificial delay - i.e. validators should immediately cast timeout vote for view 0 on startup, rather than waiting for the actual timeout delay.

node/libs/roles/src/validator/messages/tests.rs
```diff
@@ -284,43 +269,74 @@ mod version1 {
...
+        let key = validator_keys()[0].clone();
+
+        (|| {
+            // Decode hash and signature.
+            let hash: MsgHash = Text::new(hash).decode()?;
+            let sig: Signature = Text::new(sig).decode()?;
+
+            // Check if msg.hash() is equal to hash.
+            if msg.hash() != hash {
+                return Err(anyhow::anyhow!("Hash mismatch"));
```
pompon0, line 322: anyhow::bail!()

pompon0 (approved):

PR #210, opened 2024-10-22
feat: ChonkyBFT logic

brunoffranca:
- Everything in chonky_bft folder is basically new.
- UTHarness was moved to chonky_bft/testonly.rs. Several changes to it though.
- All tests that use UTHarness were moved to chonky_bft/tests. Some of the old unit tests were repurposed, but a fair amount of the tests in chonky_bft/tests are new.
- The tests in tests.rs were split between tests/mod.rs and tests/twins.rs (except for a few that used UTHarness and were moved as said before). There were no changes to them though.

Part of BFT-452


node/actors/bft/src/chonky_bft/tests/proposer.rs
```diff
@@ -0,0 +1,40 @@
...
+            }
+        });
+
+        //util.produce_block(ctx).await;
+
+        Ok(())
+    })
+    .await
+    .unwrap();
+}
```
brunoffranca, line 40: I've tried to create some unit test for the proposer loop, mainly to check that the proposer pipe worked correctly. But no luck, maybe you can help. If not, I will delete it since the Network and Twins tests already test this loop indirectly anyway.

node/actors/bft/src/chonky_bft/misc.rs
```diff
@@ -0,0 +1,105 @@
...
+        ctx: &ctx::Ctx,
+        qc: &validator::CommitQC,
+    ) -> ctx::Result<()> {
+        self.high_commit_qc = max(Some(qc.clone()), self.high_commit_qc.clone());
+        self.save_block(ctx, qc).await.wrap("save_block()")
+    }
+
+    /// Tries to build a finalized block from the given CommitQC. We simply search our
+    /// block proposal cache for the matching block, and if we find it we build the block.
+    /// If this method succeeds, it sends the finalized block to the executor.
```
pompon0, line 38: there is no "sending to executor" any more

node/actors/bft/src/chonky_bft/misc.rs
```diff
@@ -0,0 +1,105 @@
...
+            high_vote: self.high_vote.clone(),
+            high_commit_qc: self.high_commit_qc.clone(),
+            high_timeout_qc: self.high_timeout_qc.clone(),
+            proposals,
+        };
+        self.config
+            .replica_store
+            .set_state(ctx, &backup)
+            .await
+            .wrap("put_replica_state")?;
```
pompon0, line 102: replica_store.set_state

node/actors/bft/src/chonky_bft/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub(crate) timeout_qcs_cache: BTreeMap<validator::ViewNumber, validator::TimeoutQC>,
+
+    /// The deadline to receive an input message before timing out.
+    pub(crate) timeout_deadline: time::Deadline,
+    /// Time when the current phase has started. Used for metrics.
+    pub(crate) phase_start: time::Instant,
+}
+
+impl StateMachine {
+    /// The duration of the timeout.
```
pompon0, line 73: "view timeout"?

node/actors/bft/src/chonky_bft/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+                                    tracing::debug!("on_proposal: {err:#}");
+                                }
+                                _ => {
+                                    tracing::warn!("on_proposal: {err:#}");
+                                }
+                            }
+                            Err(())
+                        }
+                    };
+                    metrics::ConsensusMsgLabel::ReplicaPrepare.with_result(&res)
```
pompon0, line 182: labels do not match

node/actors/bft/src/chonky_bft/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+                }
+                ConsensusMsg::ReplicaTimeout(_) => {
+                    let res = match self
+                        .on_timeout(ctx, req.msg.cast().unwrap())
+                        .await
+                        .wrap("on_timeout()")
+                    {
+                        Ok(()) => Ok(()),
+                        Err(err) => {
+                            match err {
```
pompon0, line 219: either use "err" or "e" consistently. It is ok to shadow the variable.

node/actors/bft/src/chonky_bft/mod.rs
brunoffranca: Didn't do metrics yet.

node/actors/bft/src/chonky_bft/proposal.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let block_hash = match implied_block_hash {
+            // This is a reproposal. We let the leader repropose blocks without sending
+            // them in the proposal (it sends only the number + hash). That allows a
+            // leader to repropose a block without having it stored.
+            // It is an optimization that allows us to not wait for a leader that has
+            // the previous proposal stored (which can take 4f views), and to somewhat
+            // speed up reproposals by skipping block broadcast.
+            // This only saves time because we have a gossip network running in parallel,
+            // and any time a replica is able to create a finalized block (by possessing
```
pompon0, line 136: This comment is misleading. Reproposals are critical for liveness, they are not an optimizations. We repropose without payload, because we don't gossip non-finalized payloads (and that part is indeed an optimization).

node/actors/bft/src/chonky_bft/proposer.rs
```diff
@@ -0,0 +1,103 @@
+use crate::{metrics, Config, OutputSender};
+use std::sync::Arc;
+use zksync_concurrency::{ctx, error::Wrap as _, sync, time};
+use zksync_consensus_network::io::ConsensusInputMessage;
+use zksync_consensus_roles::validator;
+
+/// Timeout for creating a proposal. If the proposal is not created in this time, the proposer
+/// will quit trying to create a proposal for this view. This can be different from the replica
+/// timeout for the whole view.
+pub(crate) const PROPOSAL_CREATION_TIMEOUT: time::Duration = time::Duration::milliseconds(2000);
```
pompon0, line 10: why do we need this? Wouldn't just the view timeout work?

node/actors/bft/src/chonky_bft/proposer.rs
```diff
@@ -0,0 +1,103 @@
...
+            cfg.clone(),
+            justification,
+        )
+        .await
+        {
+            Ok(proposal) => proposal,
+            Err(err) => {
+                tracing::error!("failed to create proposal: {}", err);
+                continue;
+            }
```
pompon0, line 44: treat timeout separately from internal errors.

node/actors/bft/src/chonky_bft/proposer.rs
brunoffranca: It can. I thought it might be useful in the future for sticky leaders. If we want to pipeline block production.

node/actors/bft/src/chonky_bft/proposer.rs
```diff
@@ -0,0 +1,103 @@
...
+    cfg: Arc<Config>,
+    justification: validator::ProposalJustification,
+) -> ctx::Result<validator::LeaderProposal> {
+    // Get the block number and check if this must be a reproposal.
+    let (block_number, opt_block_hash) = justification.get_implied_block(cfg.genesis());
+
+    let proposal_payload = match opt_block_hash {
+        // There was some proposal last view that a subquorum of replicas
+        // voted for and could have been finalized. We need to repropose it.
+        Some(_) => None,
```
pompon0, line 68: nit: unnecessary indentation, you can write
```
if opt_block_hash.is_some() { return ... }
```

node/actors/bft/src/chonky_bft/proposer.rs
brunoffranca: Although, if we timeout in block proposal, that's likely an unexpected error.

node/actors/bft/src/lib.rs
```diff
@@ -71,56 +56,39 @@ impl Config {
...
-                                ack,
-                            };
 
-                            replica_send.send(new_req);
-                            leader_send.send(req);
-                        }
-                        M::ReplicaCommit(_) => leader_send.send(req),
-                        M::LeaderPrepare(_) | M::LeaderCommit(_) => replica_send.send(req),
-                    }
+                    replica_send.send(msg);
```
pompon0, line 91: now that we have just one message consumer, it doesn't make sense to do this extra message passing. Either inline the replica logic here, or pass the message receiver to replica process directly.

node/actors/bft/src/chonky_bft/commit.rs
```diff
@@ -1,43 +1,58 @@
-//! Handler of a ReplicaCommit message.
```
pompon0, line 1: "misc" doesn't mean anything. If you cannot find a good place for this logic, then just move it to the parent module.

node/actors/bft/src/chonky_bft/timeout.rs
```diff
@@ -0,0 +1,197 @@
...
+        self.outbound_pipe.send(output_message.into());
+
+        // Log the event.
+        tracing::info!("Timed out at view {}", self.view_number);
+        metrics::METRICS.replica_view_number.set(self.view_number.0);
+
+        // Reset the timeout. This makes us keep sending timeout messages until the consensus progresses.
+        // However, this isn't strictly necessary since the network retries messages until they are delivered.
+        // This is just an extra safety measure.
+        self.timeout_deadline = time::Deadline::Finite(ctx.now() + Self::TIMEOUT_DURATION);
```
pompon0, line 193: Can you make it actually send the timeout once? This is not the responsibility of the chonky module to retry sending messages.

node/libs/roles/src/validator/messages/leader_proposal.rs
```diff
@@ -26,47 +24,7 @@ impl LeaderProposal {
...
-        }
-
-        if let Some(payload_hash) = implied_payload {
-            if self.proposal.payload != payload_hash {
-                return Err(LeaderProposalVerifyError::BadPayloadHash {
-                    got: self.proposal.payload,
-                    want: payload_hash,
-                });
-            }
-        }
```
pompon0, line 48: This check is still relevant. Please add a test for it.

node/libs/roles/src/proto/validator/messages.proto
```diff
@@ -72,9 +72,8 @@ message ReplicaNewView {
 }
 
 message LeaderProposal {
```
pompon0, line 74: Add reserved field.

node/actors/bft/src/chonky_bft/misc.rs
```diff
@@ -0,0 +1,105 @@
+use super::StateMachine;
```
pompon0, line 1: "misc" doesn't mean anything. If you cannot find a good place for this logic, then just move it to the parent module.

node/actors/bft/src/chonky_bft/proposer.rs
pompon0: I mean that internal error should be critical and should shut down the component.

node/actors/bft/src/chonky_bft/tests/proposer.rs
pompon0: I think that testing happy path here without running the actor doesn't make much sense.

node/actors/bft/src/chonky_bft/timeout.rs
brunoffranca: I can. Note that we will need to keep the three phases, instead of reducing to two like we are discussing in Slack.

node/libs/roles/src/proto/validator/messages.proto
brunoffranca: Missed it.

node/actors/bft/src/chonky_bft/misc.rs
brunoffranca: I only wanted to have start and run on the parent module... What if I rename it to utils? I also accept other names.

node/actors/bft/src/chonky_bft/proposal.rs
brunoffranca: Ok, I'll reword it to be more clear.

brunoffranca:
One thing I wanted to note is that I don't actually check anywhere that reproposals don't have a payload set. This technically is not necessary since the consensus just ignores the payload. But maybe we want to do that check anyway.

node/actors/bft/src/lib.rs
brunoffranca: There still needs to be some conversion since the input channel to the actor is a simple mpsc channel, while the input channel to the replica process is a prunable_mpsc.
A more "radical" solution would be just to get rid of the whole actor pipe situation. It was only created because I thought we would have many components, in reality we ended up with just two big components. So I can refactor that quickly.

node/libs/roles/src/proto/validator/messages.proto
brunoffranca: Wait. It shouldn't be necessary. This is a new type. I only added it in the previous PR, and that's not merged into main yet.

node/actors/bft/src/lib.rs
pompon0: up to you, as you say, we will need to clean it up eventually.

node/actors/bft/src/chonky_bft/timeout.rs
pompon0: I would vote for having 3 phases then.

node/actors/bft/src/chonky_bft/proposer.rs
pompon0: I'd remove it for now then, but up to you

node/actors/bft/src/chonky_bft/misc.rs
pompon0: "utils" is as bad as "misc", it doesn't say anything. If you don't want to figure out a better layout, I'll take care of it later.

pompon0:
Yes, for defence in depth, please check that payload is included if and only if it is required.

node/actors/bft/src/lib.rs
brunoffranca: I'll do it in a separate PR.

node/actors/bft/src/chonky_bft/misc.rs
brunoffranca: Done.

PR #213, opened 2024-11-08
refactor: Removing actor pipes and naming

brunoffranca:
Getting rid of the whole actor pipe architecture. It was only created because we thought we would have many components, in reality we ended up with just two big components. This is just a refactor, no logic changes.


node/components/bft/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Creates a new input channel for the network messages.
+    pub fn create_input_channel() -> (
+        sync::prunable_mpsc::Sender<FromNetworkMessage>,
+        sync::prunable_mpsc::Receiver<FromNetworkMessage>,
+    ) {
+        sync::prunable_mpsc::channel(
+            Self::inbound_filter_predicate,
+            Self::inbound_selection_function,
+        )
+    }
```
pompon0, line 114: please make it global, there is no reason for it to be a part of Config API.

node/components/bft/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Filter predicate for incoming messages.
+    fn inbound_filter_predicate(new_req: &FromNetworkMessage) -> bool {
+        // Verify message signature
+        new_req.msg.verify().is_ok()
+    }
+
+    /// Selection function for incoming messages.
+    fn inbound_selection_function(
+        old_req: &FromNetworkMessage,
+        new_req: &FromNetworkMessage,
```
pompon0, line 125: nit: ditto for filter and selection

node/components/network/src/gossip/tests/mod.rs
```diff
@@ -518,12 +519,16 @@ async fn test_batch_votes_propagation() {
...
-                    .into(),
-            });
+            let (node, runner) = testonly::Instance::new_from_config(
+                testonly::InstanceConfig {
+                    cfg: cfg.clone(),
+                    block_store: store.blocks.clone(),
+                    attestation: attestation::Controller::new(Some(setup.attester_keys[i].clone()))
+                        .into(),
+                },
+                |_| true,
```
pompon0, line 529: nit: annotate these lambda arguments, like: `/*filter_predicate=*/|_| true, /*selection_function=*/...` for readability.

node/components/network/src/testonly.rs
```diff
@@ -78,15 +78,19 @@ pub(crate) async fn forward(
...
 
-/// Node instance, wrapping the network actor state and the
+/// Node instance, wrapping the network component state and the
 /// events channel.
 pub struct Instance {
     /// State of the instance.
-    pub(crate) net: Arc<Network>,
+    pub net: Arc<Network>,
     /// Termination signal that can be sent to the node.
-    pub(crate) terminate: channel::Sender<()>,
```
pompon0, line 87: Instance has a public method `terminate()` to interact with this field, why do you want to make this field public? In fact, is there anything blocking it from being private?

node/components/network/src/testonly.rs
```diff
@@ -190,27 +194,66 @@ pub struct InstanceConfig {
...
-            block_store,
-            attestation: attestation::Controller::new(None).into(),
-        })
+        Self::new_from_config(
+            InstanceConfig {
+                cfg,
+                block_store,
+                attestation: attestation::Controller::new(None).into(),
+            },
+            |_| true,
```
pompon0, line 203: ditto: `/*con_filter_predicate=*/..., /*con_selection_function=*/...`

node/components/network/src/testonly.rs
```diff
@@ -190,27 +194,66 @@ pub struct InstanceConfig {
...
 
     /// Construct an instance for a given config.
-    pub fn new_from_config(cfg: InstanceConfig) -> (Self, InstanceRunner) {
-        let (actor_pipe, dispatcher_pipe) = pipe::new();
+    pub fn new_from_config(
+        cfg: InstanceConfig,
+        con_filter_predicate: impl 'static + Sync + Send + Fn(&ConsensusReq) -> bool,
+        con_selection_function: impl 'static
+            + Sync
+            + Send
```
pompon0, line 236: oof, that's really ugly. How about we require a preconstructed prunable_mpsc instead?

pompon0: and for the "keep_all" channel (which you use in some tests - please minimize the amount of tests that use it, so that tests are as close to reality as possible), you can add a new constructor to the prunable_mpsc directly, so that the intention is clear (instead of inlining the corresponding predicates in some places).

pompon0 (commented):
LG in general, added some comments

node/components/network/src/testonly.rs
brunoffranca: I need the following two fields to be public and clippy complains about mixed visibility in structs. I'll add an exception here.

node/components/network/src/testonly.rs
brunoffranca: The predicates are defined in the bft crate, so I can't import them into network. What I can do is move the predicates to the concurrency crate (or maybe the roles crate).

brunoffranca: I can move them into network, but that seems worse...

pompon0: no, I didn't mean the real predicates. Just add
```
pub fn keep_all_channel<T>() -> (Sender<T>, Receiver<T>) {
  channel(||true,|_, _| SelectionFunctionResult::Keep)
}
```
to prunable_mpsc module.

pompon0:
The advantage is the fact that you are aware of crates that use old dependencies and you can try to avoid them. It also actively prevents people from pinning dependency versions.

node/components/network/src/testonly.rs
brunoffranca: That I got, yes. It's already in there. But you asked to minimize the amount of tests that use it. However, all tests in network must use the unpruned channel. I can (and do) use the pruned channel for all tests in bft.

brunoffranca:
We have never used a different dependency because of this. If it doesn't actually result in different behavior, then it's not useful information IMO.

node/components/network/src/testonly.rs
pompon0: ok

pompon0 (approved):

PR #214, opened 2024-11-12
fix: Add wrap around to table cells on Debug page

brunoffranca:


node/components/network/src/debug_page/style.css
```diff
@@ -1,91 +1,103 @@
...
+  padding: .5em;
+}
+
+table th,
+table td {
+  padding: .5em;
+  border: 1px solid lightgrey;
+  max-width: 600px;
+  word-wrap: break-word;
+  overflow-wrap: break-word;
```
brunoffranca, line 69: Formatting obscures the diff. Only these 3 lines were added.
I'm trying to test the debug page, unfortunately I'm having some trouble building the docker image.

pompon0 (approved):

PR #217, opened 2024-11-21
fix: Use justifications directly in the informal spec

dnkolegov:
This PR slightly improves the spec in Rust.


konnov:
I propose to fix the Quint spec in #218 instead, to keep in sync with the current version

PR #218, opened 2024-11-27
refactor: use create_justification in the quint specification as in the informal spec

konnov:
## What ❔

- Update the Quint specification to call `create_justification` in `start_new_view`, instead of passing justifications directly.
- Replace `get_justification(view)` with `create_justification()` in `replica.rs` to reflect the discussed changes.

## Why ❔

The pseudo code specification uses `create_justification` in `start_new_view` and `proposer_step`. The Quint specification was passing a justification directly in `start_new_view`. These two approaches are not entirely equivalent.

A falsy invariant `high_commit_and_timeout_qc_example` demonstrates that it's possible for a replica to have both `high_timeout_qc` and `high_commit_qc` to have the same view. The model checker found an execution that demonstrates it.


konnov:
It's interesting that the Typos job complains about a spell checking issue in a file that is not touched by this PR:

> error: `parametrized` should be `parameterized`
>  --> ./node/libs/concurrency/src/scope/state.rs:3:59

brunoffranca (dismissed):

konnov:
@brunoffranca this PR is going through now. We had an older version of Quint pinned, which was failing the tests. The latest version works.

brunoffranca (approved):

PR #220, opened 2024-12-11
fix: TimeoutQC aggregation

pompon0:
Custom Ord implementation was causing timeout votes with different high qc (with the same view) to be deduplicated, which invalidated the aggregated signature. I've added a test for this case, and preemptively removed all custom Ord implementations on messages.

### DIFF
--- a/node/components/bft/src/chonky_bft/mod.rs
+++ b/node/components/bft/src/chonky_bft/mod.rs
@@ -1,6 +1,5 @@
 use crate::{metrics, Config, FromNetworkMessage, ToNetworkMessage};
 use std::{
-    cmp::max,
     collections::{BTreeMap, HashMap},
     sync::Arc,
 };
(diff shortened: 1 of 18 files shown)


node/components/bft/src/chonky_bft/mod.rs
```diff
@@ -276,7 +275,14 @@ impl StateMachine {
         ctx: &ctx::Ctx,
         qc: &validator::CommitQC,
     ) -> ctx::Result<()> {
-        self.high_commit_qc = max(Some(qc.clone()), self.high_commit_qc.clone());
+        if self
+            .high_commit_qc
+            .as_ref()
+            .map_or(false, |old| old.view().number >= qc.view().number)
```
brunoffranca, line 281:
```suggestion
            .map_or(false, |old| old.view().number > qc.view().number)
```
Nit: Greater than should be enough. If we have two QCs at the same view, they must be for the same block.

node/components/bft/src/chonky_bft/proposal.rs
```diff
@@ -223,6 +221,7 @@ impl StateMachine {
 
         // Update the state machine.
         self.view_number = message.view().number;
+        metrics::METRICS.replica_view_number.set(self.view_number.0);
```
brunoffranca, line 224: Why move the start of the view to here?

brunoffranca (dismissed):

node/components/bft/src/chonky_bft/proposal.rs
pompon0: I'm moving the metrics updates next to the places where the variable that they track is updated. In particular, before this pr one of such places has been missed.

node/components/bft/src/chonky_bft/mod.rs
pompon0: if old view == qc view, there is nothing to be done, so IMO this WAI

PR #221, opened 2024-12-11
refactor: Removed attester set from genesis

brunoffranca:


pompon0 (commented):
lgtm, except that it shouldn't get merged before zksync-era stops using attester committee from genesis in tests and on CI.

PR #222, opened 2024-12-12
feat: Added view timeout duration as a config parameter

brunoffranca:
## What ❔

The hardcoded 2s view timeout was causing problems in Era testnet because the block time there is longer (>5s). View timeout needs to be a configurable parameter since different chains will want different block times.

### DIFF
--- a/node/components/bft/src/chonky_bft/proposer.rs
+++ b/node/components/bft/src/chonky_bft/proposer.rs
@@ -1,4 +1,3 @@
-use super::VIEW_TIMEOUT_DURATION;
 use crate::{metrics, Config, ToNetworkMessage};
 use std::sync::Arc;
 use zksync_concurrency::{ctx, error::Wrap as _, sync};
@@ -27,7 +26,7 @@ pub(crate) async fn run_proposer(
 
         // Create a proposal for the given justification, within the timeout.
         let proposal = match create_proposal(
-            &ctx.with_timeout(VIEW_TIMEOUT_DURATION),
+            &ctx.with_timeout(cfg.timeout_duration),
             cfg.clone(),
             justification,
         )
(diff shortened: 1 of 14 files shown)


node/components/bft/src/chonky_bft/proposer.rs
```diff
@@ -27,15 +26,15 @@ pub(crate) async fn run_proposer(
...
+            &ctx.with_timeout(cfg.view_timeout),
             cfg.clone(),
             justification,
         )
         .await
         {
             Ok(proposal) => proposal,
             Err(ctx::Error::Canceled(_)) => {
-                tracing::error!("run_proposer(): timed out while creating a proposal");
+                tracing::warn!("run_proposer(): timed out while creating a proposal");
```
pompon0, line 37: maybe just info? Afaiu it is expected to have infrequent blocks on some chains.

node/components/executor/src/lib.rs
```diff
@@ -39,6 +39,8 @@ pub struct Config {
     pub public_addr: net::Host,
     /// Maximal size of the block payload.
     pub max_payload_size: usize,
+    /// The duration of the view timeout, in milliseconds.
+    pub view_timeout: usize,
```
pompon0, line 43: why removing strong typing?, use time::Duration

node/tools/src/proto/mod.proto
```diff
@@ -98,6 +98,9 @@ message AppConfig {
   // Maximal size of the block payload.
   optional uint64 max_payload_size = 5; // required; bytes
 
+  // The duration of the view timeout.
+  optional uint64 view_timeout = 19; // required; milliseconds
```
pompon0, line 102: std.Duration

node/components/bft/src/chonky_bft/proposer.rs
brunoffranca: True, but maybe in that case the chain should be warned to increase the timeout. Overall, I don't think having a lot of timeouts should be considered normal operation.

node/components/executor/src/lib.rs
brunoffranca: I thought we didn't have Duration in Proto. I'll fix it.

pompon0 (dismissed):

PR #223, opened 2024-12-13
feat: loadtest improvements

pompon0:
Removed verification of redundant NewView messages.
Fixed incompatibility between zksync_conncurrency and `tokio::time::pause()`.
Added in-memory mode to the test node binary.
Fixed `payload_size` flag in the test node binary.
Removed obsolete rate limits from the network config.

### DIFF
--- a/node/components/bft/src/chonky_bft/new_view.rs
+++ b/node/components/bft/src/chonky_bft/new_view.rs
@@ -133,7 +133,7 @@ impl StateMachine {
         // In particular it is not guaranteed that the leader has the finalized block when
         // sending the NewView, so it might need to wait for the finalized block.
         //
-        // Note that for this process to work e2e, the replicas should NOT ignore th NewView from
+        // Note that for this process to work e2e, the replicas should NOT ignore the NewView from
         // the leader, even if they already advanced to the given view.
         // Note that the order of NewView and proposal messages doesn't matter, because
         // proposal is a superset of NewView message.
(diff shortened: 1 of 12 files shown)


node/components/bft/src/chonky_bft/new_view.rs
```diff
@@ -118,9 +123,23 @@ impl StateMachine {
...
         self.phase = validator::Phase::Prepare;
+        // It is important that the proposal and new_view messages from the leader
+        // will contain the same justification.
+        // Proposal cannot be produced before previous block is processed,
+        // therefore leader needs to make sure that the high_commit_qc is delivered
+        // to all replicas, so that the finalized block is distributed over the network.
+        // In particular it is not guaranteed that the leader has the finalized block when
+        // sending the NewView, so it might need to wait for the finalized block.
+        //
+        // Note that for this process to work e2e, the replicas should NOT ignore th NewView from
```
brunoffranca, line 136:
```suggestion
        // Note that for this process to work e2e, the replicas should NOT ignore the NewView from
```

node/tools/src/main.rs
```diff
@@ -20,8 +20,9 @@ struct Cli {
     #[arg(long, conflicts_with = "config_path")]
     config: Option<String>,
     /// Path to the rocksdb database of the node.
-    #[arg(long, default_value = "./database")]
-    database: PathBuf,
+    /// If not provided, an inmemory database will be used instead.
```
brunoffranca, line 23:
```suggestion
    /// If not provided, an in-memory database will be used instead.
```

brunoffranca (dismissed):

